PR #109, opened 2020-04-20
Remove duplicate ISSUE_TEMPLATE.md

henrykironde:


henrykironde:
It does look like we have this GitHub template folder which I do not see being used.
If that is correct, we can remove it else we can close the PR

PR #110, opened 2020-04-20
Keep editor file in the home directory

henrykironde:


henrykironde:
Looks like we could keep the .editorconfig file in the home directory.
If otherwise we can close the PR.

PR #124, opened 2020-05-19
Clean up Deepforest modules

henrykironde:
Embed `keras_retinanet` module imports in a try block


henrykironde:
@bw4sz, I cannot access the build logs for the failing tests, Is there a way to add me to the permissions or change them?

PR #139, opened 2020-10-20
provide thickness option for predict_image function

dingyif:


bw4sz:
Thanks for the PR. Azure pipeline passed, just need to recompile docs and then i'll merge.

PR #141, opened 2020-10-20
add progress bar for predict_tile

dingyif:
Not sure this is necessary or not, but when I tried to make a prediction for a large tiff file, it will be better if I know the remaining time.


bw4sz:
Regenerating travis build, github rate limit failed (no fault of the code), just an annoying facet of downloading the release model when building tests. An area of improvement to be more self-contained.

bw4sz:
confirmed that tqdm is in requirements.txt, since it is already used for the download progress bar. Thanks for the PR, keep it coming.

PR #142, opened 2020-10-22
Add thickness and color choices for evaluate function  

dingyif:
I found out the default color is not colorblind-friendly, especially when you work on the forest that background is green and the bounding box is red. So I think to give people more flexibility to choose their desire color and also make the bounding box thickness big enough to see.


ethanwhite:
I think it would also be great to set the defaults to something that is color blind friendly.

PR #178, opened 2021-04-03
Test SSL 

henrykironde:
Use latest cibuildwheel


henrykironde:
This may not be a stable fix. But we can rely of this and of the error persists we can try `pip install --trusted-host pypi.org --trusted-host files.pythonhosted.org --upgrade pip` specifying the trusted hosts

bw4sz:
Thanks!

Issue #49, opened 2019-12-04
Sample Image Issue


bw4sz:
Name: Ben Weinstein
Affiliation: University of Florida
Location: Frog Rock, Wyoming, United States

![example_image](https://user-images.githubusercontent.com/1208492/70186072-f0930700-169f-11ea-92ca-72ccefde27d0.png)

May we use this image in the gallery? : yes

PR #179, opened 2021-04-04
Update README.md in response to Sample Image Issue numbered 49

malayjoshi13:
This commit is in response to "Sample Image Issue #49" posted in the "weecology/DeepForest" repository's issues section.

This image will let the user get a better practical real-time use-case of this python package. It will do so as this image has a combination of both trees (that too of different variations in height and crown size) and non-tree objects (to be specific vehicles, buildings and concrete flooring ). However, still from the image attached, it is very well clear that the python package has very accurately recognized only trees and marked them with bounding boxes.

Thus, in short, this committed image will encourage more users to trust, get more comfortable using this python package in their projects and contributing in return towards the development of the DeepForest package.


bw4sz:
Please just add the image alongside the previous image, rather than replacing it.

malayjoshi13:
Okay @bw4sz

PR #180, opened 2021-04-05
Update README.md

malayjoshi13:
The new image is placed alongside the existing one.


malayjoshi13:
Hey @bw4sz, I have placed a new image below the existing one like what you suggested to do in my previous PR. Please review this PR and suggest any other changes you would like me to make in my commit. Thanks.

PR #182, opened 2021-04-08
Ease out training data generation process

malayjoshi13:
Hi, yesterday I planned to re-train the existing NEON.h5 model upon the self-prepared dataset. However, very soon, I got stuck with the problem of how to code down the script for converting all training images and their corresponding annotations (present in XML format) into a single headerless CSV file (of a particular pre-defined format) within a single loop of code by taking the reference of code present in docs/training.MD and modules like preprocess.py and utilities.py.

There were three problems that I faced with the above scenario:-
1) Firstly, the code present in docs/training.MD was just limited to a single training image and its corresponding XML annotation file. 

2) Secondly, expecting users to scale that code present in docs/training.md to get the same result with all of their training images and corresponding XML annotations files will be quite hard on their part and thus will refrain users to use DeepForest library and to enhance its accuracy by re-training its weights.

3) Thirdly, digging deep into this issue, one gets clear that the problem arises because functions like "xml_to_annotations" and "split_raster" present inside utilities.py and preprocess.py, respectively, were not utilized for working with many files at a single time. 

As a solution, I propose a module named "train_data_generator.py". 

This module will make the user's experience while generating the dataset for the re-training process very smooth. A user has to create a folder that says "dataset" in his/her local system, place the "annotations" folder (containing all XML format annotations) and "images" folder (containing all training images dataset) inside the newly created "dataset" folder, execute "pip install DeepForest" in command prompt, opening a python file and executing these code lines:- 

from deepforest import deepforest
from deepforest import utilities
location = "C:/Users/mycomputer/Desktop/dataset"
prepare_traindata(location)

..................and that is it. 
User with very few lines of code will get his/her final output in two folders, "training_annotations" (containing combined_csv.csv file) and "training_images" (containing all training images in required png format), which will be used later to re-train the model weights. 

Please do review the proposal. Thanks in advance.

### DIFF
--- a/deepforest/train_data_generator.py
+++ b/deepforest/train_data_generator.py
@@ -1,15 +1,15 @@
-"""This module is used to generate a .csv file called "combined_csv.csv" which will store annotations of all training images. Later on "combined_csv.csv" will be used for re-training
+"""This module generates a .csv file called "combined_csv.csv", which will store annotations of all training images. Later on, "combined_csv.csv" will be used for re-training
 the existing NEON.h5 model. 
 
-To generate "combined_csv.csv" using this module firstly create a directory "dataset" at any location of local system. Then place all training images inside "images" folder and all 
-corresponding annotations present in form of xml format (created by LabelImg or RectLabel) inside "annotations" folder located inside directory "dataset" present at local system.
+To generate "combined_csv.csv" using this module, firstly create a directory "dataset" at any local system location. Then place all training images inside the "images" folder and all 
+corresponding annotations present in the form of XML format (created by LabelImg or RectLabel) inside the "annotations" folder located inside directory "dataset" present at local system.
 
-Then create a new python file "get_data.py" and pass location of "dataset" directory with reference to user's local system inside function "train_data_generator.prepare_traindata". 
-Here "utilities" is imported through "from deepforest import deepforest" following with "from deepforest import utilities" command in python file.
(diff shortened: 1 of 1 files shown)


malayjoshi13:
Hi @bw4sz, can you please review this PR and tell if you feel any changes to be done to it.

bw4sz:
I do not think this is a problem that should be covered in this code base. I think a better solution is to add to the docs and make them clearer. The more functions that are added, they need to be supported and tested. Especially when it comes to data loading (a weak point in the package), we should not be making any more assumptions.
```
problem of how to code down the script for converting all training images and their corresponding annotations (present in XML format) into a single headerless CSV file (of a particular pre-defined format) 
```
Can you make an issue with the docs label and show what you tried and we can work on the docs. This doesn't strike me as 

difficult, but I'll need to look at exactly what you mean. Just

1. Convert annotations to pandas dataframe using the xml_to_annotations utility, stored in a list.
2. pd.concat() to create a master dataframe.

3. Save that dataframe alongside the crops.

untested pseudo-code
```
import deepforest
import pandas as pd
import glob

xmls = glob.glob("path/*.xml")
results = []
for xml in xmls:    
    df = deepforest.utilities.xml_to_annotations(xml)
    results.append(df)

results = pd.concat(results)
results.to_csv("path_to_crops/train_annotations.csv",header=False)
```

I welcome additions to docs.

PR #183, opened 2021-04-09
Add show parameter for predict_tile

amifunny:
[DeepForest](https://github.com/weecology/DeepForest/blob/master/deepforest/deepforest.py) gives `predict_image` function that predicts objects and also allows to show the plot of the image using `show` parameter. This parameter was not available for `predict_tile`. This would be helpful by having consistent parameters with `predict_tile` and viewing raster image examples.


bw4sz:
sounds good. If it passes i'll merge.

PR #207, opened 2021-06-24
Fix ValueError to print information, updated function Args comment

davidnyberg:
Previously the ValueError would not display which image paths were raising the error but instead: "UnboundLocalError: local variable 'plot_name' referenced before assignment.

Updated comments for parameters to function


bw4sz:
Thanks.

PR #217, opened 2021-07-15
Explicitly setting the datatype of targets['labels'] to an int64

ElliotSalisbury:
Instead of assuming the default int would be 64bits

targets["labels"] is used during training to index a tensor, and in pytorch the indexing must be performed with int64's

As seen in #201 
https://github.com/weecology/DeepForest/issues/201#issuecomment-880319824

### DIFF
--- a/deepforest/dataset.py
+++ b/deepforest/dataset.py
@@ -79,7 +79,7 @@ class TreeDataset(Dataset):
             
             # Labels need to be encoded
             targets["labels"] = image_annotations.label.apply(
-                lambda x: self.label_dict[x]).values.astype(int)
+                lambda x: self.label_dict[x]).values.astype(np.int64)
     
             augmented = self.transform(image=image, bboxes=targets["boxes"], category_ids=targets["labels"])
             image = augmented["image"]


bw4sz:
Let me see the tests, but I can't think of any downside in being safe with dtypes, thanks for hunting it down. I've never encountered this error on mac or linux.

ElliotSalisbury:
Yeah, I don't think this should change anything for mac or unix users
I did a quick test of the `astype(int)` thing by running `print(np.array([0,1,2,3]).astype(int).dtype)` on a local windows conda install, and a local windows python 3.9 virtual environment, and a docker version of python 3.9 running in an ubuntu image.

and the docker version returned int64, while the local windows versions returned int32
So my guess is that by default, the unix users have been running with the correct dtype anyway, and so won't see any change

ElliotSalisbury:
which looks to be a known weirdness with windows

https://stackoverflow.com/questions/36278590/numpy-array-dtype-is-coming-as-int32-by-default-in-a-windows-10-64-bit-machine

bw4sz:
this might be worth a minor patch.

PR #242, opened 2021-08-11
Remove pytorch channel from conda setup to fix testing

ethanwhite:


ethanwhite:
macos only installs CPU based builds, which results in slightly different versions of things which are causing the error. I'm not sure if this is useful for anything since we're really trying to get conda builds working that the feels out of our hands given the need for lightening and the issues with conda-forge, so I'm going to go ahead and close this, but we can get Ubuntu up and running in GH actions if we want to.

henrykironde:
After tests on the `macos` image, I think the `macos` image may need to be updated by the people providing the action. I have tried to replicate this repo, https://github.com/vaexio/vaex/blob/master/.github/workflows/pythonpackage.yml. There was some slight change in the error. It looks like the new error could be connected to the package `torch`. https://github.com/henrykironde/DeepForest/pull/3/checks?check_run_id=3346961787 

Not many packages are testing on `macos` and `conda` on github actions. We could give it some time as I look for places that I can place some issues related to the failures.

ethanwhite:
If conda packing is working on Linux & Windows (don't know if we've checked this) then I think it would be OK to proceed with getting the conda builds going and just have the docs note that macOS installs need to be done with pip.

ethanwhite:
Closing this since we're working on build fixes elsewhere.

Issue #15, opened 2019-11-12
add control over finetuning


bw4sz:
Depth of layers in DeepForest/retinanet_train.py

Need to be able to specify which layers to frees before compiling.

reduce the learning rate.

https://flyyufelix.github.io/2016/10/03/fine-tuning-in-keras-part1.html

bw4sz:
Just merge the boxes in the overlapping area, not use IoU, it will penalize one of the boxes too much.

PR #246, opened 2021-08-28
Add documentation for deepforestr

ethanwhite:


henrykironde:
@Ethan, this looks good. I have some suggestions.
1) Maintaining docs in both deepforestr and `here` may cause some inconsistencies. 
How about we use conf.py to fetch the README.md from deepforestr. We can also trim of the top badges and conf.py can create the docs at build time. https://raw.githubusercontent.com/weecology/deepforestr/main/README.md. This could follow the workflow in the [retriever dataset file](https://github.com/weecology/retriever/blob/main/docs/datasets_list.rst)

henrykironde:
2. On Mac am getting some errors that we need to sort out.

> library(deepforestr)
> model = df_model()
Reading config file: /Users/henrysenyondo/Library/r-miniconda/envs/r-reticulate/lib/python3.6/site-packages/deepforest/data/deepforest_config.yml
> model$use_release()
Model from DeepForest release https://github.com/weecology/DeepForest/releases/tag/1.0.0 was already downloaded. Loading model from file.
Loading pre-built model: https://github.com/weecology/DeepForest/releases/tag/1.0.0
> image_path = get_data("OSBS_029.png") # Gets a path to an example image
> bounding_boxes = model$predict_image(path=image_path, return_plot=FALSE)
OMP: Error #15: Initializing libiomp5.dylib, but found libomp.dylib already initialized.
OMP: Hint This means that multiple copies of the OpenMP runtime have been linked into the program. That is dangerous, since it can degrade performance or cause incorrect results. The best thing to do is to ensure that only a single OpenMP runtime is linked into the process, e.g. by avoiding static linking of the OpenMP runtime in any library. As an unsafe, unsupported, undocumented workaround you can set the environment variable KMP_DUPLICATE_LIB_OK=TRUE to allow the program to continue to execute, but that may cause crashes or silently produce incorrect results. For more information, please see http://www.intel.com/software/products/support/.
Abort trap: 6

PR #263, opened 2021-11-05
Export predictions for detailed analysis

ethanwhite:
Exporting the model predictions allows for image level comparisons of
counts and other detailed comparisons of predictions and observations.


ethanwhite:
Merging since this is a simple change with a locally passing test.

PR #267, opened 2021-11-05
USe docutils<0.18 due to backward compatibility

henrykironde:


dev_requirements.txt
```diff
@@ -27,3 +28,4 @@ torchvision >= 0.9.0
 xmltodict
 comet_ml
 imagecodecs
+docutils<0.18
```
ethanwhite, line 31: Looks like two copies of docutils

ethanwhite (commented):
Everything is green!

dev_requirements.txt
henrykironde: Thanks for catching that

PR #294, opened 2022-01-20
fixed basename extraction

easz:
The second `splitext` call is not necessary and
will cause a wrong name if a file name contains "." dot symbol (e.g. `my.img.file.png`)


easz:
I will separate different fixes in different PR

Issue #300, opened 2022-01-30
missing variables in  DeepForest/deepforest/data/deepforest_config.yml 


saadmazouzi:
**Describe the bug**
In the  DeepForest/deepforest/main.py we find :
in the line 169 self.config["train"]["preload_images"] -> which mean the deepforest_config should contain it but it's not the case
also in the line 413 self.config["validation"]["val_accuracy_interval"] == 0  we need to add the val_accuracy_interval in the config file 

**To Reproduce**

`m= main.deepforest()` 
` m.config['gpus'] = '-1' #move to GPU and use all the GPU resources` 
` m.config["train"]["csv_file"] = annotations_file` 
` m.config["train"]["root_dir"] = os.path.dirname(annotations_file)` 
` m.config["score_thresh"] = 0.4` 
` m.config["train"]['epochs'] = 2` 
` m.config["validation"]["csv_file"] = validation_file` 
` m.config["validation"]["root_dir"] = os.path.dirname(validation_file)` 

` m.create_trainer()` 
` m.use_release()` 
` start_time = time.time()` 
` m.trainer.fit(m)#there is the probleme` 
` print(f"--- Training on GPU: {(time.time() - start_time):.2f} seconds ---")` 

**Environment (please complete the following information):**
 - OS: windows
 - Python version and environment : 3.7 and google colab

**Screenshots**
![issue1](https://user-images.githubusercontent.com/98532815/151690393-333a1369-1003-46d4-9e24-5e6b10ec2f2a.PNG)
![issue2](https://user-images.githubusercontent.com/98532815/151690441-0f2c6054-3151-468e-af71-5b6cbfaf1a0a.PNG)

**Additional context**
we can fix it just by adding to the config file  this two missing config 
or just before every training we need to add this two lines  :

m.config["train"]["preload_images"] = "" 
m.config["validation"]["val_accuracy_interval"] = 5

bw4sz:
What version of deepforest do you have? Looking on the git repo, these values already exist? 

https://github.com/weecology/DeepForest/blob/0c7c881ada5e01a3fc33827b0eeffa96265680eb/deepforest_config.yml#L27

One idea is when you do 
```
pip install --upgrade
```

we need to make sure package data gets updated.

bw4sz:
Looking here at latest pypi release (1.2.0)

https://github.com/weecology/DeepForest/compare/1b53767ca632ae17df44d47a3e84632d62a34922...main

[...]

PR #302, opened 2022-01-30
add missing config variables

easz:
PR for #300 

alternatively there can be default values in corresponding function calls

### DIFF
--- a/deepforest/data/deepforest_config.yml
+++ b/deepforest/data/deepforest_config.yml
@@ -24,9 +24,13 @@ train:
     #Useful debugging flag in pytorch lightning, set to True to get a single batch of training to test settings.
     fast_dev_run: False
     
+    preload_images: False
+
 validation:
     #callback args
     csv_file: 
     root_dir:
     #Intersection over union evaluation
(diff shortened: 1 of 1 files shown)


ethanwhite:
These are in the current main branch. See:
* https://github.com/weecology/DeepForest/blob/main/deepforest_config.yml#L27
* https://github.com/weecology/DeepForest/blob/0c7c881ada5e01a3fc33827b0eeffa96265680eb/deepforest_config.yml#L35

I wonder if you have an older version lying around from 2+ months ago?

easz:
there is also another config in data folder: [data/deepforest_config.yml](https://github.com/weecology/DeepForest/blob/main/deepforest/data/deepforest_config.yml) which is not complete. I think the `get_data` function will search in 'data' folder.

https://github.com/weecology/DeepForest/blob/0c7c881ada5e01a3fc33827b0eeffa96265680eb/deepforest/__init__.py#L14

ethanwhite:
Thanks for the clarification @easz - I'd totally forgotten about that and didn't look at the path. Much appreciated. In it goes.

PR #305, opened 2022-01-31
Improve conftest

bw4sz:
Refactor conftest to only import use_release() once. This may help the API github checks.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,11 +3,49 @@
 import pytest
 from deepforest import utilities
 from deepforest import get_data
+from deepforest import main
 import os
+
+os.environ['KMP_DUPLICATE_LIB_OK']='True'
 
 @pytest.fixture(scope="session")
 def download_release():
-    print("running fixtures")
+    print("Downloading release model once for fixtures")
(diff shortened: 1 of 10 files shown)


ethanwhite:
I reran this and it passed, so it is passing at least some of the time, but those rate limits can be idiosyncratic

bw4sz:
I am thinking about shelving this PR. The benefits is that we use the conftest.py correctly, the tests are faster and better organized. The negatives is that it seems somewhat unstable for reasons I don't understand. Working sometimes more than others.

ethanwhite:
Is it always just the github API rate limit when it fails? We could always add an API key as a secret which would stop that.

ethanwhite:
OK. I think we're probably hitting api code in `use_release()` too many times so another option would be to download the current release once at the start of testing and then for any another uses of `use_release()` add the `check_release=False` argument so that we don't keep hitting the API. (suggestion brought to you by waiting on the really slow test to run on GH Actions this morning).

ethanwhite:
I've also noticed some instability that is independent of anything in here and have a fix coming in as part of #314.

PR #314, opened 2022-03-07
Return actual labels (not numeric codes) in predictions

ethanwhite:


bw4sz:
I'm on the fence about this, I think there will be downstream effects. What's the need? The user inputs the label dict and can convert back.

ethanwhite:
It's for consistency. We're returning `results["results"]` with actual labels (see lines 491 & 492) so presumably we should be doing the same with `results["predictions"]`. We added this line originally to allow the resulting predictions object to be compared to the results object (e.g, to compare counts of numbers of birds) so it seems weird to require post-processing on one but not the other. I'm also unclear about the potential downstream effects since we only fairly recently added this for this particular use case.

ethanwhite:
Here's the commit adding this from November: https://github.com/weecology/DeepForest/commit/df7d5a110eb6421c9faf7696965a8d105cf46f0c

Really that change was incomplete in not performing the same label translation as in `results["results"]`, so I think of this as a fix on the initial addition.

ethanwhite:
Test are now passing. They were failing due to an independent in consistency, which is fixed in this PR by bumping the Python version for testing and switching to mamba for installation. I can split if useful, but since these are so small I thought I'd leave them together for now.

ethanwhite:
@bw4sz - any further thoughts on this?

bw4sz (approved):
I think this looks good.

PR #315, opened 2022-03-12
Clarify that input data is in RGB

### LINKED ISSUE #306: confusing document about image convention
**Describe the bug**

the input argument `image` from the API `predict_tile` is documented as `in BGR channel order` which is confusing. I think it should be just in usual RGB order, therefore wrong.

https://github.com/weecology/DeepForest/blob/0c7c881ada5e01a3fc33827b0eeffa96265680eb/deepforest/predict.py#L151

Some lines later...

https://github.com/weecology/DeepForest/blob/0c7c881ada5e01a3fc33827b0eeffa96265680eb/deepforest/predict.py#L249

we see that the input image (which should be RGB) is now actually converted to BGR before calling  `plot_predictions` (with `cv2.rectangle`).

~~~
if return_plot:
            # Draw predictions on BGR 
            image = image[:,:,::-1]
            image = visualize.plot_predictions(image, mosaic_df, color=color, thickness=thickness)
            # Mantain consistancy with predict_image
            return image
~~~

actually all input argument with `image` should be just usual "RGB" images and the returned plot images (i.e. `predict_tile(predict_tile=True)` should be in "BGR". Right?

Maybe you can update the document, or just make input / output image format consistently only with "RGB" format.

ethanwhite:
Also remove comment to clarify that switch to BGR is already
handled in predict_image.

Fixes #306

### DIFF
--- a/deepforest/predict.py
+++ b/deepforest/predict.py
@@ -148,8 +148,7 @@ def predict_tile(model,
         device: pytorch device of 'cuda' or 'cpu' for gpu prediction. Set internally.
         numeric_to_label_dict: dictionary in which keys are numeric integers and values are character labels
         raster_path: Path to image on disk
-        image (array): Numpy image array in BGR channel order
-            following openCV convention
+        image (array): Numpy image array in RGB channel order ranged from 0-255
         patch_size: patch size default400,
         patch_overlap: patch overlap default 0.15,
         iou_threshold: Minimum iou overlap among predictions between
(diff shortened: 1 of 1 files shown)


ethanwhite:
@easz - does this look good to you for addressing #306?

ethanwhite:
The test fail isn't related to the PR and should be fixed when #314 is merged.

PR #330, opened 2022-07-02
Resample image and evaluating them accordingly

dassaniansh:


bw4sz:
Great! Let's go through this together.

1. This analysis should live in its own repo. Pip install deepforest and we can isolate this analysis from the source code.
2. Let's run in an ipython notebook or google colab (you can git clone the NeonTreeEvaluation data). Then we can visualize a graph with resolution on x axis, evaluation score on the y axis.
3. Create a single function that wraps this code together and we can apply it to a few hundred tiles in the NeonTreeEvaluation dataset. Then we can start to look at patterns by site and geography.


Thanks for this, really interested to see results.

bw4sz:
Can I get an update here? Let's look at the that the list of points above and either post here or in the channel. We can change aspects here if helpful.

PR #357, opened 2022-09-10
Training model on OSBS Dataset

dassaniansh:


ethanwhite:
Looks like we've got duplicates of the image files in a `_MACOSX` subdirectory. Let's remove those.

dassaniansh:
Thanks for it @ethanwhite, Sorry I didn't notice that I pushed those files too.

ethanwhite:
No problem - that's why we do reviews!

ethanwhite (approved):

ethanwhite:
@henrykironde & @bw4sz I think this PR got lost in the fall. My review was just on a detail since I wasn't involved in the project. Is this something we want to get merged in?

bw4sz:
No, this is not the location for these experiments. As part of milestone 2.0, we need some organization for where DeepForest research goes.

PR #385, opened 2023-02-21
Apply yapf

henrykironde:


henrykironde:
This should initially fail. The pr will pass once code is refactored.

henrykironde:
These are refracting changes, no change in logic.

ethanwhite (approved):

Issue #321, opened 2022-03-29
Ask user if they want to download the latest model.


bw4sz:
We have not specifically gotten into this problem yet, but there should be a mechanism to ask user if they want to download a new model, instead of forcing it here.

https://github.com/weecology/DeepForest/blob/e31ecabf07b1b7d2444ba2c1d1298df79a416c27/deepforest/utilities.py#L82

ethanwhite:
Just a note on implementation - I think this should be an optional argument like `update = True`, not an interactive query. I think we also want a related optional argument that lets you pin to a specific release.

puranjay123:
@ethanwhite @bw4sz 
Hi, I am a Puranjay Kwatra GSOC participant and I want to work on this project so can I start with this issue any guidance would really helpful if you could please guide me I want to contribute to this organization as It matches my specific area of interests.

ayeankit:
Hello @ethanwhite, I want to work on this issue but I am new to this repo and I need help. Please assign me this.

ethanwhite:
Hi @ayeankit - @henrykironde is working on getting some beginner developer docs up this week that will help you get started on things.

Having glanced at the code related to this issue I think this one is actually something to check in with @bw4sz about. The function already has a `check_release` optional argument which could be set to `False` to avoid updating. I guess @bw4sz is interested in being able to check separately from updating.

bw4sz:
Given that we don't have a use case yet, its difficult to know what the pros and cons are here. I think the simplest thing is here: https://github.com/weecology/DeepForest/blob/906730656579ed215d69d10e0757cf5dcceeb759/deepforest/utilities.py#L143, we need to ask the user if we want to update. So we need an update=True flag and then the user needs to be alerted that a new model will be downloaded.

ethanwhite:
@bw4sz - a couple of follow up questions:

1. That code only gets hit if `check_release==True`, so it feels like we basically have this kind of a flag implemented. Are you envisioning separating the check to see if there is a new release available and the installing of the new release if it exists?
2. When you say "the user needs to be alerted that a new model will be downloaded" do you just meaning displaying some output or were you envisioning an interactive prompt?

ayeankit:
Hi @ethanwhite and @bw4sz , I have done changes and raise the PR. Can you please review it? 

 Link=(https://github.com/weecology/DeepForest/pull/382)

ayeankit:
[...]

PR #388, opened 2023-02-27
Ask user if they want to download the latest model. #321 with test cases 

ayeankit:
Implementation for `check_new_release` method and write up code for the user permission if they want to download the latest model . And added test cases of new function in test folder. and clean up the code according to google chromium style. I have created test cases.


henrykironde:
@ayeankit, please make sure that you have the tests passing locally before you request a review.

PR #398, opened 2023-03-10
added drop alpha handler to predict_tile and predict_image and created unittest #392

Mu-Magdy:


bw4sz:
Thanks for your PR. Since this is your first, a few notes.

1. Please summarize your PR in the description, that way the reviewer can quickly tell what you intended to do.
2. Correct all pytest errors, which looks small, just arg ordering.

```
>       result = predict.predict_image(model=self.model,
                                       image=image,
                                       return_plot=return_plot,
                                       device=self.current_device,
                                       iou_threshold=self.config["nms_thresh"],
                                       color=color,
                                       thickness=thickness)
E       TypeError: predict_image() missing 1 required positional argument: 'path'
```
3. PRs should be minimal, that way if we break something we can tell when and where, and for what issue. For example, I'm all for adding a way to read images from path from predict_image, this is a perfectly reasonable PR, but that isn't this issue. We like to try to do small pull requests at a time, addressing just one issue. This is a good example why. The errors in the pytest are not from the issue at hand (warning about 3 band raster), but because of the arg order for reading from path.
4. Every new function should have a test. 

I see

```
def test_split_raster_alpha_drop(config):
    # Drop Alpha
    r=rasterio.open(config["path_to_raster"]).read()
    
    
    assert r.shape[0]==3
```
which doesn't fully test the issue on hand. A good test in this case would create a 4 band image, run it through predict_image and assert that a warning is raised (see pytest.warns).

Thanks for your contribution, I appreciate growing the collaborators for this code base.

deepforest/predict.py
```diff
@@ -18,6 +18,7 @@
 
 def predict_image(model,
                   image,
+                  path,
```
henrykironde, line 21: Changing the signature of a function is not the best way to go. This causes backward compatibility.

Avoid having the same code in several areas.

henrykironde (commented):
Make sure the tests pass locally

Mu-Magdy:
just made a new commit waiting for your review @bw4sz @henrykironde

henrykironde:
Make sure that the tests pass locally and let me know.

Mu-Magdy:
just made a new commit waiting for your review @henrykironde

ethanwhite:
> There are many changes because I reformatted the files but I can restore the old format and make a new PR

@Mu-Magdy - yeah, if you could restore the format and resubmit that would be great. Mixing formatting changes and meaningful code changes in a single PR makes reviewing a lot more difficult. Thanks.

PR #400, opened 2023-03-10
Ask user if they want to download the latest model. #321 with test cases passed 

ayeankit:
Implementation for '`check_new_release' `method and write up code for the user permission if they want to download the latest model . And added test cases of new function in test folder. and clean up the code according to google chromium style. I have created test cases and passed.

### DIFF
--- a/deepforest/data/current_bird_release.csv
+++ b/deepforest/data/current_bird_release.csv
@@ -1,0 +1,2 @@
+,current_bird_release
+0,https://github.com/weecology/BirdDetector/releases/tag/1.1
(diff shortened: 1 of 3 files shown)


ayeankit:
hey @henrykironde , I don't know why this is failing, Locally all the test cases are passed, as you can see.
Screenshot from 2023-03-10 17-29-55

Please suggest some solution for passing this.

henrykironde:
Can you run these two commands and let me know what the results are?

```
git clean -fxd
pytest -v -k"release"
```

ayeankit:
Hey @henrykironde , Here are results after executing the above command.

Screenshot from 2023-03-12 00-07-43
Screenshot from 2023-03-12 00-08-02
Screenshot from 2023-03-12 00-08-14

henrykironde:
So locally it does not pass the testing, Try to debug that, find out why there is a file missing.

PR #401, opened 2023-03-12
Ask user if they want to download the latest model. #321 with test cases passed and resolved

ayeankit:
Implementation for ``check_new_release' ` method and write up code for the user permission if they want to download the latest model . And added test cases of new function in test folder. and clean up the code according to google chromium style. I have created test cases and passed error solved of github issue.

### DIFF
--- a/deepforest/utilities.py
+++ b/deepforest/utilities.py
@@ -47,10 +47,20 @@ class DownloadProgressBar(tqdm):
         if tsize is not None:
             self.total = tsize
         self.update(b * bsize - self.n)
-
+def check_new_release():
+    # Find latest github tag release from the DeepLidar repo
+    _json = json.loads(
+        urllib.request.urlopen(
+            urllib.request.Request(
+                'https://api.github.com/repos/Weecology/BirdDetector/releases/latest',
+                headers={'Accept': 'application/vnd.github.v3+json'},
+            )).read())
+    asset = _json['assets'][0]
+    url = asset['browser_download_url']
+    return _json, url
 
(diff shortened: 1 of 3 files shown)


ayeankit:
Hey @henrykironde , I have some problem passing the `use_bird_release` test cases. After running this function the no **bird.pt** file is created. How can i resolve this?

henrykironde:
Describe the problem a little more.

deepforest/utilities.py
```diff
@@ -47,10 +47,20 @@ def update_to(self, b=1, bsize=1, tsize=None):
         if tsize is not None:
             self.total = tsize
         self.update(b * bsize - self.n)
-
+def check_new_release():
```
ayeankit, line 50: I have created this function for `check_new_release` which checks the for any new release.

deepforest/utilities.py
```diff
@@ -65,44 +75,37 @@ def use_bird_release(
     output_path = os.path.join(save_dir, prebuilt_model + ".pt")
```
ayeankit, line 76: Here the bird.pt file is not been created, so all the test cases file failed.

ayeankit:
Hey @henrykironde 

Expected:- I have expected **bird.pt** file to be created to pass the test cases of `use_bird_release`
Actual:- bird.pt file is not created which result in fail of `use_bird_release` test cases.

deepforest/utilities.py
henrykironde: Try to find out why it is not created

PR #409, opened 2023-03-21
Pass Python version to mamba correctly

ethanwhite:
python-version is not yet supported by the mamba action.

See:
* https://github.com/mamba-org/provision-with-micromamba#extra-specs
* https://github.com/mamba-org/provision-with-micromamba/blob/e2b397b12d0a38069451664382b769c9456e3d6d/action.yml#L121-L152


ethanwhite:
I had marked this as WIP due to the Azure error, but that is external from this change so this should be ready to go

PR #420, opened 2023-04-08
Bird model threshold

bw4sz:
This pull request sets the minimum predition confidence score at 0.3. This matches our internal testing on reliable predictions and gets rid of alot of erroneous predictions. The downside is that if a user wants to set a new threshold, they will need to do it after loading the model, since this will now overwrite config. I added a print statement. It could be a warning.

```
from deepforest import main
m = main.deepforest()
m.use_bird_release()
m.config["score_thresh"] = 0.5
```


ethanwhite (changes requested):
I think printing is fine.

There are a bunch of new test failures here, though I don't understand how they could be related to the PR. There are also multiple version change commits. I recommend that we clean up the extra commits and confirm that these test failures are happening on main (which they weren't as of a few days ago) before merging.

ethanwhite:
The new test failures are due to a new pandas release. I'm happy to merge this as is if you want, but since it had overlapping commits with other PRs I wanted to check before I did so since it may require more significant git cleanup that removing the extra commits here.

ethanwhite (approved):

PR #424, opened 2023-04-10
Increase min version for torchvision

ethanwhite:
Bump the min torchvision version to match the changes made to address pytorch warnings in cb05b95.


bw4sz:
I was just going to do this. Can you test commenting out yapf in the azure builds here: I think this is what is causing the azure failure. 

https://github.com/weecology/DeepForest/blob/5ed64f02fe9ce860e33103f4fb1b0f59f7b04a0b/azure-pipelines.yml#L35

PR #433, opened 2023-04-23
Move ignoring setup.py in testing to conftest.py

ethanwhite:
The config setting in setup.py was being ignored with the warning: `PytestConfigWarning: Unknown config option: collect_ignore`

This moves it to conftest.py, which is the location recommended in the docs: https://docs.pytest.org/en/4.6.x/reference.html#collect-ignore

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,3 @@ exclude = docs
 
 [aliases]
 test = pytest
-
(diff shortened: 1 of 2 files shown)


ethanwhite:
Test failure is due to https://github.com/weecology/DeepForest/pull/434 not this change

PR #435, opened 2023-04-23
Ignore warnings generated due to light weight testing

ethanwhite:
We use small numbers of batches and workers for testing. This triggers some warnings related to optimal production performance. This change keeps them out of the test logs.


bw4sz:
LGTM

PR #480, opened 2023-09-28
change readthedocs theme and update the landing page with new links

bw4sz:


bw4sz:
@henrykironde this is ready for review, just that silly bash error above.

henrykironde (commented):
I have gone through the PR and made some changes. 
I have done some rewording so read through and change accordingly.



In dataset.py I have restructured the return statement. In general, when creating functions, we should try not to have return statements that return different structures of objects.  


```
         Returns:
-            If train:
-                path, image, targets
-            else:
-                image
```

We can design the function to return  (image, path, targets)
In that case if not train, it will return (image-object, None, None) else (image-object, path-object, targets-objects) for train.

We do not need docs for setup and tests for now.


For this PR what is left is to read through the reworded phrases and also add `../www/OSBS_sample.png` we are missing that file.

bw4sz:
LGTM

henrykironde (approved):

PR #489, opened 2023-10-02
Update docs

bw4sz:
I added the missing image file and "Bird_panel.jpg" is actually "bird_panel.jpg".


henrykironde:
Seems like we are working from an old branch or one that was already merged. Can we get a new branch of main, and add the required changes to avoid conflicts?

PR #490, opened 2023-10-02
predicted labels should read "Bird" and not "Tree" by default when use_bird_release()

bw4sz:
I added a test and updated main.use_bird_release() to make sure that the bird label is automatically assigned instead of "Tree" default.

### DIFF
--- a/deepforest/main.py
+++ b/deepforest/main.py
@@ -116,6 +116,10 @@ class deepforest(pl.LightningModule):
         print("Setting default score threshold to 0.3")
         self.config["score_thresh"] = 0.3
 
+        # Set label dictionary to Bird
+        self.label_dict = {"Bird":0}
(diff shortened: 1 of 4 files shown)


bw4sz:
@henrykironde I cleaned up the change in the docs. This should be ready to go.

henrykironde:
Looks good to me, Can you run `yapf -i --recursive deepforest/ --style=.style.yapf`  and commit amend to clean up this code

```
deepforest/main.py	(reformatted)
117 
         self.config["score_thresh"] = 0.3
 
         # Set label dictionary to Bird
-        self.label_dict = {"Bird":0}
+        self.label_dict = {"Bird": 0}
```

bw4sz:
I added another commit since it gets squashed when merged into main.

PR #504, opened 2023-10-05
Torchmetrics integration for evaluation metrics

### LINKED ISSUE #412: Replace custom iou module with torchmetrics jaccard index.
As part of the general movement towards standardization with pytorch lightning and all of its wonderful features, the evaluate() module that employs all sort of custom logic needs to be aligned with the torchmetrics jaccard index. The benefits is greater speed and standardization. We need to document any changes in performance against the NeonTreeEvaluation benchmark. 

https://torchmetrics.readthedocs.io/en/stable/classification/jaccard_index.html

bw4sz:
The goal of this PR is to compare the iou in torchmetrics and eventually move to that implementation.  Torchmetrics is the growing state-of-art repo for well test implementations and comes from the pytorch-lightning team. The are fast and scalable to multiple GPU distributed training. Our current evaluation method is custom code and slow.

Closes #412


bw4sz:
I asked a Q, not quite clear yet how to align https://github.com/Lightning-AI/torchmetrics/discussions/2152

bw4sz:
I spent most of the day on this and my current sense is that torchmetrics are super useful addition to DeepForest evaluation, but will not replace our custom evaluation pipeline. The torchmetrics implementation is not well documented, seems rather new, and lacks some of the flexibility we'd like, see the discussion above. It is still very useful, especially during training. The current deepforest.callbacks.iou_callback is seriously slow and can drag down training, even when running ever 5 to 10 epochs. Instead we can use average intersection over union and the mAP scores as a indicator during training and save the heavier evaluation routine for once or twice, or just after training is done. As far as I can tell there is no matching algorithm between predictions and ground truth in any of the torchmetrics implementations, which stems from the classic computer vision problem of creating metrics and benchmark datasets for toy problems, like one or two objects that don't touch in an image. We have a much more cluttered task. I like having more evaluation metrics and I added some thoughts in docs/evaluation.md to articulate the different uses. I worry it might overwhelm the reader, but when it comes to evaluation, I think we want to help the community move towards more nuance. This takes more text.

bw4sz:
Tests pass locally and on azure, but the conda package fails. Clearly some version issues going on here, may need something pinned. I'm looking into it.

docs/Evaluation.md
```diff
@@ -1,18 +1,86 @@
...
+ We stress that evaluation data must be different from training data, as neural networks have millions of parameters and can easily memorize thousands of samples. Avoid random train-test splits, try to create test datasets that mimic downstream tasks. If you are predicting among temporal surveys or across imaging platforms, your train-test data should reflect these partitions. Random sampling is almost never the right choice, biological data often has high spatial, temporal or taxonomic correlation that makes it easier for your model to generalize, but will fail when pushed into new situations.
+
+DeepForest provides several evaluation metrics. There is no one size fits all evaluation approach and the user needs to consider is the evaluation metric best fits the task. There is significant information online evaluation of object detection networks. Our philosophy is to provide a user a range of statistics and visualizations. Always visualize results and trust your judgement. Never be guided by a single metric.
```
henrykironde, line 5: DeepForest provides several evaluation metrics. There is no one-size-fits all evaluation approach, and the user needs to consider which evaluation metric best fits the task. There is significant information online about the evaluation of object detection networks. Our philosophy is to provide a user with a range of statistics and visualizations. Always visualize results and trust your judgment. Never be guided by a single metric.

docs/Evaluation.md
```diff
@@ -1,18 +1,86 @@
...
+
+[MeanAveragePrecision in torchmetrics](https://medium.com/data-science-at-microsoft/how-to-smoothly-integrate-meanaverageprecision-into-your-training-loop-using-torchmetrics-7d6f2ce0a2b3)
+
+[A general explanation of the mAP metric](https://jonathan-hui.medium.com/map-mean-average-precision-for-object-detection-45c121a31173)
+
+[Comparing Object Detection Models](https://www.comet.com/site/blog/compare-object-detection-models-from-torchvision/)
+
+## Average Intersection over Union
+
+DeepForest modules use torchmetric's [IntersectionOverUnion](https://torchmetrics.readthedocs.io/en/stable/detection/intersection_over_union.html) metric. This calculates the average overlap between predictions and ground truth boxes. This can be considered a general indicator of model performance, but not sufficient on its own for model evaluation. There are lots of reasons predictions might overlap with ground truth, for example consider a model that covered an entire image with boxes. This would have high IoU, but low value to the user.
```
henrykironde, line 17: This calculates the average overlap between predictions and ground truth boxes. This can be considered a general indicator of model performance but is not sufficient on its own for model evaluation. There are lots of reasons predictions might overlap with ground truth; for example, consider a model that covers an entire image with boxes. This would have a high IoU but a low value.

docs/Evaluation.md
```diff
@@ -1,18 +1,86 @@
...
+
+mAP is the standard COCO evaluation metric and the most common for comparing computer vision models. It is useful as a summary statistic. However, it has several limitations for an ecological use case. 
+
+1. Not intuitive and difficult to translate to ecological applications. Read the sections above and visualize the mAP metric, which is essentially the area under the precision-recall curve at a range of IoU values. 
+2. The vast majority of biological applications use a fixed cutoff to determine an object of interest in an image. Perhaps in the future we will weight tree boxes by their confidence score, but currently we do things like, "All predictions > 0.4 score are considered positive detections". This does not connect well with the mAP metric.
+
+## Precision and Recall at a set IoU threshold.
+
+This was the original DeepForest metric, set a IoU = 0.4. This means that all predictions that overlap a ground truth box at IoU > 0.4 are true positives. As opposed to the torchmetrics above it is intuitive and matches downstream ecological tasks. The drawback is it slow, coarse, does not fully reward the model for having high confidence scores on true positives.
```
henrykironde, line 28: This was the original DeepForest metric, set to an IoU of 0.4. This means that all predictions that overlap a ground truth box at IoU > 0.4 are true positives. As opposed to the torchmetrics above, it is intuitive and matches downstream ecological tasks. The drawback is that it is slow, coarse, and does not fully reward the model for having high confidence scores on true positives.

There is an additional difference between ecological object detection methods like tree crowns and traditional computer vision methods. Instead of a single or set of easily differentiated ground truths, we could have 60 or 70 objects that overlap in an image. How do you best assign each prediction to each ground truth?

docs/Evaluation.md
```diff
@@ -1,18 +1,86 @@
...
+
+```    
+    m = main.deepforest()
+    csv_file = get_data("OSBS_029.csv")
+    root_dir = os.path.dirname(csv_file)
+    m.config["validation"]["csv_file"] = csv_file
+    m.config["validation"]["root_dir] = root_dir
+    results = m.trainer.validate(m)
+```
+This creates a dictionary of the average IoU ('iou') as well as 'iou' for each class. Here there is just one class, 'Tree'. Then the COCO mAP scores. See Further Reading above for explanation of mAP level scores. The val_bbox_regression is the loss function of the object detection box head, the loss_classification is the loss function of the object classification head.
```
henrykironde, line 46: This creates a dictionary of the average IoU ('iou') as well as 'iou' for each class. Here there is just one class, 'Tree'. Then the COCO mAP scores. See Further Reading above for an explanation of mAP level scores. The val_bbox_regression is the loss function of the object detection box head, and the loss_classification is the loss function of the object classification head.

docs/Evaluation.md
```diff
@@ -1,18 +1,86 @@
...
+        mar_small                   0.0
+   val_bbox_regression      0.5196723341941833
+   val_classification       0.4998389184474945
+```
+
+> **_Advanced tip:_**  Users can set the frequency of pytorch lightning evaluation using kwargs passed to main.deepforest.create_trainer(). For example [check_val_every_n_epochs](https://lightning.ai/docs/pytorch/stable/common/trainer.html#check-val-every-n-epoch).
+
+## Recall and Precision at a fixed IoU Score
+
+To get a recall and precision at a set IoU evaluation score, specify an annotations file an using the m.evaluate method.
```
henrykironde, line 73: To get a recall and precision at a set IoU evaluation score, specify an annotations' file using the m.evaluate method.

docs/Evaluation.md
```diff
@@ -40,6 +107,8 @@ results["box_precision"]
 0.781
 ```
 
+### Worked example of calculating IoU and recall/precision values
+
 To convert overlap among predicted and ground truth bounding boxes into measures of accuracy and precision, the most common approach is to compare the overlap using the intersection-over-union metric (IoU).
```
henrykironde, line 112: IoU is the ratio between the area of overlap between the predicted polygon box and the ground truth polygon box divided by the area of the combined bounding box region.

docs/Evaluation.md
```diff
@@ -140,8 +209,10 @@ result["class_recall"]
 0  Tree     1.0    0.67033    61
 ```
 
+### How to average evaluation metrics across images?
+
 One important decision was how to average precision and recall across multiple images. Two reasonable options might be to take all predictions and all ground truth and compute the statistic on the entire dataset. 
-This strategy makes more sense for evaluation data that are relatively homogenous across images. We prefer to take the average of per-image precision and recall. This helps balanace the dataset if some images have many trees, and other have few trees, such as when you are comparing multiple habitat types. 
+This strategy makes more sense for evaluation data that are relatively homogenous across images. We prefer to take the average of per-image precision and recall. This helps balanace the dataset if some images have many objects, and other have few objects, such as when you are comparing multiple habitat types.
```
henrykironde, line 215: This strategy makes more sense for evaluation data that is relatively homogenous across images. We prefer to take the average of per-image precision and recall. This helps balance the dataset if some images have many objects and others have few objects, such as when you are comparing multiple habitat types.
Users are welcome to calculate their own statistics directly from the results dataframe.

docs/Evaluation.md
```diff
@@ -154,7 +225,7 @@ result["results"].head()
 34             34         4  0.595862  ...       Tree  OSBS_029.tif   True
 ```
 
-## Evaluating tiles
+### Evaluating tiles too large for memory
 
 The evaluation method uses deepforest.predict_image for each of the paths supplied in the image_path column. This means that the entire image is passed for prediction. This will not work for large images. The deepforest.predict_tile method does a couple things under hood that need to be repeated for evaluation.
```
henrykironde, line 230: The deepforest.predict_tile method does a couple things under `the` hood that need to be repeated for evaluation.

henrykironde (changes requested):
This looks good. I have just made a few recommendations and rephrased some parts.

docs/Evaluation.md
bw4sz: Changed.

henrykironde (approved):

PR #514, opened 2023-10-15
Close #511, replaced convert_to_boxes with geometry_type

### LINKED ISSUE #511: convert_to_boxes arg name in shapefile_to_annotations is poorly named
The point of this argument is to allow users that have annotated points and create a fixed box size of size arg 'buffer_size'. How about if buffer_size is not None, convert from points to boxes, and just drop convert_to_boxes arg completely. 

https://github.com/weecology/DeepForest/blob/8de66db9add730a7dca8fb5f8f0abdeccfb669da/deepforest/utilities.py#L257

JSpencerPittman:
This is just a simple swap with convert_to_boxes to with geometry_type to allow polygons to be used in the shape file further down the road. I also added some documentation regarding the handling of different geometries within the shape-file in the doc-strings body.

### DIFF
--- a/deepforest/utilities.py
+++ b/deepforest/utilities.py
@@ -251,7 +251,7 @@ def shapefile_to_annotations(shapefile,
     Convert a shapefile of annotations into annotations csv file for DeepForest training and evaluation
 
     Geometry Handling:
-    The geometry_type is the form of the objects in the given shapefile it can be "boundingbox" or "point".
+    The geometry_type is the form of the objects in the given shapefile. It can be "boundingbox" or "point".
     If geometry_type is set to "boundingbox" (default) then the bounding boxes in the shapefile will be used as is and transferred over
     to the annotations file. However, if the geometry_type is "point" then a bounding box will be created in its place that's centered on
     the original point with an apothem equal to buffer_size.
--- a/deepforest/utilities.py
+++ b/deepforest/utilities.py
@@ -262,14 +262,15 @@ def shapefile_to_annotations(shapefile,
         rgb: Path to the RGB image on disk
         savedir: Directory to save csv files
         buffer_size: size of point to box expansion in map units of the target object, meters for projected data, pixels for unprojected data. The buffer_size is added to each side of the x,y point to create the box. 
-        geometry_type (bbox): Specifies the spatial representation used in the shapefile; can be "bbox" or "point"
+        geometry_type: Specifies the spatial representation used in the shapefile; can be "bbox" or "point"
     Returns:
         results: a pandas dataframe
     """
 
     # Verify the geometry_type is valid
     if geometry_type not in ["bbox", "point"]:
-        raise ValueError(f"Invalid argument for 'geometry_type'. Expected 'point' or 'bbox'.")
+        raise ValueError(
+            "Invalid argument for 'geometry_type'. Expected 'point' or 'bbox'.")
 
     # Read shapefile
     gdf = gpd.read_file(shapefile)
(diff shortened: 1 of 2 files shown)


deepforest/utilities.py
```diff
@@ -245,24 +245,33 @@ def xml_to_annotations(xml_path):
...
+                             geometry_type="boundingbox",
                              savedir="."):
     """
     Convert a shapefile of annotations into annotations csv file for DeepForest training and evaluation
+
+    Geometry Handling:
+    The geometry_type is the form of the objects in the given shapefile. It can be "boundingbox" or "point".
+    If geometry_type is set to "boundingbox" (default) then the bounding boxes in the shapefile will be used as is and transferred over
+    to the annotations file. However, if the geometry_type is "point" then a bounding box will be created in its place that's centered on
+    the original point with an apothem equal to buffer_size.
```
ethanwhite, line 257: Just for a little added specificity how about changing the last sentence to something like:

> If the geometry_type is "point" then a bounding box will be created for each point that is centered on the point location and has an apothem equal to buffer_size, resulting in a bounding box with dimensions of 2 times the value of buffer_size.

ethanwhite (changes requested):
This looks really good @JSpencerPittman! Thanks for the contribution!

Other than a wording suggestion (included in-line) I have two suggestions:

1. Let's add some argument checking code since `geometry_type` is now a string instead of a boolean. Given the structure of the code the easiest thing is probably an assert statement at the beginning on the function that checks that `geometry_type` is either `boundingbox` or `point` and throws an error informing that user that it needs to be one of those two if something else is passed.
2. My one other thought (and I realize I'm suggesting we consider a change to my own wording from #511), is to mirror pytorch's typical naming for "bounding boxes". A quick search suggest that "box"/"boxes" and "bbox" are both commonly used, so let's pick one of those for better consistency.

JSpencerPittman:
Just finished pushing the changes. The geometry_type should be more secure and understandable now.

deepforest/utilities.py
```diff
@@ -245,24 +245,37 @@ def xml_to_annotations(xml_path):
...
+    point that is centered on the point location and has an apothem equal to buffer_size, resulting in a bounding box with dimensions of 2 
+    times the value of buffer_size.
+    
     Args:
         shapefile: Path to a shapefile on disk. If a label column is present, it will be used, else all labels are assumed to be "Tree"
         rgb: Path to the RGB image on disk
         savedir: Directory to save csv files
         buffer_size: size of point to box expansion in map units of the target object, meters for projected data, pixels for unprojected data. The buffer_size is added to each side of the x,y point to create the box. 
-        convert_to_boxes (False): If True, convert the point objects in the shapefile into bounding boxes with size 'buffer_size'.  
+        geometry_type (bbox): Specifies the spatial representation used in the shapefile; can be "bbox" or "point"
```
ethanwhite, line 265: Let's drop the `(bbox)` here for consistency. We should probably do a general docstring consistency cleanup here, but the old `(False)` isn't consistent with any of the major styles so let's just drop it here since we're changing it.

ethanwhite (changes requested):
Looks fantastic @JSpencerPittman! We've got a style test failing. You can fix it by running:

```
yapf -i --recursive deepforest/ --style=.style.yapf
```

or manually making the change indicted in the tests:

```
-        raise ValueError(f"Invalid argument for 'geometry_type'. Expected 'point' or 'bbox'.")
+        raise ValueError(
+            f"Invalid argument for 'geometry_type'. Expected 'point' or 'bbox'.")
```

Since we need one more change anyway I've also noted one last docstring style change and then we can get this in.

Thanks for your work on this!

JSpencerPittman:
Alright, it should pass the style test now. I'm very new to Open Source Contributing. So, thank you for the help.

ethanwhite:
Thanks @JSpencerPittman! We're always happy to help out new contributors.

PR #532, opened 2023-10-24
Test jupyter notebooks

henrykironde:
Add Jupyter notebook to docs and test them

### DIFF
--- a/.github/workflows/Conda-app.yml
+++ b/.github/workflows/Conda-app.yml
@@ -38,7 +38,12 @@ jobs:
           cache-env: true
           cache-downloads: true
 
-      - name: Run pytest and style tests
+      - name: Run pytest, test style, and notebooks
         run: |
           pytest -v
           yapf -d --recursive deepforest/ --style=.style.yapf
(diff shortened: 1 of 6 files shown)


bw4sz:
Awesome. Will any notebook in /docs automatically be built? Will this slow down making readthedocs? For example, should I not put large example runs in there?

henrykironde:
Good questions @bw4sz. The documents are not run on readthedocs. They are just processed into HTML.
However, we test the notebooks to see if they are actually working [here](https://github.com/weecology/DeepForest/pull/532/files#diff-83665bc59dabaaa1aadfca6b5bfe42201be7209b1ea465055e2012f2ca9e90ecR45).

So we can skip the files that are heavy for GitHub actions since that is where we are doing the testing.
Right now, it tests any notebook in the `docs` using the wildcard. We can actually select the files that we want to test and skip the ones that may overload the testing platform.


`pytest --nbmake docs/*.ipynb`

or 

Something like this 

`pytest --nbmake docs/notebooks_test_on_action*.ipynb`

bw4sz (approved):
Thanks!

PR #520, opened 2023-10-16
warn and set projected to False if flip_y_axis is set to True, closes #440

bw4sz:
This is a small PR to add a warning and set projected=True in deepforest.utilities.boxes_to_shapefile when setting flip_y_axis=True.

PR #534, opened 2023-10-30
Raise a warning and confirm if a user sets projected to True when flip_y_axis is True 

### LINKED ISSUE #440: flip over y axis does not do anything in utilities.boxes_to_shapefiles if projected is not set to false.
The ongoing saga of where is 0,0 continues to be a headache. We have an awkward helper function in utilities.boxes_to_shapefile

https://github.com/weecology/DeepForest/blob/742e4b0ec681185cfad70971d31db91167fd1276/deepforest/utilities.py#L365

https://github.com/weecology/DeepForest/blob/742e4b0ec681185cfad70971d31db91167fd1276/deepforest/utilities.py#L424

When you set flip_y_axis, you need to also set projected is false. It should automatically do that. There is no use case where you would flip the y axis of projected data.

```
image_predictions = utilities.boxes_to_shapefile(
    df=image_predictions,
    flip_y_axis=True,
    projected=False,
    root_dir="/Users/benweinstein/Dropbox/Weecology/everglades_species/InspireRawData/Vacation/Vacation_03_20_2023/")
```

For example, this -> 
```
image_predictions = utilities.boxes_to_shapefile(
    df=image_predictions,
    flip_y_axis=True,
    root_dir="/Users/benweinstein/Dropbox/Weecology/everglades_species/InspireRawData/Vacation/Vacation_03_20_2023/")
```

does not honor the flag.

henrykironde:
fixes #440  #520

### DIFF
--- a/deepforest/utilities.py
+++ b/deepforest/utilities.py
@@ -388,6 +388,14 @@ def boxes_to_shapefile(df, root_dir, projected=True, flip_y_axis=False):
     Returns:
        df: a geospatial dataframe with the boxes optionally transformed to the target crs
     """
+    # Check if the user has set flip_y_axis, but not projected. Warn and confirm.
+
+    if (flip_y_axis and projected):
+        warnings.warn(
+            "flip_y_axis is {}, and projected is {}. In most cases, projected should be False when inverting y axis. Setting projected=False"
+            .format(flip_y_axis, projected), UserWarning)
+        projected = False
+
     plot_names = df.image_path.unique()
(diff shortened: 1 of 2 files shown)


henrykironde:
@bw4sz does this look good

bw4sz (commented):
In the test, doesn't that assert that the warning is always coming up. Why doesn't the test fail when flip_y_axis is True and projected is False. In this case no warning is raised, and the test should fail?

Issue #533, opened 2023-10-27
adding a way to assess model robustness


meilunzhou:
**Please describe a feature you would like to be added.**
I'd like a way to run the same training scheme numerous times in order to assess the model's robustness and get an understanding of the uncertainty within the model.

**Describe the solution you'd like**
I'd like a parameter that can be set to train the model X amount of times and have it return a mean and standard deviation of the evaluation metrics. 

**Describe alternatives you've considered**
Initially I restarted the kernel within one Jupyter Notebook script and ran the training multiple times while hand noting the evaluation values. Then when I tried to create a new model and rerun the training within a for loop instead of restarting the kernel, I would receive a "Weakly referenced object no longer exists" error message despite creating a new trainer. 

**Additional context**
This issue was addressed in the DeepForest FAQ documentation. A different naming for that section would make it more clear that some code bug issues reside there.

PR #537, opened 2023-11-05
Update FAQ.md

bw4sz:
Related to #533, make it clearer that there are known code workarounds in this section.


ethanwhite (commented):
I think "Trouble Shooting" is pretty standard terminology and clarifies that it's not something like "known bugs", but if you think the current version is better I'm also happy to merge it.

bw4sz:
adopted suggested change.

henrykironde (approved):

PR #544, opened 2023-11-07
Update use.md

bw4sz:
Another paper from the same author.


ethanwhite (approved):
Ah, yes, this is the one I was talking about with the drone thermal data. LGTM!

PR #550, opened 2023-11-09
Move main.predict_file to predict.predict_file and uses trainer.predict() for predict_file(). Speeds up main.evaluate!

### LINKED ISSUE #538: main.evaluate is too slow
It has nothing to do with predictions, it needs to be profiled, I'm guessing its the shapely matching function. This needs to be re-thought.

bw4sz:
1. Migrate the code in main.predict_file to predict.predict_file with just a wrapper in main. main.py is too long and shouldn't have any complex logic in it.
2. Used the trainer logic instead of manually moving batches to GPU, this causes significant speed up in evaluate, which should close #538. 

Before was around 37 seconds.

After
Screenshot 2023-11-09 at 11 30 16 AM

Tested on hpc with 1 gpu

```
(base) [b.weinstein@login12 ~]$ cat tunnel.sh
#!/bin/bash
#SBATCH --job-name=tunnel   # Job name
#SBATCH --mail-type=END               # Mail events
#SBATCH --mail-user=benweinstein2010@gmail.com  # Where to send mail
#SBATCH --account=ewhite
#SBATCH --nodes=1                 # Number of MPI ran
#SBATCH --cpus-per-task=1
#SBATCH --mem=70GB
#SBATCH --time=12:00:00       #Time limit hrs:min:sec
#SBATCH --output=/home/b.weinstein/logs/tunnel.out   # Standard output and error log
#SBATCH --error=/home/b.weinstein/logs/tunnel.err
#SBATCH --partition=gpu
#SBATCH --gpus=1
```
```
(base) [b.weinstein@login12 tests]$ cat profile_predict_file.py
#Profile the dataset class on gpu
from deepforest import main
from deepforest import get_data
import os
import pandas as pd
import numpy as np
import cProfile, pstats
import tempfile
from PIL import Image
import cv2

def run(m, csv_file, root_dir):
    predictions = m.predict_file(csv_file=csv_file, root_dir=root_dir)

if __name__ == "__main__":
    profiler = cProfile.Profile()
    profiler.enable()
    m = main.deepforest()
    m.use_release()
    m.config["workers"] = 0
    m.config["batch_size"] = 24

    csv_file = get_data("OSBS_029.csv")
    image_path = get_data("OSBS_029.png")
    tmpdir = tempfile.gettempdir()
    df = pd.read_csv(csv_file)

    big_frame = []
    for x in range(100):
        img = Image.open("{}/{}".format(os.path.dirname(csv_file), df.image_path.unique()[0]))
        cv2.imwrite("{}/{}.png".format(tmpdir, x), np.array(img))
        new_df = df.copy()
        new_df.image_path = "{}.png".format(x)
        big_frame.append(new_df)

    big_frame = pd.concat(big_frame)
    big_frame.to_csv("{}/annotations.csv".format(tmpdir))


    run(m, csv_file = "{}/annotations.csv".format(tmpdir), root_dir = tmpdir)
    profiler.disable()
    stats = pstats.Stats(profiler).sort_stats('cumtime')
    stats.print_stats()
    stats.dump_stats('predict_file.prof')
```

### DIFF
--- a/deepforest/callbacks.py
+++ b/deepforest/callbacks.py
@@ -63,7 +63,7 @@ class images_callback(Callback):
                   "skipping upload, images were saved to {}, "
                   "error was rasied {}".format(self.savedir, e))
 
-    def on_validation_epoch_end(self, trainer, pl_module):
+    def on_validation_end(self, trainer, pl_module):
         if trainer.sanity_checking:  # optional skip
             return
 
--- a/deepforest/main.py
+++ b/deepforest/main.py
@@ -370,7 +370,6 @@ class deepforest(pl.LightningModule):
 
     def predict_file(self, csv_file, root_dir, savedir=None, color=None, thickness=1):
         """Create a dataset and predict entire annotation file
-
         Csv file format is .csv file with the columns "image_path", "xmin","ymin","xmax","ymax" for the image name and bounding box position.
         Image_path is the relative filename, not absolute path, which is in the root_dir directory. One bounding box per line.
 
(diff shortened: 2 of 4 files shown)


ethanwhite (approved):
This all looks good. My one question is whether or not we could just import `predict_file()` directly from `predict.py` instead of having what is basically just a pass through function, but I'm guessing there's something special about the lightening module class that means we need to put a copy there, so I'm going to go ahead and merge.

Issue #559, opened 2023-12-01
Cannot set model.create_trainer(log_every_n_steps=) yields MisconfigurationException: You are trying to `self.log()` but the loop's result collection is not registered yet. 


aguirrejuan:
Hi, 

Thanks for the package. 

I am having the next problem

```
from pytorch_lightning.loggers import TensorBoardLogger
import os 

from deepforest import main
from deepforest import get_data

annotations_file = get_data("testfile_deepforest.csv")
logger = TensorBoardLogger(save_dir="lightning_logs/")

model = main.deepforest()
model.use_release()
model.config['gpus'] = '-1'
#model.config["workers"] = 4
model.config["train"]['epochs']= 5
model.config["score_thresh"] = 0.3
model.config["nms_thresh"] = 0.05
model.config["save-snapshot"] = False
model.config["train"]["csv_file"] = annotations_file
model.config["train"]["root_dir"] = os.path.dirname(annotations_file)
model.config["validation"]["csv_file"] = annotations_file
model.config["validation"]["root_dir"] = os.path.dirname(annotations_file)

#Create model trainer and fit model
model.create_trainer(logger=logger, log_every_n_steps=1)
model.trainer.fit(model)
```

Error

```
MisconfigurationException                 Traceback (most recent call last)
[<ipython-input-2-0635753caf65>](https://localhost:8080/#) in <cell line: 25>()
     23 #Create model trainer and fit model
     24 model.create_trainer(logger=logger, log_every_n_steps=1)
---> 25 model.trainer.fit(model)
     26 
     27 

12 frames
[/usr/local/lib/python3.10/dist-packages/pytorch_lightning/core/module.py](https://localhost:8080/#) in log(self, name, value, prog_bar, logger, on_step, on_epoch, reduce_fx, enable_graph, sync_dist, sync_dist_group, add_dataloader_idx, batch_size, metric_attribute, rank_zero_only)
    431         results = trainer._results
    432         if results is None:
--> 433             raise MisconfigurationException(
    434                 "You are trying to `self.log()` but the loop's result collection is not registered"
    435                 " yet. This is most likely because you are trying to log in a `predict` hook,"

MisconfigurationException: You are trying to `self.log()` but the loop's result collection is not registered yet. This is most likely because you are trying to log in a `predict` hook, but it doesn't support logging
```

bw4sz:
[...]

Issue #558, opened 2023-12-01
NameError: name 'a' is not defined 


aguirrejuan:
Hi, 

I have looked at the callback code,  and in nowhere the variable `a` is found,  maybe is `results`  instead?

https://github.com/weecology/DeepForest/blob/8144face5ed34509e1872aa482408437091da120/deepforest/callbacks.py#L104

Thanks.

ethanwhite:
Hi @aguirrejuan!  Thanks for the issue. Can you provide a reproducible example showing this error being generated so that we can look into it more?

bw4sz:
Yes, I see this, I find it strange that this crept in, I'm looking in git logs when this happened, it must have been very recently (last week?). The bigger question is how none of our tests caught it. Pushing a quick fix now, but then we need to open an issue on why the tests didn't cover this. Nothing obvious.

PR #560, opened 2023-12-01
add tensorboard logger test and fix callback typo

bw4sz:
Closes the #559 and #558 , there was a typo in callbacks only for the create_trainer(logger=), function. I added a test, but it really should have been caught somewhere else. This is high priority.

### DIFF
--- a/deepforest/callbacks.py
+++ b/deepforest/callbacks.py
@@ -101,7 +101,7 @@ class iou_callback(Callback):
             self.log("box_recall", results["box_recall"])
             self.log("box_precision", results["box_precision"])
 
-            if isinstance(a, pd.DataFrame):
+            if isinstance(results, pd.DataFrame):
                 for index, row in results["class_recall"].iterrows():
                     self.log(
                         "{}_Recall".format(pl_module.numeric_to_label_dict[row["label"]]),
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -18,6 +18,8 @@ from deepforest import get_data
 from deepforest import dataset
 from pytorch_lightning import Trainer
 from pytorch_lightning.callbacks import Callback
+from pytorch_lightning.loggers import TensorBoardLogger
+
 from PIL import Image
 
 #Import release model from global script to avoid thrasing github during testing. Just download once.
@@ -101,6 +103,22 @@ def big_file():
 def test_main():
     from deepforest import main
 
+def test_tensorboard_logger(m, tmpdir):
+    annotations_file = get_data("testfile_deepforest.csv")
+    logger = TensorBoardLogger(save_dir=tmpdir)
+    m.config["train"]['epochs']= 5
+    m.config["score_thresh"] = 0.3
+    m.config["nms_thresh"] = 0.05
+    m.config["save-snapshot"] = False
+    m.config["train"]["csv_file"] = annotations_file
+    m.config["train"]["root_dir"] = os.path.dirname(annotations_file)
+    m.config["validation"]["csv_file"] = annotations_file
+    m.config["validation"]["root_dir"] = os.path.dirname(annotations_file)
+
+    #Create model trainer and fit model
+    m.create_trainer(logger=logger, check_val_every_n_epoch=1)
+    m.trainer.fit(m)
+    
 def test_use_bird_release(m):
     imgpath = get_data("AWPE Pigeon Lake 2020 DJI_0005.JPG")    
     m.use_bird_release()


ethanwhite:
Everything looks good. Will approve once the tests pass.

bw4sz:
I think there is still something hapenning with the tensorboard logger and the trainer object, I have a meeting until 1pm, but i'll have look this afternoon.

ethanwhite:
Sounds good. I'll leave this for now. Ping me when you're ready for a review.

bw4sz:
This turned out to be a large issue and definitely warranting of 1.3.0. Our tests did not cover the logger, we had discontinued it because of the annoyances of mocking the comet_ml experiment. This turns out to be tangential to the goal, but hid the actual problem, which was quite huge. The problem is upstream of any given logger, and with pytorch lightning trainer.predict methods. 

In https://github.com/weecology/DeepForest/commit/f71eef4e8c09f4031bb8be7276f551982b246e2f, I changed the predict_file methods to user trainer.predict. The virtue of this was that it made prediction much faster. We used to have to do alot of annoying workarounds, like not run evaluation that often because that function was so slow. What we didn't see was that when using a logger, like comet_ml or tensorboard, you cannot mix trainer.predict inside of an evaluation_hook. I have documented this on pytorch lightning repo. See lots of relevant conversation.

https://github.com/Lightning-AI/pytorch-lightning/issues/19101

The bad news means that training with a logger has been dead for three weeks and we didn't know.
The good news is that with every error, we find ways to make the whole API better. I have refactored the evaluate routine to include a new wrapper that directly uses the predictions made during pytorch lightnings hooks, instead of re-running them separately. This will be faster. It also allowed me to move code out of main.py and into other modules, which is a long term goal to making the API more flexible, main.py should really just have wrappers and not custom code. During this I was then able to remove quite a bit of duplicate code.

bw4sz:
i'm tempted to force push this, but I suppose it can wait until Monday. Once @ethanwhite merges I will bumpversion minor and push to pypi.

deepforest/predict.py
```diff
@@ -140,10 +140,12 @@ def across_class_nms(predicted_boxes, iou_threshold=0.15):
     return new_df
 
 
-def predict_file(model,
-                 csv_file,
-                 root_dir,
+def predict_file(trainer,
+                 model,
+                 dataloader,
```
ethanwhite, line 145: I found this change a little confusing. `main.predict_file()` and `predict.predict_file()` now have different arguments and `predict.predict_file()` doesn't take a file as an argument. Assuming that the trainer and dataloader move was necessary as part of the broader refactor then I think we should consider renaming `predict.predict_file()` to make this easier to follow. `predict_single_file_dataloader` feels clunky, but something in this space to help communicate what's happening?

deepforest/predict.py
```diff
@@ -154,27 +156,18 @@ def predict_file(model,
 
     Args:
         model: deepforest.main object
-        csv_file: path to csv file
+        trainer: a pytorch lightning trainer object
+        dataloader: pytorch dataloader object
```
ethanwhite, line 160: Order should match arguments in function definition

deepforest/predict.py
```diff
@@ -154,27 +156,18 @@ def predict_file(model,
...
         root_dir: directory of images. If none, uses "image_dir" in config
         nms_thresh: Non-max supression threshold, see config["nms_thresh"]
+        df: pandas dataframe with bounding boxes, label and scores for each image in the csv file
```
ethanwhite, line 163: This isn't provided as an argument and matches the previous description for the return value. Seems like this shouldn't be here?

deepforest/main.py
```diff
@@ -382,13 +377,23 @@ def predict_file(self, csv_file, root_dir, savedir=None, color=None, thickness=1
...
+                                 root_dir=root_dir,
+                                 transforms=None,
+                                 train=False)
+        dataloader = self.predict_dataloader(ds)
+
         results = predict.predict_file(model=self,
-                                       csv_file=csv_file,
+                                       trainer=self.trainer,
+                                       annotations=df,
+                                       dataloader=dataloader,
```
ethanwhite, line 390: I'm assuming there's a reason related to the overarching PR for moving the dataloader and trainer creation, presumably related to handling the trainer? On the off chance there isn't I think it's simplest to move it back into `predict_file` (see main comment above)

deepforest/predict.py
```diff
@@ -154,27 +156,18 @@ def predict_file(model,
 
     Args:
         model: deepforest.main object
-        csv_file: path to csv file
+        trainer: a pytorch lightning trainer object
+        dataloader: pytorch dataloader object
         root_dir: directory of images. If none, uses "image_dir" in config
```
ethanwhite, line 161: Missing docs for new `annotations` argument

ethanwhite (changes requested):
Looks really good. My one meaningful thing is that I found the fact that `predict.predict_file()` doesn't work directly on a file confusing. It sent me on a couple of tangents while reviewing which might suggest it could lead to confusing maintenance. Other than potentially renaming that there are a few docstring issues to cleanup and I think this should be ready to go it.

deepforest/predict.py
bw4sz: A couple thoughts. The original function didn't do a good job isolating from main.DeepForest class methods, before you were literally passing the model object to use them to generate a dataloader.

https://github.com/weecology/DeepForest/blob/327ceb9094ec649f54822d88406b492c65b6851c/deepforest/predict.py#L173C1-L174C1

This doesn't make alot of sense. We want main.py to handle all within-class logic, and then hand off to predict.predict_* for tile, image or file prediction which is isolated from the class methods. I'm looking through docs right now, but I think the use should never use predict.predict_* methods directly. I think renaming is a good option 

predict.predict_tile = predict._predict_tile_ to highlight that these are internal function following 'private' names. https://peps.python.org/pep-0008/. I'm fine with renaming this predict._predict_dataloader_.

deepforest/main.py
bw4sz: We want to isolate predict.predict_* methods from main.predict_* methods. If you have a csv file as the argument, you need to pass the class to the function to generate the dataloader, see above comment.

deepforest/predict.py
ethanwhite: Thanks for the explanation. Sounds good to me.

deepforest/main.py
ethanwhite: :+1:

deepforest/predict.py
bw4sz: just noticed that I don't know how to escape underscores in github markdown "predict._predict_image".

ethanwhite (changes requested):
A couple more small cleanups and then I'll get this merged.

deepforest/main.py
```diff
@@ -384,7 +371,7 @@ def predict_file(self, csv_file, root_dir, savedir=None, color=None, thickness=1
                                  train=False)
         dataloader = self.predict_dataloader(ds)
 
-        results = predict.predict_file(model=self,
+        results = predict._predict_a_dataloader_(model=self,
```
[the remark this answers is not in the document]
bw4sz: Yes, I saw that too, but main also have main.predict_dataloader, which is a reserved method for pytorch lightning.

https://github.com/weecology/DeepForest/blob/327ceb9094ec649f54822d88406b492c65b6851c/deepforest/main.py#L287

But has nothing to do with the main.predict_file workflow. Part of the goal is to improve readability by reduce overlap in main.predict* and predict._predict* naming. I'll change it to predict._predict_single_dataloader_

deepforest/predict.py
```diff
@@ -8,51 +8,44 @@
 
 import torch
 from torchvision.ops import nms
+import typing
 
 from deepforest import preprocess
```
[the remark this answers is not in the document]
bw4sz: removed unused modules.

PR #569, opened 2023-12-07
Introduces 'path' as an alternative to 'raster_path' with a deprecation warning

manascb1344:


ethanwhite:
Thanks for the contribution @manascb1344. I'm going to close this PR in favor of #617 since this one just adds the argument but doesn't make all of the downstream changes to make that new argument work. Those changes are all completed in #617 if you want to see a complete set of changes for this issue looks like.

Issue #566, opened 2023-12-07
Predicted_raster doesn't load


matismatis:
Hello everyone,

I am a forestry student, and trying to learn python for remote sensing/GIS. This is one of my first projects with python, so I am sorry if this is a a simple issue.

I am trying to make ''deepforest'' run with Anaconda Powershell Prompt (miniconda3) and Jupyter Notebook, using Windows 11 Home version 22h2. 
I installed ''deepforest'' and associated packages on the powershell prompt, and tried running the code in Jupyter. My first TIFF file consists of 4 TIF rasters that I merged in QGIS. It could not be read/recognized by the code, so I tried running an individual TIF file. The file could be read, and it gave me a number of detections. However, I could not open the predicted_raster, so I can't verify the output. 

I have uploaded the code and the response in 2 separate text files. Sorry if this is not how you're supposed to attach them, kind of new to coding and such. I tried uploading the tif that I use, but I am unable to do so, my apologies.
[deepforest_code_jupyter.txt](https://github.com/weecology/DeepForest/files/13598893/deepforest_code_jupyter.txt)
[response_deepforest_code.txt](https://github.com/weecology/DeepForest/files/13598895/response_deepforest_code.txt)

Later on I tried different ways to open the predicted_raster.
First I verified it using:
print(predicted_raster)

The response that I got was:

[[ 67  56  44]
 [ 66  56  44]
 [ 64  54  43]
 ...
 [126 145 157]
 [119 140 153]
 [120 141 155]]

Then I tried to open it with:
plt.imshow(predicted_raster, vmin=0, vmax=255)
plt.show()

The response I got was:
<img width="229" alt="predicted_rasterwrong" src="https://github.com/weecology/DeepForest/assets/117834607/b723bc84-0fd3-4a0d-9acf-8eb3a1de2aa6">

I also tried:
plt.imshow(predicted_raster, aspect='auto')
plt.show()

And got:
<img width="442" alt="predicted_rasterwrong2" src="https://github.com/weecology/DeepForest/assets/117834607/534d4c53-a9c2-4941-b035-8d1f8d0a1b50">

Finally I tried:
plt.plot(predicted_raster[:, 0], label='Intensity')
plt.legend()
plt.show()

And got:
<img width="425" alt="predicted_rasterwrong3" src="https://github.com/weecology/DeepForest/assets/117834607/3fb09b3a-78dc-45f5-9381-dbef8663d7ec">

[...]

PR #570, opened 2023-12-08
Add a plotting code snippet for predict_tile to getting_started page.

bw4sz:
Everytime a user is confused about something I try to make some changes to the docs. 

In 

https://github.com/weecology/DeepForest/issues/566

user tried to do

```
# Perform prediction on the tile
result = model.predict_tile(raster_path, return_plot=True, patch_size=patch_size, patch_overlap=patch_overlap)

# Extract the predicted raster and boxes from the result
predicted_raster = result[0]
predicted_boxes = result[1]

# Load the original image for visualization
original_image = plt.imread(raster_path)

# Display the original image
plt.imshow(original_image)

# Plot the predicted bounding boxes on the original image
for _, box in predicted_boxes.iterrows():
    xmin, ymin, xmax, ymax = box["xmin"], box["ymin"], box["xmax"], box["ymax"]
    rect = plt.Rectangle((xmin, ymin), xmax - xmin, ymax - ymin, linewidth=1, edgecolor="r", facecolor="none")
    plt.gca().add_patch(rect)
```

So I made it clearer in getting started that just the image is return in ```predict_tile(return_plot=True)```, and we can call matplotlib directly.


bw4sz:
Closed for a new clean PR #573

Issue #546, opened 2023-11-08
Better document main(config_args)


bw4sz:
The doc page for the config does not mention that you can pass args directly to deepforest.main(config_args={"batch_size":10}). This is useful in iterating through params.

bw4sz:
I'm actually concerned that the config_args are not properly being set. We need better tests, this is a priority I will address soon.

bw4sz:
From deeptreeattention

```
if args.m:
    hot_config_fix = json.loads(args.m)
    for key, value in hot_config_fix.items():
        print("setting config {} to {}".format(key, value))
        config[key] = value
```

Issue #571, opened 2023-12-08
Multi class detection


jyo94:
Thank you for your response. Apologise for not clearly specifying my problem. I will try to rephrase it,

I want to built a multi class detection model using deepforest. I have 2 labels ( "cassava" and "weed"). I didn't get a proper idea (after reading the documentation) on how to change the config, to make it a multi class model. So, I made 2 models, one for detecting cassava and the other for weed. 
In both the models, I changed the label to tree as shown below and removed the other label,
```
crop_dir = os.path.join(os.getcwd(),f"train_data_folder/train_data_folder_{image_name.split('.')[0]}")
train_file= os.path.join(crop_dir,"train.csv")
train_df['label'] = train_df['label'].replace({'Weed':'Tree'})
train_df_mod = train_df[train_df['label'] != 'Cassava']
train_df_mod.to_csv(os.path.join(crop_dir,"train_mod_weed.csv"))
```
And used this file in the config settings,
<img width="433" alt="image" src="https://github.com/weecology/DeepForest/assets/146371475/0b17e3cc-c78b-4719-8454-2de220574903">

And have saved this model, 
`m.trainer.save_checkpoint("{}/checkpoint_weed.pl".format(tmpdir))`

Next loaded the saved model using "load_from_checkpoint" function and predicted for a new image using "predict_tile" function. The result I'm getting is the below screenshot,

![sample_output_1 (1)](https://github.com/weecology/DeepForest/assets/146371475/3f8785f8-5d47-4910-8611-48cec0faf7e3)

Now the issues are,
1. All objects are detected as Tree
2. All objects (cassava and weed) are detected though I loaded cassava checkpoint model

I have no idea on how to train for a custom data. And also how to approach a multi class problem.

Hope I have cleared in explaining my issue this time.

jyo94:
Here is the link to the single class detection, ([colab link](https://colab.research.google.com/drive/1XdjvYzqH1aTWPU345J4CB8ZGvn4ozExy?usp=sharing))

jyo94:
@bw4sz @ethanwhite @bendmorris @easz @fluby 
Requesting the developers to help in this matter

ethanwhite:
A couple of general suggestions for helping us help you as quickly as  possible:

1. Each issue should be limited to one topic. So, your question about the two colored boxes in the single class detection should be in a separate issue from your questions about multi-class detection.
[...]

PR #586, opened 2023-12-14
update config args and add to docs

bw4sz:
Closes https://github.com/weecology/DeepForest/issues/546 and addresses in part #571

### DIFF
--- a/deepforest/main.py
+++ b/deepforest/main.py
@@ -70,7 +70,7 @@ class deepforest(pl.LightningModule):
                         "Config argument {} not found in config file".format(key))
                 if type(value) == dict:
                     for subkey, subvalue in value.items():
-                        print("setting config {} to {}").format(subkey, subvalue)
+                        print("setting config {} to {}".format(subkey, subvalue))
                         self.config[key][subkey] = subvalue
                 else:
                     print("setting config {} to {}".format(key, value))
--- a/deepforest/main.py
+++ b/deepforest/main.py
@@ -55,19 +55,34 @@ class deepforest(pl.LightningModule):
 
         print("Reading config file: {}".format(config_path))
         self.config = utilities.read_config(config_path)
-        self.model = model
-
-        # release version id to flag if release is being used
-        self.__release_version__ = None
-
         self.config["num_classes"] = num_classes
-        self.create_model()
-
         # If num classes is specified, overwrite config
         if not num_classes == 1:
             warnings.warn(
                 "Directly specifying the num_classes arg in deepforest.main will be deprecated in 2.0 in favor of config_args. Use deepforest.main(config_args={'num_classes':value})"
             )
+
+        # Update config with user supplied arguments
+        if config_args:
+            for key, value in config_args.items():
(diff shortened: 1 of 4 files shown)


ethanwhite:
Looks like there's a meaningful test failure on Azure

bw4sz:
> Looks like there's a meaningful test failure on Azure

just fixed it. typo in the model name.

docs/ConfigurationFile.md
```diff
@@ -47,6 +47,24 @@ validation:
     iou_threshold: 0.4
     val_accuracy_interval: 20
 ```
+## Passing config arguments at runtime using a dict
+
+It can often be useful to pass config args directly to a model instead of editing the config file. By using a dict with that matches with the config keys, main.deepforest will update the config after reading from file.
```
ethanwhite, line 52: I found the second sentence confusing. How about:

"By using a dict containing the config keys and their values. Values provided in this dict will override values provided in deepforest_config.yml."

docs/ConfigurationFile.md
```diff
@@ -47,6 +47,24 @@ validation:
...
+
+```
+m = main.deepforest()
+assert not m.config["num_classes"] == 2
```
ethanwhite, line 56: I think the use of `assert` statements here could be be confusing for some users. Any reason not to just show the result with `m.config["num_classes"]` and equivalent throughout?

ethanwhite (changes requested):
Looks good. I've me a couple of comments requesting clarification.

Can you also update the multi-class model docs to reflect the new usage?

deepforest/main.py
```diff
@@ -109,7 +124,7 @@ def use_release(self, check_release=True):
             check_release=check_release)
         if self.config["architecture"] != "retinanet":
             warnings.warn(
-                "The config file specifies architecture {}, but the release model is torchvision retinanet. Reloading with deepforest.main with a retinanet model"
+                "The config file specifies architecture {}, but the release model is torchvision retinanet. Reloading with main.deepforest with a retinanet model"
```
ethanwhite, line 127: I think there's an extra "with" in here maybe?

docs/ConfigurationFile.md
```diff
@@ -47,6 +47,26 @@ validation:
...
     val_accuracy_interval: 20
 ```
+## Passing config arguments at runtime using a dict
+
+It can often be useful to pass config args directly to a model instead of editing the config file. By using a dict containing the config keys and their values. Values provided in this dict will override values provided in deepforest_config.yml.
+
+```
+# Default model has 1 class
+m = main.deepforest()
+not m.config["num_classes"] == 2
```
ethanwhite, line 57: What I mean here was to drop all of the equality stuff and just now `m.config["num_classes"]`. That will first show that there is 1 class and then after setting it that there are 2 classes. Basically making this make sense to the average reader instead of structuring the docs like the tests. Does that make sense?

ethanwhite (changes requested):
Two last little cleanup things

ethanwhite (approved):

PR #589, opened 2023-12-21
when the CRS of the shapefile and RGB data don't match, don't proceed…

bw4sz:
I found that we just proceed with converting data even if it obviously doesn't overlap. We should expect the user to have proper georeferenced data. I also cleaned up the annotations_to_shapefile function even though its going to be deprecated, it had an obvious edge that users could run into.


ethanwhite (approved):
LGTM!

Issue #460, opened 2023-07-13
Integrate segment anything for post model bounding box to polygons.


bw4sz:
https://segment-anything.com/ was a new transformer object detection model. @MarconiS showed good results for bounding box to polygon workflows. The main problem was it was too slow. There is a new CNN module that claims similar performance at 50x speed. This needs to be investigated. 
 
https://arxiv.org/pdf/2306.12156.pdf

![Screenshot from 2023-07-05 21-04-27 (1)](https://github.com/weecology/DeepForest/assets/1208492/57a4877c-0f85-4034-9c40-8a8a2a5ce5df)

# Next steps.

[ ] @MarconiS can you give us links and a couple code snippets to assess next steps for getting this into main branch.
[ ] Determine how to evaluate results given bounding box test data we have on hand
[ ] Identify potential workflow: after prediction (predict_image or predict_tile), take bounding box coordinates as prompts for segment anything. Do we use the 4 vertexes of the bounding box, or the centroids of the boxes?
[ ] Add fastSA model weights to git download. Use for either bird or tree release.

MarconiS:
I am off the grid the rest of the week but I am definitively  giving a crack at it next week. I've got the function for it up and running on the [GeotreeHealth](https://github.com/MarconiS/GEOtreehealth) repo. 
Just need to refactor the function to work directly with deepForest and allow for either SAM or fastSAM (fastSAM  seems it is already integrated with transformers library, which is a plus: https://huggingface.co/An-619/FastSAM)

It is slow, but it is not overwhelming slow: the good news is that

PR #590, opened 2023-12-22
DeepForest support for multiple annotation types (point, box, polygon).

bw4sz:
We want to prepare DeepForest to be able to ingest annotation geometries covering point, box and polygon formats. This is the first step to allow models and predictions in each and among each format. I think this is crucial for accessing every possible dataset available. This was spurred during development of the MillionTrees benchmark. Here is a roadmap

Pros:
- More flexible
- Better data coverage
- Opportunities for weak annotation learning
- More general

Cons:
- We will need to update docs and transition the fundamental annotation holder from pandas to geopandas. This is pretty fundamental, many functions just reach for columns 'xmin', 'xmax', 'ymin' 'ymax'.  For example:

https://github.com/weecology/DeepForest/blob/b87f9314e666903bb361c614e24daf53e5abad03/deepforest/predict.py#L112C16-L112C16

https://github.com/weecology/DeepForest/blob/b87f9314e666903bb361c614e24daf53e5abad03/deepforest/predict.py#L69

and many more. We will need to first check what geometry type we have and perform the correct transformations. 

- There will be a breaking change in data formats that will ask users to update the default csv file if they have been doing something custom and bypassing the loader. For most users we can make this a somewhat easy transition by automatically converting within functions when we see a xmin, xmax, ymin, and ymax column. Most users will not notice this. 

## Roadmap for merge

- [x] Read in shapefiles of point, box, and polygon annotations and format them as geopandas dataframes.
- [x] preprocess.split_raster needs to have multi-geometry logic
- [x] conversion utilities to move between geometry types.

## Long term goals
- a backbone tree and bird model, or atleast a model class that natively supports points and polygons. This connects to many many issues on this repo, for example #460


bw4sz:
@henrykironde and @ethanwhite I invite thoughts on how we proceed. My plan is to work on this branch while developing MillionTrees dataset https://milliontrees.idtrees.org/ and when I feel like the kinks are worked out, trigger the PR. This would be enough for DeepForest 1.4.0.

bw4sz:
Add better release notes.

henrykironde (approved):

henrykironde:
This looks good in terms of logic. There are a few changes needed, but I will handle them later as part of a general cleaning and update.

Issue #594, opened 2024-01-04
Take in data_loader


bw4sz:
There is a general shift towards providing data not as on disk .csv or .json files, but as pytorch data loaders.

For example in torch geo

https://torchgeo.readthedocs.io/en/stable/api/datasets.html

or the Wilds benchmark

https://github.com/p-lambda/wilds?tab=readme-ov-file

```
from wilds import get_dataset
from wilds.common.data_loaders import get_train_loader
import torchvision.transforms as transforms

# Load the full dataset, and download it if necessary
dataset = get_dataset(dataset="iwildcam", download=True)

# Get the training set
train_data = dataset.get_subset(
    "train",
    transform=transforms.Compose(
        [transforms.Resize((448, 448)), transforms.ToTensor()]
    ),
)

# Prepare the standard data loader
train_loader = get_train_loader("standard", train_data, batch_size=16)

# (Optional) Load unlabeled data
dataset = get_dataset(dataset="iwildcam", download=True, unlabeled=True)
unlabeled_data = dataset.get_subset(
    "test_unlabeled",
    transform=transforms.Compose(
        [transforms.Resize((448, 448)), transforms.ToTensor()]
    ),
)
unlabeled_loader = get_train_loader("standard", unlabeled_data, batch_size=16)

# Train loop
for labeled_batch, unlabeled_batch in zip(train_loader, unlabeled_loader):
    x, y, metadata = labeled_batch
    unlabeled_x, unlabeled_metadata = unlabeled_batch
    ...
```

DeepForest should be able to take in a train_dataloader and val_dataloader and then reassign them here:

https://github.com/weecology/DeepForest/blob/ee9ee5d8c9a6dfcb3ed02d519dae797cbf40b1dd/deepforest/main.py#L264

We will need sanity checks to make sure the dataloaders are formatted as expected.

Om-Doiphode:
@bw4sz, if I am correct, we are expected to change the following methods by removing the dependency on csv or json file right?
![image](https://github.com/weecology/DeepForest/assets/83624310/b247b59c-90ae-4545-986f-8678b1ad467c)

So instead of passing csv files, the user should pass the custom dataloader? I guess we need to change the structure of the whole `deepforest` class for this. Is this the expected outcome?

PR #595, opened 2024-01-10
Modified train and val dataloaders

Om-Doiphode:
Fixes issue #594


bw4sz:
I appreciate the effort, but closing in favor of #602 . To make a PR, we need to have unit tests that assert desired functionality and a fuller explanation of its behavior documented.

Issue #584, opened 2023-12-13
savedir is not documented in evaluate.


bw4sz:
It can be really helpful to write a set of plots to see how evaluation boxes overlap. We have a method that is currently hidden by using savedir to save plots to a directory with annotations overlaid.

https://deepforest.readthedocs.io/en/latest/_modules/deepforest/evaluate.html#evaluate

Its here:
https://github.com/weecology/DeepForest/blob/04a108483160242e2c68a57673864e0344f03f8e/deepforest/evaluate.py#L25

but not here:

https://github.com/weecology/DeepForest/blob/04a108483160242e2c68a57673864e0344f03f8e/deepforest/evaluate.py#L140

Om-Doiphode:
Can you please elaborate more on what needs to be done for this issue?

Om-Doiphode:
Hi @bw4sz, please review the pull request. Thanks!

ethanwhite:
Closed by d228d83aeb49dcd97b86b93af95fd8644b96e2ef. Thanks again for your work on this @Om-Doiphode!

PR #596, opened 2024-01-14
Added doc string for save_dir in evaluate function

Om-Doiphode:
Fixes issue #584


henrykironde (approved):
Thanks @Om-Doiphode for the changes. Looks good to me

Issue #516, opened 2023-10-15
boxes_to_shapefile should take in absolute path to rgb.


bw4sz:
Issue (#368) highlights that the boxes_to_shapefile is looking for an image_path column. This isn't flexible enough. 

Instead we want a 'rgb' argument that is the full path. Use this test:

```
from deepforest import main
from deepforest import get_data
from deepforest import utilities
import os
import rasterio as rio
import numpy as  np

raster_path = get_data("OSBS_029.tif")
r = rio.open(raster_path)
df = r.read()

rolled_df = np.rollaxis(df, 0,3)

m = main.deepforest()
m.use_release()

boxes = m.predict_tile(image=rolled_df, patch_size=1500)

shp = utilities.boxes_to_shapefile(boxes, rgb=raster_path, projected=True)
shp.to_file("~/Downloads/boxes.shp")
```

https://github.com/weecology/DeepForest/blob/b612d85bbc9d3f31749653a6f27623a93e6a049a/deepforest/utilities.py#L365

Om-Doiphode:
Hi @bw4sz, please review my pull request. Thanks!

ethanwhite:
A couple of possible ways to avoid or mitigate this being a breaking change older uses of the package:

1. One option when deprecating positional args is to replace the function with one with a new name and then deprecate the old function. This might actually be the best solution here since then we wouldn't be rushed into 2.0. Possible names include `reproject_boxes`, `boxes_to_projection`, `boxes_to_original_projection`.
2. This also relates to #638 and #608. So one option is to move the officially supported behavior to inside the relevant functions with an option argument to maintain projection when present. We then deprecate the function instructing the user to use the optional argument to the relevant function (and since this is now internal we avoid  future issues with user facing changes).
3. If (1) and (2) don't work for some reason then we could at least check the new `rgb` argument to see if it includes a filename. If it doesn't then perform the previous behavior and throw a deprecation warning. This would at least prevent breaking code where `root_dir` was used positionally.

PR #597, opened 2024-01-14
Modified boxes_to_shapefile function

Om-Doiphode:
Fixes Issue #516: 

The following modifications have been made:

- Updated the `boxes_to_shapefile` function in `utilities.py` to use the 'rgb' argument instead of 'image_path'.
- Modified the test functions in `test_utilities.py` to ensure the correct functionality of the modified `boxes_to_shapefile` function.

### DIFF
--- a/deepforest/utilities.py
+++ b/deepforest/utilities.py
@@ -382,13 +382,13 @@ def check_image(image):
                          "found image with shape {}".format(image.shape))
 
 
-def boxes_to_shapefile(df, root_dir, projected=True, flip_y_axis=False):
+def boxes_to_shapefile(df, rgb, projected=True, flip_y_axis=False):
     """
     Convert from image coordinates to geographic coordinates
     Note that this assumes df is just a single plot being passed to this function
     Args:
        df: a pandas type dataframe with columns: name, xmin, ymin, xmax, ymax. Name is the relative path to the root_dir arg.
-       root_dir: directory of images to lookup image_path column
+       rgb: path for the image file
        projected: If True, convert from image to geographic coordinates, if False, keep in image coordinate system
        flip_y_axis: If True, reflect predictions over y axis to align with raster data in QGIS, which uses a negative y origin compared to numpy. See https://gis.stackexchange.com/questions/306684/why-does-qgis-use-negative-y-spacing-in-the-default-raster-geotransform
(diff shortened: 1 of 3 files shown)


bw4sz:
We just need to update docs and examples to match. Quick search found 

https://deepforest.readthedocs.io/en/latest/annotation.html#how-can-i-view-current-predictions-as-shapefiles

ethanwhite:
I like this change, but note that it is backwards breaking (including for some of the lab's code). Ideally once we merge this the next release should be 2.0. A couple of possible ways to avoid needing to do this:

1. One option when deprecating positional args is to replace the function with one with a new name and then deprecate the old function. This might actually be the best solution here since then we wouldn't be rushed into 2.0. Possible names include `reproject_boxes`, `boxes_to_projection`, `boxes_to_original_projection`.
2. This also relates to #638 and #608. So one option is to move the officially supported behavior to inside the relevant functions with an option argument to maintain projection when present. We then deprecate the function instructing the user to use the optional argument to the relevant function (and since this is now internal we avoid  future issues with user facing changes).
3. If (1) and (2) don't work for some reason then we could at least check the new `rgb` argument to see if it includes a filename. If it doesn't then perform the previous behavior and throw a deprecation warning. This would at least prevent breaking code where `root_dir` was used positionally.

Thoughts?

ethanwhite (changes requested):
See my comment about backwards compatibility and alternative approaches before merging this.

ethanwhite:
@bw4sz - can you response to https://github.com/weecology/DeepForest/pull/597#issuecomment-2030852489 when you have a minute so that we can get this one moving forward

bw4sz:
I don't think we are ready for this change. There is a geo-type branch that will hold points, polygons, and boxes as part of geodataframe objects. We should wait until I finish that branch this summer.

ethanwhite:
OK. Thanks @bw4sz. Given that I'm going to close this PR and I've assigned #516 to you so that folks know not to work on it further until you say so. I'm going to move the text of my comment over there.

ethanwhite:
Thanks for the contribution @Om-Doiphode - we just weren't quite ready for this one yet. Once @bw4sz lets us know this is ready to proceed we can chat about coming back to it.

PR #599, opened 2024-01-16
Fixed error related to tile assignment

Om-Doiphode:
Fixes issue #427

### DIFF
--- a/deepforest/main.py
+++ b/deepforest/main.py
@@ -472,6 +472,8 @@ class deepforest(pl.LightningModule):
                 # Draw predictions on BGR
                 if raster_path:
                     tile = rio.open(raster_path).read()
+                else:
+                    tile = self.image
                 drawn_plot = tile[:, :, ::-1]
                 drawn_plot = visualize.plot_predictions(tile,
                                                         results,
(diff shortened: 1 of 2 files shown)


ethanwhite:
Thanks for this fix @Om-Doiphode! Do you have time to add a test associated with the change to avoid a regression in the future?

Om-Doiphode:
Hi @ethanwhite, can you guide me on how to write a test for this change? Because I have looked at the `deepforest/tests/test_main.py` file and I can't understand on how to write a test for the current issue. Like how the `tile` variable can be accessible in another file (`deepforest/tests/test_main.py` file). Should I write some assert condition within the `predict_tile` function in the `deepforest/main.py`?

ethanwhite:
I think we basically just need a copy of the current `test_predict_tile_from_array` function:

https://github.com/weecology/DeepForest/blob/54288a62753de4df0919884b3c40eb88258a9aca/tests/test_main.py#L248-L257

But with `return_plot = TRUE`. We need to change the assert statement a little since we're getting back a numpy ndarray. We could use `isinstance(prediction, np.ndarray)` to check the type of the returned value or check its size to make sure it's not empty `np.size(prediction) > 0`.

Does that help?

Om-Doiphode:
HI @ethanwhite, Thanks for the explanation! I have made the necessary changes. Please review, thanks!

ethanwhite (approved):
Looks great @Om-Doiphode! Thanks for the fix and the test!

PR #601, opened 2024-01-31
Make the Getting Started example executable

ethanwhite:
The first thing most folks want to do on a Getting Started page is to run the example so this swaps out the local path for one present on all installs.

Also moves the `os` import down to where it is used to keep things simpler.


bw4sz (commented):
looks good.

bw4sz:
Thanks.

bw4sz (approved):

PR #602, opened 2024-02-06
Optionally allow existing dataloaders for training, evaluation and prediction

### LINKED ISSUE #594: Take in data_loader

bw4sz:
- [X] Existing Train, val and predict dataloaders can be passed into m.trainer
- [X] Tests to assert that config arguments can be correctly changed.  
- [x] Update docs with examples. 

Closes #594 

This PR allows for existing dataloaders to be used, instead of the user needing to start from a csv file. This is part of the broader effort to create better interoperability between deepforest and other resources. For MillionTrees, we are copying the WILDS benchmark (https://github.com/p-lambda/wilds), which yields dataloaders directly.


bw4sz:
Okay @henrykironde this is ready for review, style changes just pushed.

henrykironde (approved):

Issue #576, opened 2023-12-12
split raster without annotation file


jyo94:
I found while training splitting the image into smaller tiles and predicting gives good prediction results. Now for a new image which doesn't have respective annotation file, is it possible to split_raster function. If not, what is the function for this senerio?

ethanwhite:
We typically handle this process using `predict_tile()`, which doesn't require an annotations file, does the raster splitting for you, and returns everything reassembled https://deepforest.readthedocs.io/en/latest/getting_started.html#predict-a-tile

Is there a reason you need to use `split_raster()` directly?

ethanwhite:
Closing for now. @jyo94 - please feel free to reopen if you still have questions/issues accomplishing what you're trying to do.

PR #604, opened 2024-02-07
Allow annotations_file to be None, for use in data preprocessing

### LINKED ISSUE #603: Allow split raster to be used without annotations.
Useful for data processing, there is no reason we need to insist that users have annotations when splitting rasters into pieces.

https://deepforest.readthedocs.io/en/latest/_modules/deepforest/preprocess.html#split_raster

https://github.com/weecology/DeepForest/blob/66fc43e73ce1b9a26a070c8a35da674bbfff2cf9/deepforest/preprocess.py#L141
1. This is the problem -> https://github.com/weecology/DeepForest/blob/66fc43e73ce1b9a26a070c8a35da674bbfff2cf9/deepforest/preprocess.py#L205

2. just set allow_empty = True if annotations_file = None.

3. Add a test here to assert case if there no annotation file.

https://github.com/weecology/DeepForest/blob/main/tests/test_preprocess.py

bw4sz:
Closes #603 and #576

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,10 @@
 DeepForest Change Log
 =====================
 
+**1.3.3**
+
+* Allow for annotations_file to be none in split_raster, for use in data preprocessing.
+
 **1.3.0**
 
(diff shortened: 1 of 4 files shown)


henrykironde (approved):
I made some minor refactor changes to clean up the code.

Issue #404, opened 2023-03-20
Profile the predict_tile method.


bw4sz:
The predict_tile function is the workhorse of the entire package. Most people will arrive with large geospatial tiles that cannot fit into memory. The predict_tile function needs to be more thoroughly profiled and understood. 

General instructions and psuedo-code
1. Take a large tile from https://zenodo.org/record/5912107#.ZBiLR-zMKDU

```
wget https://zenodo.org/record/5912107/files/2018_TEAK_3_315000_4094000_image_crop.tif?download=1 .
```

2. Use the predict tile function to generate predictions

3. Post the cProfile results here and identify which facets are slow under the following four conditions, cpu/gpu with config["workers"] > 0, 2) and cpu/gpu with config["workers"] == 0. What number of workers is fastest? Please post full details of operating system and GPU. 

The motivation was I was running predict_tile on a large number of RGB tiles on SLURM linux setup with GPU with 10 workers. I'm seeing alot of variation in the run time among tiles. This makes sense when there are no trees predicted, since there will be no non-max suppression to run (which suggests NMS is costly), but beyond that the iterations/per second seem to vary wildly. 

One tile 389 crops took 30 min

```

Predicting DataLoader 0: 100%|██████████| 389/389 [30:40<00:00,  4.73s/it]
```

Another tile 729 crops took 23 seconds just a minute later. Something is going on here with the workers locking?

```

100%|██████████| 729/729 [00:23<00:00, 30.65it/s]
```

martibosch:
Hello!

I am trying the `predict_tile` method on my computer's CPU (8 cores) and setting `workers` to 1, 4 or 8 in:

```
model = main.deepforest(config_args={"workers": 1})
model.use_release()
```

does not make any difference on the time taken by `predict_tile`. I wonder if this has anything to do with the warning at https://pytorch.org/docs/stable/data.html#multi-process-data-loading (which points to this thorough comment: https://github.com/pytorch/pytorch/issues/13246#issuecomment-905703662).

Thank you. Best,
Martí

henrykironde:
Have you tried to change the numbers `num_workers = 1  # could be 4 or 8`

henrykironde:
Ah never mind I think its something else, Let me check that out

bw4sz:
[...]

PR #609, opened 2024-02-17
issue #404, can we assert that pytorch workers are correctly connected?

bw4sz:
This is in response to #404 where user reports that the workers argument doesn't effect behavior. The original report is focused on the time elapsed. This doesn't necessarily mean something is wrong, there are many times when num_workers doesn't lead to a speedup. However, user reported that they can see the same number of workers being generated in task manager. I'm not yet sure how to assert this, since it happens fast and we don't want to resort to staring at task manager. 

I'm reading here

https://discuss.pytorch.org/t/communicating-with-dataloader-workers/11473/8
https://stackoverflow.com/questions/20353956/get-number-of-workers-from-process-pool-in-python-multiprocessing-module

I propose we start with a reasonably set of tests that assert config arguments and make sure it gets to the worker. Its going to hard to take it further, since at that point its up to torch and way upstream of DeepForest.

### DIFF
--- a/tests/test_multiprocessing.py
+++ b/tests/test_multiprocessing.py
@@ -1,10 +1,14 @@
-# Ensure that multiprocessing is behaving as expected. 
+# Ensure that multiprocessing is behaving as expected.
 from deepforest import main, get_data
 from deepforest import dataset
+from deepforest.utilities import read_config
+
 import pytest
 import os
+import shutil
+import yaml
 
-@pytest.mark.parametrize("num_workers",[0, 1 ,2])
+@pytest.mark.parametrize("num_workers",[0 ,2])
 def test_predict_tile_workers(m, num_workers):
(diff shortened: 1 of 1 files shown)


tests/test_multiprocessing.py
```diff
@@ -0,0 +1,25 @@
+# Ensure that multiprocessing is behaving as expected. 
+from deepforest import main, get_data
+from deepforest import dataset
+import pytest
+import os
+
+@pytest.mark.parametrize("num_workers",[0, 1 ,2])
+def test_predict_tile_workers(m, num_workers):
+    # Default workers is 0
```
henrykironde, line 9: Are we not using the value in the `deepforest_config_test.yml` file ?

bw4sz:
@henrykironde I added the test we discussed.

PR #617, opened 2024-03-05
Issue #569 fixed/Introduces 'path' as an alternative to 'raster_path'…

Abhishek-kumar0503:
Issue #569 fixed


bw4sz:
Thanks, running tests not.

deepforest/main.py
```diff
@@ -416,6 +416,7 @@ def predict_file(self, csv_file, root_dir, savedir=None, color=None, thickness=1
 
     def predict_tile(self,
                      raster_path=None,
+                     path= None,
```
ethanwhite, line 419: Remove the extra space after the `=`

deepforest/main.py
```diff
@@ -453,15 +454,22 @@ def predict_tile(self,
...
         self.model.nms_thresh = self.config["nms_thresh"]
 
+        if raster_path is not None:
+            warnings.warn(
+                "The 'raster_path' argument is deprecated and will be removed in future versions. Use 'path' instead.",
+            DeprecationWarning
+            )
+            path = raster_path
+
         if (raster_path is None) and (image is None):
```
ethanwhite, line 464: This line also need to be changed from `raster_path` to `path`

ethanwhite (changes requested):
Thanks for this @Abhishek-kumar0503. It looks like there are a number of additional uses of `raster_path` that need to be swapped out, and since this tests are currently passing this suggests that we need a new test of this function that uses `path` instead of `raster_path`. Can you add that test and then fix any remaining issues?

deepforest/main.py
henrykironde: We can also add the `path` as the last parameter to minimize the signature changes. But I am okey with leaving it in that position.

ethanwhite: Good to consider this @henrykironde, but I think that since  `raster_path` and `image` should never both be specified there was no way for folks to not use named arguments after `raster_path`. Therefore there shouldn't be any backwards breaking changes and I think leaving `path` first makes the most sense. Does that sound right?

henrykironde: Correct, that skipped my mind.

deepforest/main.py
Abhishek-kumar0503: okay i removed extra spaces after the  `=` and changed from `raster_path` to `path`.

Abhishek-kumar0503: @ethanwhite  please approve the testing workflow.

ethanwhite:
Thanks @Abhishek-kumar0503. As I noted in my overall review comment https://github.com/weecology/DeepForest/pull/617#pullrequestreview-1943695662 there are additional uses of `raster_path` that need to be updated and we need a test for this. Can you please add the test and then use it to check for other instances of `raster_path` that need to be changed.

Abhishek-kumar0503:
okay @ethanwhite I created the function name `def test_predict_tile_with_projection` and add it in test 
`def test_predict_tile_with_projection(m, raster_path):
    # test with projection enabled
    m.create_trainer()
    prediction, projection = m.predict_tile(raster_path=raster_path,
                                            patch_size=300,
                                            patch_overlap=0.1,
                                            return_plot=False,
                                            keep_projection=True)
    
    assert isinstance(prediction, pd.DataFrame)
    assert set(prediction.columns) == {"xmin","ymin","xmax","ymax","label","score","image_path"}
    assert not prediction.empty
    
    assert isinstance(projection, pd.DataFrame)  # Assert that projection dataframe is returned
    `
    and check it then it pass the test. Please Review it.

ethanwhite:
Thanks for adding the test. Have you tried debugging the resulting error? If so, where did you get stuck?

Abhishek-kumar0503:
ya it give error in unnecessary data type keep_projection. Solved!

ethanwhite:
Great! Can you rebase this to remove the merge conflicts and then I'll give it another look

Abhishek-kumar0503:
@ethanwhite it give error in check style. Can you help me to solve this issue.

ethanwhite:
@Abhishek-kumar0503 - try running this locally:

```sh
yapf -i --recursive deepforest/ --style=.style.yapf
```

That should cleanup the style issue.

Abhishek-kumar0503:
yeah,thanks it worked

ethanwhite (changes requested):
Thanks for the cleanup @Abhishek-kumar0503.

I see there are still uses of `raster_path` on lines 493, 497, and 498. It seems like those should all be `path` now.

I think the reason the test is passing is because you're test uses `raster_path`, but that is the deprecated behavior. If you update the test to use `path` instead of `raster_path` I suspect that it will fail, until you replace the uses of `raster_path` at the end of `predict_tile`. Does that make sense?

ethanwhite:
Closing due to inactivity. Feel free to reopen if you return to this.

Issue #625, opened 2024-03-20
Update package to address deprecation warnings


ethanwhite:
We have a growing number of deprecation warnings accumulating. Below are a number of these warnings that are produced by the current test suite. These aren't all necessarily from DeepForest directly (they may be upstream), but we should start finding the ones that are in DeepForest and update to get ahead of these changes. For folks looking for starter issues a great thing to do is to pick a single warning below, find one of our tests that produces the deprecation warning, and then see if the warning is triggered directly by the code run by that test or if it is upstream. If it is in the DeepForest code then work on a PR to fix just that warning and run the whole test suite to make sure that your fixes catches all instances of it in the code base.

- [x] The parameter 'pretrained' is deprecated since 0.13 and may be removed in the future, please use 'weights' instead.
- [x] Arguments other than a weight enum or `None` for 'weights' are deprecated since 0.13 and may be removed in the future. The current behavior is equivalent to passing `weights=FasterRCNN_ResNet50_FPN_Weights.COCO_V1`. You can also use `weights=FasterRCNN_ResNet50_FPN_Weights.DEFAULT` to get the most up-to-date weights.
- [x] datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
- [x] Expected None, but tests/test_FasterRCNN.py::test_retinanet returned <deepforest.models.FasterRCNN.Model object at 0x7f91cba5d940>, which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?
- [x] The verbose parameter is deprecated. Please use get_last_lr() to access the learning rate.
- [ ] The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
- [x] Type google._upb._message.MessageMapContainer uses PyType_Spec with a metaclass that has custom tp_new. This is deprecated and will no longer be allowed in Python 3.14.
- [x] Type google._upb._message.ScalarMapContainer uses PyType_Spec with a metaclass that has custom tp_new. This is deprecated and will no longer be allowed in Python 3.14.
- [x] The `op` parameter is deprecated and will be removed in a future release. Please use the `predicate` parameter instead.
[...]

PR #626, opened 2024-03-20
fixing the deprecation warning in the first two tasks of the issue #625

Mu-Magdy:
Fixing the first warning 

 `The parameter 'pretrained' is deprecated since 0.13 and may be removed in the future, please use 'weights' instead.`

Fixed the second warning

 `Arguments other than a weight enum or None for 'weights' are deprecated since 0.13 and may be removed in the future. The current behavior is equivalent to passing weights=FasterRCNN_ResNet50_FPN_Weights.COCO_V1. You can also use weights=FasterRCNN_ResNet50_FPN_Weights.DEFAULT to get the most up-to-date weights.`

### DIFF
--- a/deepforest/models/FasterRCNN.py
+++ b/deepforest/models/FasterRCNN.py
@@ -13,8 +13,7 @@ class Model(Model):
 
     def load_backbone(self):
         """A torch vision retinanet model"""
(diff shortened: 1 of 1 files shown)


ethanwhite (approved):
Thanks @Mu-Magdy! Looks great!

PR #627, opened 2024-03-20
solving the op deprecation warning 

Mu-Magdy:
solving the last task of #625 

 * _The op parameter is deprecated and will be removed in a future release. Please use the predicate parameter instead._


henrykironde (approved):
LGTM

PR #630, opened 2024-03-21
environment.yml error fix

Karwot:
Fixing the error of conda not being able to create the environment correctly:

CondaValueError: invalid package specification: setuptools <59.6.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -28,7 +28,7 @@ dependencies:
   - recommonmark
   - rtree
   - slidingwindow
-  - setuptools <59.6.
+  - setuptools <59.6
   - sphinx
   - sphinx_rtd_theme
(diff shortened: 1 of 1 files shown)


ethanwhite (changes requested):
Thanks @Karwot. Since we're here can you remove the extra space before the `<` as well. Should be `setuptools<59.6` right?

Karwot:
> Thanks @Karwot. Since we're here can you remove the extra space before the `<` as well. Should be `setuptools<59.6` right?

Good catch :) I just removed the dot since that was causing the error for me but it's definitely better to keep one convention inside the file.
Updated.

ethanwhite (approved):

ethanwhite:
Thanks @Karwot!

PR #632, opened 2024-03-24
solving the num_classes arg deprecation

Mu-Magdy:
solving the last deprecation error #625

> "Directly specifying the num_classes arg in deepforest.main will be deprecated in 2.0 in favor of config_args. Use main.deepforest(config_args={'num_classes':value})"


Mu-Magdy:
Hi @ethanwhite 

This is just a beginning to tackle the issue to know if I understand it correctly

PR #633, opened 2024-03-24
Issue #631: Remove the version dependency for setuptools 

### LINKED ISSUE #631: Check if version limit on setuptools can be removed from environment.yml
There is currently a limit to use older versions of setuptools in the conda environment. I believe this was implemented to work around an upstream bug at some point, that has hopefully been fixed. To test, remove the version dependency for setuptools from environment.yml, create a new conda environment using that file, and then run the tests in that environment.

Abhishek-kumar0503:
Fixed #631 Remove the version dependency for setuptools from environment.yml, Create a new conda environment using that file, and then check by run the tests in that environment.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -28,7 +28,7 @@ dependencies:
   - recommonmark
   - rtree
   - slidingwindow
-  - setuptools <59.6.
+  - setuptools
   - sphinx
   - sphinx_rtd_theme
   - torchvision>=0.13


ethanwhite:
Thanks @Abhishek-kumar0503! Just to confirm - have you tested this on your local system and it works for you?

Abhishek-kumar0503:
yes @ethanwhite

ethanwhite (approved):
Thanks @Abhishek-kumar0503! Looks good on my machine as well, so in it goes!

PR #636, opened 2024-03-26
opencv uses "setuptools==59.2.0",

henrykironde:
Keep setuptools to >59
Ref:
https://github.com/opencv/opencv-python/blob/8ad8ec1bae8aad196b3e9117a4b2b47d58547d47/pyproject.toml#L15

### DIFF
--- a/.github/workflows/Conda-app.yml
+++ b/.github/workflows/Conda-app.yml
@@ -44,6 +44,7 @@ jobs:
 
       - name: Run pytest, test style, and notebooks
         run: |
+          pip install -e .
           pytest -v
           yapf -d --recursive deepforest/ --style=.style.yapf
           pytest --nbmake docs/*.ipynb
--- a/environment.yml
+++ b/environment.yml
@@ -28,7 +28,7 @@ dependencies:
   - recommonmark
   - rtree
   - slidingwindow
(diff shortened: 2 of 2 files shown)


ethanwhite:
Thanks @henrykironde. I totally believe that this will fix things and that this is why we had that version pegged in the first place, but I'm confused as why. The build system should handle any needed upstream dependencies and all of this works fine locally:

```bash
$ conda create -n opencvtest python=3
$ conda activate opencvtest
$ conda install -c conda-forge opencv
$ conda list | grep setuptools
```

```bash
# Note that setuptools is the default 68.x, not the old version
setuptools                68.2.2          py312h06a4308_0 
```


```python
Python 3.12.2 | packaged by conda-forge | (main, Feb 16 2024, 20:50:58) [GCC 12.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import setuptools
>>> setuptools.__version__
'68.2.2'
>>> import cv2
>>> cv2.__version__
'4.9.0'
````

Are we building from source on GH Actions or something?

ethanwhite (approved):
I think it's worth getting this in to get the tests passing again even though we don't totally understand why it's an issue. Merging :crossed_fingers:

Issue #656, opened 2024-04-12
Compatibility with Apple M3 pro Chips


hesamossanloo:
**Please describe A feature you would like to be added.**
Support and compatibility of the PyTorch with Apple M3 pro Chips.

**Describe the solution you'd like**
Please update the training logic so that it is compatible when it is running on MPS devices (a.k.a Apple M chips). Atm I get the following error while training:
Code: 
m.trainer.fit(m)

Error:
"Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64. Please use float32 instead."

**Describe alternatives you've considered**
I couldn't find any solutions, so I had to train the model on Google Colab, which it worked.

```
**Additional context**
Traceback log:
TypeError                                 Traceback (most recent call last)
Cell In[252], [line 2](vscode-notebook-cell:?execution_count=252&line=2)
      [1](vscode-notebook-cell:?execution_count=252&line=1) start_time = time.time()
----> [2](vscode-notebook-cell:?execution_count=252&line=2) m.trainer.fit(m)
      [3](vscode-notebook-cell:?execution_count=252&line=3) print(f"--- Training on GPU: {(time.time() - start_time):.2f} seconds ---")

File [~/Projects/Startup/SkogApp/DeepForest/.venv/lib/python3.11/site-packages/pytorch_lightning/trainer/trainer.py:543](https://file+.vscode-resource.vscode-cdn.net/Users/hesam.ossanloo/Projects/Startup/SkogApp/DeepForest/~/Projects/Startup/SkogApp/DeepForest/.venv/lib/python3.11/site-packages/pytorch_lightning/trainer/trainer.py:543), in Trainer.fit(self, model, train_dataloaders, val_dataloaders, datamodule, ckpt_path)
    [541](https://file+.vscode-resource.vscode-cdn.net/Users/hesam.ossanloo/Projects/Startup/SkogApp/DeepForest/~/Projects/Startup/SkogApp/DeepForest/.venv/lib/python3.11/site-packages/pytorch_lightning/trainer/trainer.py:541) self.state.status = TrainerStatus.RUNNING
    [542](https://file+.vscode-resource.vscode-cdn.net/Users/hesam.ossanloo/Projects/Startup/SkogApp/DeepForest/~/Projects/Startup/SkogApp/DeepForest/.venv/lib/python3.11/site-packages/pytorch_lightning/trainer/trainer.py:542) self.training = True
[...]

PR #667, opened 2024-05-09
Update data type for bounding box coordinates to float32

hesamossanloo:
This pull request updates the data type for bounding box coordinates to float32. This change ensures compatibility with Apple M3 pro Chips #656

### DIFF
--- a/deepforest/dataset.py
+++ b/deepforest/dataset.py
@@ -100,7 +100,7 @@ class TreeDataset(Dataset):
                                                  self.image_names[idx]]
             targets = {}
             targets["boxes"] = image_annotations[["xmin", "ymin", "xmax",
-                                                  "ymax"]].values.astype(float)
+                                                  "ymax"]].values.astype("float32")
 
             # Labels need to be encoded
             targets["labels"] = image_annotations.label.apply(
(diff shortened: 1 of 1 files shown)


ethanwhite (changes requested):
Nice work figuring this out @hesamossanloo!

My one thought is that there are a few additional uses of `.astype(float)` in `utilities.py` and I wonder if they should also be changed as part of this PR in case folks using M3's want to use those helper functions.

Whenever we're explicit we are using `float32` everywhere else, so the current changes  look good to me and I think making the same change in `utilities.py` makes sense.

That said I'll want @bw4sz to review and be the one to merge this one in case there is something subtle.

henrykironde:
In the future, we should add test for MPS chips (Apple Silicon). Exploring current practices, it appears that many tests are conducted using purchased cloud or local instances.

bw4sz (approved):
This looks good to me, there is no downside to float versus float32 for just the annotations.

bw4sz:
@ethanwhite did you request changes? I can see your review is blocking.

ethanwhite:
> @ethanwhite did you request changes? I can see your review is blocking.

I was thinking about the additional uses of `.astype(float)` in `utilities.py` so we don't get caught by this again with the helper functions, but I'll approve this and open a new issue for those.

ethanwhite (approved):

ethanwhite:
Thanks for the contribution @hesamossanloo! Since @bw4sz has approved this as well I'm merging.

hesamossanloo:
Cheers @ethanwhite, @bw4sz :) I should thank you for keeping this great project running.

PR #673, opened 2024-05-22
log training loss in pytorch trainer

bw4sz:
With the updates to pytorch lightning 2.0, we appear to no longer to auto log the training loss! This is kinda a big deal, and took me by surprise. 

We previously got the training loss every epoch automatically.

Simple fix is to add self.log within training_step.


ethanwhite (approved):

ethanwhite:
Looks like the style tests are flagging an issue. Once that's fixed this is ready to go.

ethanwhite (approved):

Issue #317, opened 2022-03-28
Model library


bw4sz:
@ethanwhite, @henrykironde and I were discussing that as the number of trained models increase, we probably want some more refined way of versioning them and calling them. Right now they are saved alongside the release tags. Each new model would get a  "use_release" method, with redundant code, etc. 

An example release looks like:

https://github.com/weecology/DeepForest/releases/tag/1.0.0

Which is fine. When use_release is called, the model checks github for the latest release. A couple problems.

1. It doesn't ask the user if they want to use the new release, so any update to the model should be considered breaking. That seems like a big problem and we should atleast ask if we should update.
2. The metadata doesn't live alongside the model in python. There is a link, but not alot of info.

```
from deepforest import main
m = main.deepforest()
Reading config file: deepforest_config.yml
m.use_release()
Model from DeepForest release https://github.com/weecology/DeepForest/releases/tag/1.0.0 was already downloaded. Loading model from file.
Loading pre-built model: https://github.com/weecology/DeepForest/releases/tag/1.0.0
m.__release_version__
'https://github.com/weecology/DeepForest/releases/tag/1.0.0'
```

3. Torchvision recently released some API improvements. I think they are largely redundant with pytorch lightning. 
https://pytorch.org/blog/introducing-torchvision-new-multi-weight-support-api/
I have not read deeply enough to know if they can be used in parallel.

ayeankit:
Hey  @bw4sz , I would like to work on this issue, Can you please guide me?

henrykironde:
What have you done so far?

ethanwhite:
It seems like moving to Hugging Face for model distribution is the most common solution here. If we use the pytorch integration:

https://huggingface.co/docs/hub/models-uploading#upload-a-pytorch-model-using-huggingfacehub

then we can use the `from_pretrained()` method and pass in any of our models (and potentially others models as well; we can wrap `from_pretained` in `use_release` as well for backward compatibility):

```python
model.from_pretrained("weecology/deepforest-trees")
```

I setup an org account https://huggingface.co/weecology and can add @bw4sz & @henrykironde if this sounds like the general way to go.

bw4sz:
@henrykironde this will be an issue i'll connect to alive/dead model.

bw4sz:
# What we want

[...]

PR #675, opened 2024-06-04
Create a crop classifier workflow

bw4sz:


bw4sz:
We are long overdue to open-source the alive/dead model workflow in which a model at the _box_ level is used after tree segmentation. I propose we create a crop_classifier object that is passed to predict_tile as an argument.

```
from deepforest.model import Crop_Classifier
Alive_Dead = Crop_Classifier(<name>)
m.predict_tile(<existing_args>, crop_classifier=Alive_Dead)
```
same with predict image.

@henrykironde the path to the crop classifier will be the first test of the hugging face model repository #317.

## Roamap

- [ ] create class object in deepforest.model module and write tests
- [ ] Add existing alive/dead model architecture from DeepTreeAttention to deepforest/models folder https://github.com/weecology/DeepTreeAttention/blob/main/src/models/dead.py 
- [ ] upload and coordinates with @henrykironde to get existing model 
- [ ] create a crop_classifier dataloader
- [ ] Document workflow including how to finetune with new annotations at the crop level

For this particular alive/dead model, we should make it clear the model is overfit and the logistic scores are not that helpful.

bw4sz:
@henrykironde think I just put the alive/dead model weights into a github release as normal and when you work on #317 we can split them back out into hugging face.

henrykironde (approved):

PR #688, opened 2024-06-13
Update empty annotations handling

### LINKED ISSUE #681: Annotations Failed to load
image

``` python
crop_dir = os.path.join(os.getcwd(),'train_data_folder')
annotation_path=os.path.join(root_dir,'Colony43-March-2022.csv')
all_annotations=[]
for image in image_names:
  image_path=os.path.join(root_dir,image)
  annotations=preprocess.split_raster(
            path_to_raster=image_path,
            annotations_file=annotation_path,
            patch_size=400,
            patch_overlap=0.05,
            base_dir=crop_dir,
            # allow_empty=True
        )
  all_annotations.append(annotations)
train_annotations = pd.concat(all_annotations, ignore_index=True)
```
``` python
ValueError                                Traceback (most recent call last)
[<ipython-input-18-c02d1519539b>](https://localhost:8080/#) in <cell line: 4>()
      4 for image in image_names:
      5   image_path=os.path.join(root_dir,image)
----> 6   annotations=preprocess.split_raster(
      7             path_to_raster=image_path,
      8             annotations_file=annotation_path,

[/usr/local/lib/python3.10/dist-packages/deepforest/preprocess.py](https://localhost:8080/#) in split_raster(annotations_file, path_to_raster, numpy_image, base_dir, patch_size, patch_overlap, allow_empty, image_name, save_dir)
    300         # Only concat annotations if there were supplied
    301         if not annotations_files:
--> 302             raise ValueError(
    303                 "Input file has no overlapping annotations and allow_empty is {}".format(
    304                     allow_empty))

ValueError: Input file has no overlapping annotations and allow_empty is False
```

Mu-Magdy:
Delete empty bounding boxes that is all zeros.  closes #681

### DIFF
--- a/deepforest/preprocess.py
+++ b/deepforest/preprocess.py
@@ -263,6 +263,16 @@ def split_raster(annotations_file=None,
                 "path (e.g. 'image_name.tif'), not the full path "
                 "(e.g. path/to/dir/image_name.tif)".format(annotations_file, image_name))
 
+        # Find all 0 coordinates occurrences in an image
+        zero_annotations = (image_annotations[['xmin', 'ymin', 'xmax', 'ymax']] == 0).all(axis=1)
+        if zero_annotations.any():
+            
+            dropped_count = zero_annotations.sum()
+            image_annotations = image_annotations[~zero_annotations]
+            warnings.warn(f"Dropped {dropped_count} annotations with zero coordinates."
+                          "If you want to process empty frames, consider running a separate "
+                          "job with `allow_empty=True` to include images with no annotations.")
+
         required_columns = ["image_path", "xmin", "ymin", "xmax", "ymax", "label"]
         if not all(column in annotations.columns for column in required_columns):
             raise ValueError(f"Annotations file should have columns {required_columns}")


bw4sz:
@Mu-Magdy thanks for the PR, this one isn't quite right. 

```
    if not allow_empty:
        if image_annotations.empty:
            raise ValueError(
                "No image names match between the file:{} and the image_path: {}. "
                "Reminder that image paths should be the relative "
                "path (e.g. 'image_name.tif'), not the full path "
                "(e.g. path/to/dir/image_name.tif)".format(annotations_file, image_name))
```

is for dataframes without annotations. That is subtly different from when we allow users to provide negative samples by mocking a row. For example.

```
>>> a = pd.DataFrame({"xmin":[0],"xmax":[0],"ymin":[0],"ymax":[0],"label":"Bird"})
>>> a
   xmin  xmax  ymin  ymax label
0     0     0     0     0  Bird
>>> a.empty
False
```
pd.DataFrame.empty is pandas dataframes that have no rows.

```
>>> a
Empty DataFrame
Columns: []
Index: []
>>> a.empty
True
```

The point of #681 is just quietly drop the negative samples when using split raster. 

 ```
 non_empty_annotations = image_annotations[
            (image_annotations['xmin'] != 0.0) &
            (image_annotations['ymin'] != 0.0) &
            (image_annotations['xmax'] != 0.0) &
            (image_annotations['ymax'] != 0.0)
            ]
```
is fine, but let's use '0', and not 0.0 which implies float, and that makes me nervous. I need to keep thinking about this workflow but moving the logic outside of the .empty is the first step. It still feels like we haven't found the right idea here yet. What if someone wants to split empty frames, should they do that separately? We should provide advice to the user in the warning.

Mu-Magdy:
Hi @bw4sz,

I made a new commit to the pr 
can you please check it?
I don't really understand what advice can be given here.
i thought to make it an input if  the input is yes i drop all zero annotations else i keep them

PR #691, opened 2024-06-18
Added condition for different schedulers

Om-Doiphode:
I have updated the `configure_optimizers` in the `deepforest/main.py` file to account for the different learning rate schedulers and also made the necessary changes in the `deepforest_config.yml` file.

image

### DIFF
--- a/deepforest/main.py
+++ b/deepforest/main.py
@@ -635,15 +635,41 @@ class deepforest(pl.LightningModule):
         optimizer = optim.SGD(self.model.parameters(),
                               lr=self.config["train"]["lr"],
                               momentum=0.9)
-        scheduler = torch.optim.lr_scheduler.ReduceLROnPlateau(optimizer,
-                                                               mode='min',
-                                                               factor=0.1,
-                                                               patience=10,
-                                                               threshold=0.0001,
-                                                               threshold_mode='rel',
-                                                               cooldown=0,
-                                                               min_lr=0,
-                                                               eps=1e-08)
+        
+        scheduler_config = self.config["train"]["scheduler"]
+        scheduler_type = scheduler_config["type"]
+        params = scheduler_config["params"]
+        
+        if scheduler_type == "cosine":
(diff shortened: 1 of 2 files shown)


bw4sz:
Great thanks for this PR. 

couple pieces

1. With every pull request, you want to write a pytest that asserts the new functionality. Something like here
https://github.com/weecology/DeepForest/blob/199d3b71aaf1d4d6f3680d8f56691c1625d06eab/tests/test_main.py#L178
but where you change the config, using config_args when creating the deepforest object (https://deepforest.readthedocs.io/en/latest/ConfigurationFile.html#passing-config-arguments-at-runtime-using-a-dict) to show that the changes don't have unexpected consequences.

2. This change would be breaking for anyone with a previous version of a config file. Copy your new config to where it gets distributed with the rest of the package. @henrykironde this one of those long term maintenance things we want to automate on push, we should have two spots to update config files. But for now, its fine.

deepforest/data/deepforest_config.yml

3. Its good to have some arg defaults in case users have old custom configs. I think the existing code should stay in the code base above the new code as a default until a major release and mark it as to be deprecated. 

```
https://github.com/Om-Doiphode/DeepForest/blob/6e959757f9bfac578031234b72ca2f4c6097efd9/deepforest_config.yml#L37
```

Om-Doiphode:
HI @bw4sz, I have included 7 test functions in the `test_main.py` file to account for the different schedulers, updated the `deepforest_config.yml` files in other folders, and updated the documentation for the scheduler. I have run all the tests, and they have passed successfully. 

image

bw4sz:
Great, instead of writing 7 separate tests. Its one test with multiple parameters. Using pytest.mark.parameterize. 

For example:

https://github.com/weecology/DeepForest/blob/dba368b69b96f791beac596741de63e276900167/tests/test_main.py#L159

bw4sz (approved):
Looks good, thanks.

bw4sz:
Nice job @Om-Doiphode

Issue #697, opened 2024-06-25
Integrate visualizations from roboflows supervision annotator package


bw4sz:
One of the core beliefs of DeepForest has been to be *minimal*, and not reproduce code made by others. I am seeing alot of nice work on viz by Roboflow's supervision package

https://supervision.roboflow.com/latest/detection/annotators/#__tabbed_1_1

<img width="1158" alt="image" src="https://github.com/weecology/DeepForest/assets/1208492/cb5d9f53-fc4b-4bde-8557-f952f9c48614">

# Roadmap

- [ ] produce a working an example of how to take the results of main.deepforest.predict_tile or predict_image and get a supervision `bounding_box_annotator` class. I can see examples for YOLO and transformers packages, but not an immediate example of how to get torchvision results. We may need to either submit a PR, or just write a simple converter.

```
import supervision as sv
from deepforest import main
from deepforest import get_data
import cv2

m = main.deepforest()
m.use_release()
img_path = get_data("OSBS_029.tif")



image = cv2.imread(img_path)
result = m.predict_image(img_path)

# Some kind of converter to a format that supervision knows about, this does not yet exist.
sv_format = convert_to_sv_format(result)

detections = sv.Detections(sv_format)

bounding_box_annotator = sv.BoundingBoxAnnotator()
annotated_frame = bounding_box_annotator.annotate(
    scene=image.copy(),
    detections=detections
)
```

@Mu-Magdy do you want to have a look at this portion and see if you can play with this idea. The issue itself is large, but the connector function is a good first step.

Mu-Magdy:
Hi @bw4sz 

I made this function to do so 

```python
def convert_to_sv_format(df):

    """
    Convert DeepForest prediction results to a supervision Detections object.

    Args:
        df (pd.DataFrame): The results from `predict_image` or `predict_tile`.
                           Expected columns: ['xmin', 'ymin', 'xmax', 'ymax', 'label', 'score', 'image_path'].

    Returns:
        sv.Detections: A supervision Detections object containing bounding boxes, class IDs,
                       confidence scores, and class names object mapping classes ids to corresponding
                       class names inside of data dictionary.


    Example:
        detections = convert_to_sv_format(result)
    """
    # Extract bounding boxes as a 2D numpy array with shape (_, 4)
    boxes = df[['xmin', 'ymin', 'xmax', 'ymax']].values.astype(np.float32)

    label_mapping = {label: idx for idx, label in enumerate(df['label'].unique())}

[...]

PR #709, opened 2024-07-08
Add missing aiohttp dependency

ethanwhite:
aiohttp was added in e5dc2e70fc4446eb86fc16dc882658276c97235f but was not included as a dependency


bw4sz:
Redundant with #697 that happened inside of it, it would have been more organized to do it separately, but done now.

PR #726, opened 2024-07-19
Adding pre-commit install to setup.py

Mu-Magdy:


ethanwhite:
My guess is you all have talked about this already, but I'm not sure how I feel about putting developer specific things in `setup.py`. This feels like a case where we might be introducing some difficult to anticipate cross-OS edge cases into our install process (our experience with maintaining `subprocess.call` across user operating systems is not great) to save developers following a couple of extra instructions. Given that I think I lean towards just sticking with the details that have now been added to the developer docs in bb707d2a0b5368b76fa72f3603f913179f1d1524.

Mu-Magdy:
HI @ethanwhite 

I can't say we all have talked about this, it just was a suggestion from me to keep the install process simple as most of the library users are not developers so I wanted to find a way to keep the process one-liner but I didn't find a good way so this is not final if you have other idea I can search about

ethanwhite:
Thanks @Mu-Magdy. I like the thought process about keeping things simple for non-developers, but I'm confused about why non-developers would need the pre-commit hooks. It seems like if they aren't going to be submitting pull requests then this isn't something that folks need. Am I misunderstanding the use case?

Mu-Magdy:
I didn't think about this, you are right, I will close this

ethanwhite:
Thanks!

PR #729, opened 2024-07-29
doc cleaning

bw4sz:
While on the plane, I went through docs and improved wording.


docs/annotation.md
```diff
@@ -36,7 +54,7 @@ for crop in output_crops:
     assert os.path.exists(crop)
```
ethanwhite, line 54: I suggest dropping the assert statements here. Typically the user just wants to know the commands, not have proof that they work and a lot of beginners won't really understand what's going on with assert. If we want something that shows the files were created then I think this would be clearer for most novices:

```python
import os

os.listdir()
```

ethanwhite (changes requested):
One typo and a suggestion to cut a little more material. It's fine with me if you prefer to just fix the typo.

ethanwhite:
Looks like you've done a bunch of new work here in the revision. Can you clean up the history so the additions vs. cleanup changes are well described in the commit log and then I'll take another look.

bw4sz:
Squashed and gave new commit message.

docs/training.md
```diff
@@ -215,3 +211,80 @@ m = main.deepforest(transforms=get_transform)
...
+https://github.com/weecology/DeepForest/issues/338
+
+If you run into this error, users (e.g https://github.com/weecology/DeepForest/issues/443), have found that creating the trainer object within the loop can resolve this issue. 
+
+```
+for tile in tiles_to_predict:
+    m.create_trainer()
+    m.predict_tile(tile)
+```
+Usually creating this object does not cost too much computational time.
```
ethanwhite, line 256: Add a newline after the code block

docs/training.md
```diff
@@ -215,3 +211,80 @@ m = main.deepforest(transforms=get_transform)
...
+
+**Training across multiple nodes on a HPC system**
```
ethanwhite, line 258: This should be a level 4 header

docs/training.md
```diff
@@ -215,3 +211,80 @@ m = main.deepforest(transforms=get_transform)
...
+    trainer = Trainer(
+        accelerator="gpu",
+        strategy="ddp",
+        devices=model.config["devices"],
+        enable_checkpointing=False,
+        max_epochs=model.config["train"]["epochs"],
+        logger=comet_logger
+    )
+trainer.fit(m)
+```
```
ethanwhite, line 287: Newline

ethanwhite (changes requested):
Looks great! Couple of minor things to cleanup.

Also, one more thing on the commit message. We want the first line to be a summary no longer than (about) 50 characters and then the rest goes in "the body" after a line break. So in this case something like:

```
Condense docs and add info on reading/writing data

Also:
* Condenses & moves the FAQ doc into better location
* Improves the table of contents ordering to be more logical
```

Nice explanation of the how and why of good commit messages here: https://cbea.ms/git-commit/

ethanwhite (approved):
LGTM!

ethanwhite:
@henrykironde - I'm going to merge this. You'll want to rebase any of the docs work you're doing on this since it does change the Table of Contents

PR #733, opened 2024-08-05
Remove dependency duplications

ethanwhite:
In e0dbabd4722a240a8d9c8978e813625caae56bc7 two lines duplicating
dependencies snuck into setup.py. This removes them.


henrykironde (approved):
Thank you @ethanwhite. I was sure we cleaned that part well at the time but I was wrong.

PR #749, opened 2024-08-14
Added augmentations

### LINKED ISSUE #735: Improved zoom augmentations through albumentations.
One of the core challenges of machine learning for airborne biodiversity observation is trying to generalize across sensors and acquisition conditions. Given low data sizes, data augmentations are crucial for good generalization across resolutions, focal views and object size.

A quick search of albumentations suggests hits a few existing classes that should be useful:

https://huggingface.co/spaces/qubvel-hf/albumentations-demo?transform=Downscale

https://albumentations.ai/docs/api_reference/augmentations/crops/transforms/#albumentations.augmentations.crops.transforms.RandomSizedBBoxSafeCrop

https://huggingface.co/spaces/qubvel-hf/albumentations-demo?transform=RandomSizedBBoxSafeCrop

https://albumentations.ai/docs/api_reference/augmentations/geometric/transforms/#albumentations.augmentations.geometric.transforms.PadIfNeeded

# Checklist

- [ ] Implement augmentations in their own module, not within preprocessing. Currently lives inline
https://github.com/weecology/DeepForest/blob/80cb7d80f5c51d19930205cf6b8878c7640dfa1e/deepforest/dataset.py#L35

- [ ] Allow the user to choose the augmentations either through the config file. Careful to allow defaults to remain unchanged and sets reasonable defaults if not specified in existing config files.

- [ ] Make a doc page showing example augmentations

## Optional

- [ ] Compare training with augmentations and without when predicting _across_ resolutions.

Om-Doiphode:
Fixes #735 

- Added support for `RandomSizedBBoxSafeCrop` and `PadIfNeeded` in the `deepforest/augmentations.py` file 
- Added a doc page for augmentations
- Modified the `deepforest_config.yml` file to include default and user defined augmentations

### DIFF
--- a/deepforest/augmentations.py
+++ b/deepforest/augmentations.py
@@ -1,0 +1,145 @@
+import albumentations as A
+from deepforest import utilities, get_data
+import cv2
+from albumentations.pytorch import ToTensorV2
+
+def get_augmentations(augment=True):
+    try:
+        config_path = get_data("deepforest_config.yml")
+    except Exception as e:
+        raise ValueError("No config file provided and deepforest_config.yml "
+                         "not found either in local directory or in installed "
+                         "package location. {}".format(e))
+    
+    augmentations = []
+    format_bbox='pascal_voc'
+    label_fields=['category_ids']
+    try:
+        config = utilities.read_config(config_path)
+        augment_config = config["train" if augment else "validation"]["augmentations"]
+    except KeyError as e:
+        raise ValueError("Missing key in the configuration file: {}".format(e))
+    
+    if not augment_config:
+        default_augmentations=config["train" if augment else "validation"]["default_augmentations"]
+        for augment in default_augmentations:
+            augment_config_type=augment.get('type')
+            params = augment.get('params', {})
+            
+            if augment_config_type.lower()=='horizontalflip':
+                augmentations.append(A.HorizontalFlip(**params))
+                
+            elif augment_config_type.lower()=='verticalflip':
+                augmentations.append(A.VerticalFlip(**params))
+                
+            elif augment_config_type.lower()=='randomrotate90':
+                augmentations.append(A.RandomRotate90(**params))
+                
+            elif augment_config_type.lower()=='randomcrop':
+                augmentations.append(A.RandomCrop(**params))
+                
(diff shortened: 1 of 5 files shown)


ethanwhite (changes requested):
Thanks for working on this @Om-Doiphode! It's definitely going to be super useful having easier flexible augmentation for training.

I have a couple of interrelated questions about design that I'd like us to think through on this one:

1. Should the defaults be included in the config file or in the code?
2. If they are stored in the config should it be a separate `default_augmentation` section or just be what is provided under the `augmentations` section in the default config file?
3. Should the augmentation config be stored in our config file, or should it be stored in a separate config file following the albumentations JSON format?

I'll preface my thoughts on these by acknowledging that I'm not familiar with how this is normally handled so I'm happy to defer to others or look at examples from other packages.

1\. If we have defaults that are basically always applied in our current code and we think we are almost always going to want to apply them I lean towards keeping those in the code. However if we envision users wanted to regularly change those defaults then I think it's reasonable to include them in a config file.

2\. Typically in this sort of case the default values would just be what is in the relevant section of the default config file. So instead of having a `default_augmentations` section, those augmentations would be what is in `augmentations` section. We would then provide instructions for how to modify this section. One way that this is done (in addition to docs) is by having chunks of the config file commented out with a header line that says something like "Uncomment the code below if you want to add the PadIfNeed augmentation".  

3\. `albumentations` already has a [loadable config file format](https://albumentations.ai/docs/autoalbument/how_to_use/#step-3-use-albumentations-to-load-augmentation-policies-and-utilize-them-in-your-training-pipeline) that can be stored as either YAML or JSON. The YAML version looks quite similar to the YAML design in this PR. The advantage of using this format and moving to a separate augmentation config file is that it eliminates a lot of new custom code for reading and setting up augmentations.

So, having written all of this out my initial thought is that we should use the existing `albumentations` config and associate loading code and either have our defaults in the default albumentations config file provided or have them created in the code and then automatically combined with the additional augmentations provided in the `albumentations` config file. What do you think @bw4sz & @henrykironde?

bw4sz:
I'm back monday, this will have to wait till then.

PR #750, opened 2024-08-15
Switch to opencv-python-headless

### LINKED ISSUE #464: Dependencies in `setup.py` have module conflicts.
**Describe the bug**
There are two dependencies mentioned in the `setup.py` file: `opencv-python` and `albumentations` and the 'albumentations' depends on `opencv-python-headless`. The official [spec](https://pypi.org/project/opencv-contrib-python/)  mentioned that the `opencv-python` package is for the desktop environment, while `opencv-python-headless` is for the server environment. The documentation also states that these two packages cannot be installed simultaneously (the exact wording is: “There are four different packages (see options 1, 2, 3, and 4 below) and you should SELECT ONLY ONE OF THEM.”). This is because they both use the same module name cv2.

During the installation process using pip, the package installed later will override the cv2 module from the previously installed package (specifically, the modules within the cv2 folders that exist in both packages). Furthermore, the dependency graph even includes different versions of these two packages. It is certain that the common files with the same path in these two packages contain different contents. Therefore, there may be functional implications when using them. However, without analyzing the specific code and function call hierarchy of this project, it can be stated that issues related to overwriting and module conflicts do exist. 

**To Reproduce**
`pip install deepforest`

**Screenshots**
image

**Additional context**
Indeed, it is not an ideal behavior for modules to be overwritten, even if they are not actively used or if the overwritten module is the one being called. It introduces uncertainty and can cause issues in the long run, especially if there are changes or updates to the overwritten modules in future development. It is generally recommended to avoid such conflicts and ensure that only the necessary and compatible dependencies are declared in the requirements to maintain a stable and predictable environment for the project. 

We believe that although this project can only modify direct dependencies and indirect dependencies are a black box, it is possible to add additional explanations rather than directly declaring both conflicting packages in the requirements.txt file.

Adding extra explanations or documentation about the potential conflicts and the need to choose only one of the conflicting packages can help developers understand the issue and make informed decisions. Including clear instructions or warnings in the project’s documentation can guide users to choose the appropriate package based on their specific requirements.

ethanwhite:
Using opencv-python was causing installation conflicts with
albumentations which relies on the headless version. We were only using
the GUI aspects of opencv in a single, debugging only, function. So
this switches that function to matplotlib and updates the dependencies.

This does drop the `cv2.waitKey` functionality. It can be replicated in
matplotlib, but it's definitely more complicated and it didn't seem crucial
given the debugging use case.

Fixes #464.


henrykironde (approved):

henrykironde:
Thanks @ethanwhite for the PR

Issue #754, opened 2024-08-18
More cookbooks for quick start


bw4sz:
User asks

```
we were wondering if you have any sample scripts that you share that e.g., load images, load annotations, train canopy delineation/extent model on image subset, run model on the entire image, assess accuracy of canopy delineation, attribute tree species based on field data collection, and assess accuracy of tree species. We can figure this out ourselves (possibly with some additional help from you), but I just thought I’d ask if you have a sample script for these tasks.
 ```

My sense is that all of this already exists within the docs, but the user started to look for it, couldn't find what they wanted and email me. @henrykironde this is part of the broader effort to improve navigation within the docs.
For each, let's point to location we would expect the user to find them in docs and figure out how to name and direct them there more cleanly.
- [ ] load images
- [ ] load annotations
- [ ] train model on image subset (i'm not sure exactly what subset means here, crop?)
- [ ] run model on entire image
- [ ] assess accuracy of model
- [ ] Multi-class point data to boxes
- [ ] Train Multi-class Model and Validate.

We could approach this as one long use case for a particular project in the example/ dir with jupyter notebooks, or we can make them tiny gists in a cookbook section. @ethanwhite thoughts?
Let's use this issue to keep track of the different places

colbyrand:
Perhaps this is already somewhere in the docs, and I can't seem to find it, but is is there an example script that you could share that touches on the following points:

- loads images and annotations for both the training and validation areas
- runs the default model on an image and explains what model parameters you can change and where to change those in context of forest conditions (e.g., open v closed canopy)
- trains the canopy delineation model to a subset of the image based on the annotations you supply and validates model with a second set of annotations
- Runs the trained model on the entire image and exports the predictions and model accuracy assessment
- Use field validated training and test data to predict tree species and assess accuracy of tree species prediction.

Thanks!

bw4sz:
Good example from supervision

https://roboflow.github.io/cheatsheet-supervision/

robAndrus34:
@bw4sz, thanks for the example in supervision. I'll take a look at that. 

Additionally, I was curious if there was any movement on creating a tutorial that addresses the points in @colbyrand's post above (Aug. 21st, 2024)?

Thank you!

bw4sz:
Thanks for reminding me, let's try to get that done in connection with upcoming workshops. Putting on the roadmap.

Issue #761, opened 2024-08-28
Consider decoupling plotting from predict_* functions


ethanwhite:
Currently our predict functions return two possible outputs:
- A numpy array that represents an image of the RGB data with the predictions drawn on top
- A Pandas DataFrame with bounding box coordinates

The difference is the value of the optional `return_plot` argument, with `True` returning the numpy array.

This can be confusing (see #760). I'm wondering if the more intuitive design in the long run is to have `predict_*` only return data frames (including maybe spatial data frames see #608) and then have a function (or suite of functions) called something like `visualize_predictions()` to make graphs. This seems like a relatively straightforward split with the new function(s) taking either a raster object or path and a data frame returned by `predictions_*`.

One added complexity is that `predict_tile()` can also return a third data type if the optional argument `mosaic` is set to `False`:
- "a tuple of image crops and predictions"

This one doesn't come up at the moment because it isn't described in the docs. I'm not super clear on the use case for this last one, so clarifying that might make the design clearer.

bw4sz:
I agree. As part of this, I want to hide the function [convert_to_sv](https://github.com/weecology/DeepForest/blob/d834b17b949e25df0f75905108955ec4928f025e/deepforest/visualize.py#L260) from users 

- [ ] Better doc and example of visualizing the output of predict_* functions.
- [ ] Streamline and reduce the visualizing module, too much duplications and not clear the user workflow.

PR #764, opened 2024-08-30
related to #754, add more simple examples to help quickstart

bw4sz:


docs/examples/simple_load_and_predict.py
```diff
@@ -0,0 +1,39 @@
+from deepforest import main
+from deepforest import visualize
+import supervision as sv
+import cv2
+
+# This script loads an image, performs tree detection using the DeepForest library, and visualizes the detected trees with bounding boxes.
+
+# Import necessary libraries
+import matplotlib.pyplot as plt
```
ethanwhite, line 9: Seems like this should just be up with the rest of the imports

ethanwhite (changes requested):
This looks good except for a small fix related to imports.

It does highlight the need for #761 because I think for our users Lines 24-34 (and maybe even through 39) should basically be:

```python
visualize.visualize_predictions(trees)
```

Put all the imports together and then I think this is ready to merge.

ethanwhite (approved):

PR #770, opened 2024-09-06
add thoughts on similar tools

bw4sz:
@ethanwhite i've been thinking about adding text to the landing to orient users towards the developing ecosystem. Thoughts?


ethanwhite:
No objections. I think it would be fine towards the bottom of what is now `docs/introduction/index.rst`

ethanwhite (changes requested):
Looks like you added a copy or near copy of the "How can I contribute?" section that is higher up in the current document. Presumably we only want one copy of this. I don't have strong feelings about where in the doc it should be, so take your pick.

ethanwhite (approved):
Looks good!

PR #777, opened 2024-09-12
Added a sidebar

Om-Doiphode:
A sidebar on the documentation page was added for easy user access.

image


henrykironde:
Added in #793. Thank you @Om-Doiphode for these changes

PR #783, opened 2024-09-16
Cleanup docs on reading shape files to avoid user confusion

ethanwhite:
* Clearly show reading files from a path without complexity of temp dirs and format strings from the tests
* Remove unnecessary local shell command


henrykironde (approved):
Thanks @ethanwhite

Issue #790, opened 2024-09-25
Multi-color plotting length error


bw4sz:
I'll find a reproducible example, but looks like the multi-class color isn't quite right, introduced in #789 

```
Traceback (most recent call last):
  File "<string>", line 6, in <module>
  File "/blue/ewhite/b.weinstein/miniconda3/envs/AirborneFieldGuide/lib/python3.10/site-packages/deepforest/visualize.py", line 405, in plot_results
    if num_labels > 1 and len(results_color_sv) != num_labels:
TypeError: object of type 'Color' has no len()
```

```
image_annotations = read_file(image_annotations, root_dir=train_image_dir)
visualize.plot_results(image_annotations, res)
```

```
image_annotations.iloc[0]
xmin                                                      804.0
ymin                                                       23.0
xmax                                                      854.0
ymax                                                       73.0
label                                                White Ibis
image_path               C4_L24_F873_T20240417_075715_656_7.png
annotator                 Brier Klossing b.klossing@ufl.edu, 20
geometry      POLYGON ((854 23, 854 73, 804 73, 804 23, 854 ...
Name: 256, dtype: object
```

We'll need another look at this.

ethanwhite:
Closed by #791

PR #791, opened 2024-09-25
Allow multi-class plot in plot_results

bw4sz:
There was an error in the multi-class logic. 

https://github.com/weecology/DeepForest/issues/790

This introduces a simple fix to check the object type and a test to assert this function.

### DIFF
--- a/deepforest/visualize.py
+++ b/deepforest/visualize.py
@@ -402,7 +402,7 @@ def plot_results(results,
                                      ground_truth_color[2])
 
     num_labels = len(results.label.unique())
-    if num_labels > 1 and len(results_color_sv) != num_labels:
+    if num_labels > 1 and type(results_color_sv) == sv.ColorPalette:
         warnings.warn(
             """Multiple labels detected in the results and results_color argument provides a single color.
             Each label will be plotted with a different color using a built-in color ramp.
--- a/deepforest/visualize.py
+++ b/deepforest/visualize.py
@@ -402,7 +402,7 @@ def plot_results(results,
                                      ground_truth_color[2])
 
     num_labels = len(results.label.unique())
-    if num_labels > 1 and type(results_color_sv) == sv.ColorPalette:
+    if num_labels > 1 and not type(results_color_sv) == sv.ColorPalette:
         warnings.warn(
             """Multiple labels detected in the results and results_color argument provides a single color.
             Each label will be plotted with a different color using a built-in color ramp.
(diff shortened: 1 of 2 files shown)


ethanwhite (commented):
I'm confused by the change here. It checks for a multi-class context and then only (and always) sets the color palette if the user has passed in a color palette. It seems like that means that: 1) the multi-color palette will only be set if the user passes a color palette (i.e., not if they leave the default); and 2) it will always override a user specified set of multi-class colors. Am I misunderstanding something?

ethanwhite (commented):
Looks like it might just need to be:

```python
if num_labels > 1 and len(results_color_sv.colors) != num_labels:
```

ethanwhite:
I also just got a update merged into `supervision` so this will work as originally written with the next release: https://github.com/roboflow/supervision/pull/1550 (but we should go ahead and make the change I mentioned above and then we can change it back later)

bw4sz:
The suggested change raises
```
import os
import pandas as pd
from deepforest import get_data
from deepforest import visualize
from deepforest.utilities import read_file

results = pd.read_csv(get_data("testfile_multi.csv"))
results = read_file(results, root_dir=os.path.dirname(get_data("SOAP_061.png")))
visualize.plot_results(results)
```

```
>>> visualize.plot_results(results)
  File "/Users/benweinstein/Documents/DeepForest/deepforest/visualize.py", line 405, in plot_results
    if num_labels > 1 and len(results_color_sv.colors) != num_labels:
                              ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Color' object has no attribute 'colors'
```

bw4sz:
I thin

> I'm confused by the change here. It checks for a multi-class context and then only (and always) sets the color palette if the user has passed in a color palette. It seems like that means that: 1) the multi-color palette will only be set if the user passes a color palette (i.e., not if they leave the default); and 2) it will always override a user specified set of multi-class colors. Am I misunderstanding something?

I think you are right here, it should be 

```
    if num_labels > 1 and not type(results_color_sv) == sv.ColorPalette:
```

if the user has multi-class labels, and has not already passed a colorPalette, make a color palette.

ethanwhite:
Oops, sorry. The multiple types thing is definitely a little confusing to keep track of. I do think we still want to consider the length of the provided color palette if provided. I just pushed an (actually tested) edit that refactors a little to put all of the type specific code together at the beginning and hopefully get exactly the behavior that we want. See what you think.

bw4sz:
looks good to me, merge when ready.

ethanwhite (approved):

PR #795, opened 2024-09-30
Set evaluate iou_threshold recall

bw4sz:
There should never be a situation in which the user wants to pass None to iou_threshold. It creates a strange effect and isn't reasonable. I am not going to spend the time to investigate what mechanism prevented this from being obvious sooner, or how it interacted with the config arguments.


ethanwhite (approved):
LGTM

PR #801, opened 2024-10-02
delete use_release and use_bird_release from tests and docs

Mu-Magdy:
delete use_release and use_bird_release from tests and docs


henrykironde:
Thank you @Mu-Magdy for these changes.  LGTM

henrykironde (approved):

PR #802, opened 2024-10-02
Predict docs should state returns a Pandas DataFram

henrykironde:


ethanwhite (approved):
Thanks @henrykironde!

PR #804, opened 2024-10-09
Release version update

henrykironde:


ethanwhite (approved):
LGTM

henrykironde:
@ethanwhite do you think we should amend/squash the second commit or leave it in the history?

ethanwhite:
> @ethanwhite do you think we should amend/squash the second commit or leave it in the history?

No strong feelings. It's a weird situation. Whatever you think is best is good by me

henrykironde:
@ethanwhite with the current structure we have 
```Python
>>> from deepforest._version import __version__
>>> __version__
'1.4.1'

```

ethanwhite:
Is that a new Python standard?

ethanwhite:
I'm booked most of the afternoon so @henrykironde just go with what you think is best. My thinking was based on `deepforest.__version__` being a de facto standard:

https://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package

But if that has changed I'm fine with doing it differently. I'm going to change to Approve so that I'm not blocking you and @bw4sz finishing off the release

ethanwhite (approved):

PR #805, opened 2024-10-10
BuildAndRelease job runs if push to main or a tag is created

henrykironde:
Skip pre-release tags (like 1.4.0-dev0, 1.4.0-rc1)


ethanwhite (approved):
LGTM

PR #808, opened 2024-10-14
load_model should not return anything

bw4sz:
Keeping the behavior with load_model and use_release, there isn't a use case where a user would capture the output of load_model

```
from deepforest import main
m = main.deepforest()
```
it should not return self.


ethanwhite (approved):
This is consistent with prior use, the use in the docs, and the naming. :+1:

PR #816, opened 2024-10-21
Python Required version

henrykironde:


ethanwhite:
3.12 is fine right? It's 3.13 that isn't support by torch and therefore us

henrykironde:
Thanks @ethanwhite. The errors from 3.12 were repaired with ubuntu latest patches.

ethanwhite (approved):

PR #822, opened 2024-10-24
Add Authors, update to pyproject.toml and publish with GitHub Actions

henrykironde:


henrykironde:
@bw4sz or @ethanwhite can any one please comment on this. I tried to change `pyproject.toml` but I think we should wait on that.

bw4sz (approved):

PR #824, opened 2024-10-27
Do not package tests and docs

henrykironde:


henrykironde:
We failed to release because the files were too large. Removing the docs and tests reduces the built artifacts by almost 80%

ethanwhite (approved):

Issue #809, opened 2024-10-14
Point detection workflow


bw4sz:
As part of DeepForest 2.0 we need a model for point prediction.

A couple examples we can work from

https://github.com/xingyizhou/CenterNet.
https://github.com/FateScript/CenterNet-better
https://arxiv.org/abs/1904.07850

# Roadmap

- [ ] Add small point dataset example to get_data. We should use the existing box dataset (OSBS_029) and just get the centroids
- [ ] Add a point dataloader to datasets.py similar to https://github.com/weecology/DeepForest/blob/7e0aa4394b50beecc0584fcc62511a8aa7d75e21/deepforest/dataset.py#L204
- [ ] Add a point model to models. 
- [ ] Example training script in examples/
- [ ] Confirm visualization
- [ ] Confirm can be extended to multi-class 

@Mu-Magdy might be interested, since its isolated from the code base, doesn't have a time frame and fits objectives.

bw4sz:
Also check https://github.com/airctic/icevision/tree/master/icevision/models/torchvision/keypoint_rcnn

bw4sz:
The question is whether we use segment anything to just convert these points into box structures to train.

musaqlain:
Hi @bw4sz @jveitchmichaelis,
I would like to tackle this issue. since the current focus is on expanding prediction tasks (as per this comment [[here](https://github.com/weecology/DeepForest/issues/1260#issuecomment-3740012205)]), I think this issue should be worked on next. (also this issue is part of `DeepForest 2.2` milestone)
Plus I have also explored the relevant merged PR **#1078**. I see that it laid the groundwork (metrics and parsing), but the actual model and training workflow for points are still missing. I plan to build directly on top of that work.
**My Proposed plan:**
* **Dataset (`PointDataset`):** I will implement a new dataset class. I propose adopting the **TorchGeo-style dictionary format** (e.g., returning `{'image': tensor, 'points': tensor}`) instead of the legacy tuple format. This aligns with the API modernization discussed previously.
* **Model Wrapper:** I will implement a point-specific model class in `models/`. I plan to use the wrapper structure introduced in #1078 as a template but adapted for point regression (likely using a heatmap approach).
* **Training Loop:** I will extend the `training_step` to handle point-specific loss functions, ensuring it integrates smoothly without breaking the existing box detection logic.


I am ready to start with the dataset implementation. Let me know if this approach looks good to you!

PR #833, opened 2024-11-09
[WIP] Point detection model

Mu-Magdy:
Added an example data for the OSBS_29 as csv
Added a center point dataloader 
A converter to convert bounding boxes to centroids for future usage


Mu-Magdy:
solves #809

PR #846, opened 2024-12-06
Add an image argument to plot_results for tranformed data.

bw4sz:
I ran into an annoying issue with plot_results. It assumes that the image exists on file, but often we augment the images (resize, zoom, etc) and there isn't a way to pass that augmented image during prediction. Added a simple image argument to bypass image loading from disk when needed.


bw4sz:
Do not merge yet, I need to write a test, which passed locally, but failed when applying to MillionTrees.

bw4sz:
Ready to merge.

ethanwhite (approved):

PR #850, opened 2024-12-11
Update the pyproject author list

henrykironde:


ethanwhite (approved):
CI error is unrelated to this PR so approving and merging, but looks like we may have another build change on GH Actions that we need to cleanup

PR #852, opened 2024-12-12
updated main.deepforest.predict_file for DF usage

### LINKED ISSUE #797: main.deepforest.predict_file should be able to take in a dataframe, not just a csv file.
Looking at the code base.

https://deepforest.readthedocs.io/en/latest/_modules/deepforest/main.html#deepforest.predict_file

it actually looks like dataset.TreeDataset() is the problem here, because utilities.read_file is definitely flexible enough. The docstring of predict_file should be updated and the argument name should be changed (but a deprecation warning added until 2.0), but its the dataset class, which the user really never sees that often that can be updated. I think this would still be considered a patch.

naxatra2:
###  Fixes #797 
In the `main.deepforest.predict_file` should now also be able to take in a dataframe, not just a csv file.
There was another PR open (PR #838) for similar issue but it wasn't working so I improved it.

### DIFF
--- a/docs/user_guide/07_scaling.md
+++ b/docs/user_guide/07_scaling.md
@@ -1,8 +1,6 @@
 # Scaling DeepForest using PyTorch Lightning
 
-Often we have a large number of tiles we want to predict. DeepForest uses [PyTorch Lightning](https://lightning.ai/docs/pytorch/stable/) to scale inference. This gives us access to powerful tools for scaling without any changes to user code. DeepForest automatically detects whether you are running on GPU or CPU. Within a single GPU node, you can scale training without needing to specify any additional arguments, since we use the ['auto' devices](https://lightning.ai/docs/pytorch/stable/common/trainer.html#devices) detection within PyTorch Lightning. For advanced users, DeepForest can [run across multiple SLURM nodes](https://lightning.ai/docs/pytorch/stable/clouds/cluster_advanced.html), each with multiple GPUs. 
-
-## Increase batch size
+### Increase batch size
 
 It is more efficient to run a larger batch size on a single GPU. This is because the overhead of loading data and moving data between the CPU and GPU is relatively large. By running a larger batch size, we can reduce the overhead of these operations. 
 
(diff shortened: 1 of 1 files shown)


naxatra2:
Its clearing all the tests in local environment. I don't know why its not showing in github

ethanwhite:
> Its clearing all the tests in local environment. I don't know why its not showing in github

That is weird. Can you try either opening a new PR from the same branch or pushing a tiny change to a comment to see if it will trigger the build (or ask us to trigger it).

ethanwhite:
That did it. I think there was just hiccup with the prompt asking us to allow the tests to run (since you're a new contributor). The build failure is something that just cropped up with an image change. Hopefully #855 will fix that and then we can rerun

naxatra2:
Okay. I was very confused as to what went wrong. This certainly helps.

ethanwhite:
@naxatra2 - If you rebase this on `main` the tests should run now

naxatra2:
Is this good ?

ethanwhite:
Thanks @naxatra2 - tests are passing now. We'll try to review the PR next week.

ethanwhite:
@naxatra2 - looks like a style issue crept in with your updated commits. See https://deepforest.readthedocs.io/en/v1.4.1/development/contributing.html#using-yapf to fix it and then we'll get this reviewed for you now that the holidays are over

tests/test_main.py
```diff
@@ -649,8 +649,7 @@ def test_predict_tile_with_crop_model(m, config):
         "xmin", "ymin", "xmax", "ymax", "label", "score", "cropmodel_label", "geometry",
         "cropmodel_score", "image_path"
     }
-
-
+
```
henrykironde, line 652: Is it possible to come up with some test for the added feature functionality?

henrykironde (commented):
The function looks good. I have left a comment on the test section

tests/test_main.py
naxatra2: okay. I will write a new test case and commit it

PR #854, opened 2024-12-13
Optional silencing of predict_tile, verbose=False

bw4sz:
The logs for predict tile when running at large scales are cluttered with hundreds of stdout for the predict dataloader. Added a verbose (defaults to true) and added a test to assert it silences verbose=False.

### DIFF
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -477,7 +477,8 @@ class deepforest(pl.LightningModule, PyTorchModelHubMixin):
                      thickness=1,
                      crop_model=None,
                      crop_transform=None,
-                     crop_augment=False):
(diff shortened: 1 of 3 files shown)


ethanwhite:
If you rebase this on `main` the tests should run now

henrykironde:
@bw4sz can you check the test

bw4sz:
Runs locally.

image


I'll need to check pytorch-lightning version.

bw4sz:
I have updated pytorch-lightning version to match between github actions and local, and I still cannot produce this error. 

image

bw4sz:
I simply cannot recreate this error locally. Might need others to try.

PR #858, opened 2024-12-17
Refactor evaluation loop for empty frames

bw4sz:
This PR aims to standardize, document and better test those situations in which 1) the model doesn't make predictions, but there are ground truth to evaluate and 2) where there are no ground truth in validation, but the model makes predictions. I introduce a new evaluation metric 'empty frame accuracy' that will be useful for many users, I added tests and docs. During this, I found that the test that was called validation_step wasn't fully testing validation_step, but rather trainer.validate, which is related, but not the same. I added a proper validation step test, which required silencing the loggers, due to pytorch lightning being overeager in throwing errors if self.validation_step is run outside of trainer.validate, which is only needed for testing situations.

### DIFF
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -36,7 +36,6 @@ class deepforest(pl.LightningModule, PyTorchModelHubMixin):
 
     Returns:
         self: a deepforest pytorch lightning module
-
     """
 
     def __init__(self,
@@ -143,7 +142,6 @@ class deepforest(pl.LightningModule, PyTorchModelHubMixin):
 
         Returns:
             None
-
         """
         # Load the model using from_pretrained
         self.create_model()
@@ -196,7 +194,6 @@ class deepforest(pl.LightningModule, PyTorchModelHubMixin):
 
         Returns:
             None
-
         """
         if self.model is None:
             model_name = importlib.import_module("deepforest.models.{}".format(
@@ -213,7 +210,6 @@ class deepforest(pl.LightningModule, PyTorchModelHubMixin):
 
         Returns:
             None
-
         """
         # If val data is passed, monitor learning rate and setup classification metrics
         if not self.config["validation"]["csv_file"] is None:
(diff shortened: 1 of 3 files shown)


bw4sz:
Tensorboard is no longer a required module in pytorch lightning, so need to be explictly installed. I am updating the tests since this got caught more as a side-effect and not anything to do with tensorboard.

bw4sz:
@henrykironde this is ready, sorry for the delay with all the docstring commits, my sphinx version was outdated.

src/deepforest/main.py
```diff
@@ -682,31 +672,97 @@ def validation_step(self, batch, batch_idx):
...
         # Log loss
         for key, value in loss_dict.items():
-            self.log("val_{}".format(key), value, on_epoch=True)
+            try:
+                self.log("val_{}".format(key), value, on_epoch=True)
+            except:
+                pass
 
         for index, result in enumerate(preds):
             # Skip empty predictions
```
henrykironde, line 681: We can remove the comment

src/deepforest/main.py
```diff
@@ -720,8 +776,21 @@ def on_validation_epoch_end(self):
             ground_df = utilities.read_file(self.config["validation"]["csv_file"])
             ground_df["label"] = ground_df.label.apply(lambda x: self.label_dict[x])
 
+            # If there are empty frames, evaluate empty frame accuracy separately
+            empty_accuracy = self.calculate_empty_frame_accuracy(
+                ground_df, self.predictions_df)
+
+            if empty_accuracy is not None:
+                try:
```
henrykironde, line 784: The try feels so broad.

henrykironde (commented):
@bw4sz, this looks good! One thought: could we handle the try blocks more specifically? They feel a bit broad as they are.

src/deepforest/main.py
bw4sz: I added a specific Exception.

src/deepforest/main.py
bw4sz: Added specific exception.

bw4sz:
I rebased, I noticed that somehow the batch tests were commented out, I'll make a separate PR, that is unrelated.

PR #860, opened 2024-12-24
fixes the main.predict_file for numeric labels 

### LINKED ISSUE #705: CropModel should return labels not numeric values.
CropModel should return labels not numeric values.

```
    raster_path = get_data("SOAP_061.png")
    patch_size = 400
    patch_overlap = 0.05
    iou_threshold = 0.15
    return_plot = False
    mosaic = True


    # Set up the crop model
    crop_model = model.CropModel()

    # Call the predict_tile method with the crop_model
    m.create_trainer()
    result = m.predict_tile(raster_path=raster_path,
                                           patch_size=patch_size,
                                           patch_overlap=patch_overlap,
                                           iou_threshold=iou_threshold,
                                           return_plot=return_plot,
                                           mosaic=mosaic,
                                           crop_model=crop_model)

    # Assert the result
    assert isinstance(result, pd.DataFrame)
    assert set(result.columns) == {"xmin", "ymin", "xmax", "ymax", "label", "score", "cropmodel_label","cropmodel_score","image_path"}
```

cropmodel_label here is 0 or 1. I think that it should better match result["label"] from the detection as being a string. Users will be happier. 

* Implement a label dict just like the detection model
https://github.com/weecology/DeepForest/blob/cce441c1a126920e9f5ab88346dc3c16e604cb22/deepforest/main.py#L27

naxatra2:
Fixes #705
---
I have updated the cropmodel_label in `main.predict_file`. Earlier it was showing 0 or 1, now it better matches the  result["label"] from the detection as being a string. 
The testcases are also running.

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -661,9 +661,6 @@ def test_predict_tile_with_crop_model_empty():
     mosaic = True
     crop_model = model.CropModel()
 
-    # Configure the label dictionary
-    m.label_dict = {"Tree": 0, "Bush": 1}
-
     # Call the predict_tile method with the crop_model
     m.config["train"]["fast_dev_run"] = False
     m.create_trainer()
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -659,8 +659,10 @@ def test_predict_tile_with_crop_model_empty():
     patch_overlap = 0.05
     iou_threshold = 0.15
     mosaic = True
-    # Set up the crop model
     crop_model = model.CropModel()
+
+    # Configure the label dictionary
+    m.label_dict = {"Tree": 0, "Bush": 1}
 
     # Call the predict_tile method with the crop_model
     m.config["train"]["fast_dev_run"] = False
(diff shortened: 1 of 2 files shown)


tests/test_main.py
```diff
@@ -659,9 +659,11 @@ def test_predict_tile_with_crop_model_empty():
     patch_overlap = 0.05
     iou_threshold = 0.15
     mosaic = True
-    # Set up the crop model
     crop_model = model.CropModel()
 
+    # Configure the label dictionary
+    m.label_dict = {"Tree": 0, "Bush": 1}
```
henrykironde, line 665: @naxatra2 Why do we have this here?

henrykironde (commented):
@naxatra2 Thank you for your contribution! I'm not sure if this part is necessary in the tests (`test_predict_tile_with_crop_model_empty`). Since the changes are focused on `predict_file`, I was expecting the test updates to primarily be in `test_[predict]*[file]`.

tests/test_main.py
naxatra2: I accidentally committed that part, earlier I was using a different logic where I used this. I should remove this

naxatra2:
I though that I did changes on `main.py` and not on` predict.py` that's why I added that test in the `test_main.py` file. What changes should I do to improve the new testcase for this. I didn't get it ?

ethanwhite:
@henrykironde - looks like this PR is waiting on additional feedback from you

henrykironde:
@naxatra2 can you update this PR with main and we see if it passes

ethanwhite:
A fix for this was merged in #936. Thanks again for your help on this @naxatra2 and for your further improvements in #966!

PR #856, opened 2024-12-14
Batch Predicitions

RohitP2005:
I added a batch function mirroring the functionalities of predict_image,predict_tile and predict_file. I documented this behaviour inside a file called batch_predicitions.md since i was unsure where to document those. I also added test cases for the batch function, but commented them in the last of the file. Waiting for any corrections needed

PR #876, opened 2025-01-08
Improve predict batch

bw4sz:
In #856, I didn't realize that the tests were there, but they were commented out. I shortened them and made them a bit more explicit. I also made the function run slightly faster by running each batch, not each image in each batch.


bw4sz:
I think this merits 1.5.0 if you can trigger it.

henrykironde (approved):

henrykironde:
About the 1.5.0 release, I think I will need some more time to go through the code and do some tests before.

PR #878, opened 2025-01-09
Allow silencing of predict_tile stdout based on verbose flag

henrykironde:


bw4sz:
@henrykironde what was the secret here, before I merge.

henrykironde:
@bw4sz @ethanwhite, while this solution works, it's not ideal. The script relies on keeping the console output clean to force the test to pass, which feels like a workaround. It's challenging to make `verbose=False` tests pass reliably since other output (beyond the expected print statements) might be present.


```Python
        if mosaic:
             # Suppress output if not verbose
             if not verbose:
                 f = io.StringIO()
                 with contextlib.redirect_stdout(f):
```

bw4sz:
We are sitting on this issue until we find a better solution, possibly that pytorch-lightning gets updated.

PR #879, opened 2025-01-09
Improve point plotting behavior for ground truth with no scores.

bw4sz:
The current behavior fails if you have ground truth that don't have a 'scores' column. Which makes no sense, since ground truth don't have confidence scores, from the model atleast. This behavior was already fixed for boxes and polygons. I added a try except, added a dummy 1 for scores in the dataframe just within the plotting function, and added a test.


ethanwhite (approved):
LGTM

ethanwhite:
Looks like there's a style error that needs fixing before merge

ethanwhite:
Thanks. Just waiting on the tests

ethanwhite:
Docs error is independent

Issue #862, opened 2024-12-26
The use-bird-release method is outdated compared to the hugging face method, change it in the docs.


bw4sz:
The prebuilt model page has the old method.

PR #890, opened 2025-01-16
made sure that docs never use deprecated use_release

bw4sz:
I asked cursor to have a look at the docs and make sure we never use_release anywhere. Found a few.


bw4sz:
closes #862

henrykironde (approved):

PR #894, opened 2025-01-18
num_classes shouldn't be an optional arg for CropModel

bw4sz:
This is more of a style change than a bug. CropModel had num_classes = 2 as a default. But there is no reason besides the example we gave that this should be the case. Having it be an optional argument in reloading from checkpoint, since I forgot to specify num_classes=5 and it silently failed. I added a test for better coverage.


ethanwhite (changes requested):
Style tests are failing. Otherwise looks good

ethanwhite (approved):
Thanks @bw4sz

PR #895, opened 2025-01-22
Update the docs version switcher to include the newest release

ethanwhite:


ethanwhite:
@henrykironde Do you understand why 1.5.0 was showing up in the switcher even though it wasn't in this file previously?

PR #900, opened 2025-01-27
Update version switcher

henrykironde:


ethanwhite:
1.5.1 was just a test tag and it isn't currently included in the built docs or releases. I'd recommend closing this and deleting the tag, but if you thing we should merge this let me know and I'm happy to follow your lead.

PR #916, opened 2025-02-10
profiled test_main.py to improve speed

bw4sz:
Screenshot 2025-02-09 at 7 31 44 PM
I was able to bring tests in test_main.py in half by making sure that we never run trainer.fit unless we fully need to. Still more work to be done, but a good start. 

Before, 34 minutes.

Screenshot 2025-02-09 at 7 31 44 PM

After 18, note this is just test_main.py

```
--------------- generated report log file: reportlog_test_main2.jsonl ----------------
=================== 54 passed, 128 warnings in 1117.63s (0:18:37) ====
```

### DIFF
--- a/tests/test_callbacks.py
+++ b/tests/test_callbacks.py
@@ -8,9 +8,8 @@ from pytorch_lightning.callbacks import ModelCheckpoint
 from deepforest import get_data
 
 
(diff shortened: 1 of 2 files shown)


ethanwhite (approved):
Definitely a big improvement. Nice work!

Issue #924, opened 2025-02-16
Fix path separator inconsistency in crop_raster tests on Windows


Abhishek-Dimri:
**Describe the bug**
The test_crop_raster_valid_crop and test_crop_raster_png_unprojected tests are failing on Windows due to path separator inconsistencies. The crop_raster function returns paths with forward slashes (/), while the tests expect backslashes (\). This causes the following AssertionError:

![Image](https://github.com/user-attachments/assets/f7749a9b-f0f1-4cb3-9beb-c0bb84ae26e3)

**To Reproduce**
Step1-> Run the test suite on Windows:
         ```pytest -v tests/test_utilities.py```

Step2-> Observe the failures in test_crop_raster_valid_crop and test_crop_raster_png_unprojected.

**Environment (please complete the following information):**
 - OS: Microsoft Windows 11 Home
 - Python version and environment : Python 3.10.11 and Global Environment

**Screenshots and Context**

![Image](https://github.com/user-attachments/assets/85ecf00f-a108-4214-8ca4-2196111d41a5)

![Image](https://github.com/user-attachments/assets/ab0b46a6-eb18-4fb1-9eca-f3aa528d0bfa)

![Image](https://github.com/user-attachments/assets/dfe88c38-d2f2-4bd9-942b-d03802157148)

**Proposed Solution**
Normalize both the result and expected_filename paths in the tests using os.path.normpath(). This ensures that the comparison is done correctly, regardless of the path separators used by the function or the operating system.
Here’s the updated test code:
```python
def test_crop_raster_valid_crop(tmpdir):
    rgb_path = get_data("2018_SJER_3_252000_4107000_image_477.tif")
    raster_bounds = rio.open(rgb_path).bounds

    # Define the bounds for cropping
    bounds = (raster_bounds[0] + 10, raster_bounds[1] + 10, raster_bounds[0] + 30, raster_bounds[1] + 30)

    # Call the function under test
    result = utilities.crop_raster(bounds, rgb_path=rgb_path, savedir=tmpdir, filename="crop")

    # Assert the output filename (normalize both paths)
    expected_filename = str(tmpdir.join("crop.tif"))
    assert os.path.normpath(result) == os.path.normpath(expected_filename)

    # Assert the saved crop
    with rio.open(result) as src:
        # Round to nearest integer to avoid floating point errors
        assert np.round(src.bounds[2] - src.bounds[0]) == 20
        assert np.round(src.bounds[3] - src.bounds[1]) == 20
[...]

PR #926, opened 2025-02-17
Fix path separator inconsistency in crop_raster tests on Windows

Abhishek-Dimri:
This PR fixes the path separator inconsistency in the `test_crop_raster_valid_crop` and `test_crop_raster_png_unprojected` tests on Windows (Issue #924). The issue occurs because the `crop_raster` function returns paths with forward slashes (`/`), while the tests expect backslashes (`\`).

Changes:
- Normalized both the `result` and `expected_filename` paths in the tests using `os.path.normpath()`.
- Ensured that the tests work correctly on all operating systems.

Testing:
- Ran the test suite on Windows and confirmed that the tests pass:
  pytest -v
  
  
image


ethanwhite (approved):
Thanks for this @Abhishek-Dimri!

PR #930, opened 2025-02-22
Fixed broken prebuilt link in overview.rst

Abhishek-Dimri:
Before
image

After
image

Take 3!

### DIFF
--- a/docs/getting_started/overview.rst
+++ b/docs/getting_started/overview.rst
@@ -81,7 +81,7 @@ The first is the DeepForest paper, which describes the Python package:
 
    Weinstein, B.G., Marconi, S., Aubry‐Kientz, M., Vincent, G., Senyondo, H. and White, E.P., 2020. DeepForest: A Python package for RGB deep learning tree crown delineation. Methods in Ecology and Evolution, 11(12), pp.1743-1751. `https://doi.org/10.1111/2041-210X.13472 <https://doi.org/10.1111/2041-210X.13472>`_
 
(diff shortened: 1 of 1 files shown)


ethanwhite:
Is there a reason you're closing your own PRs? This seems like a good fix, though one that I'll suggest an additional improvement to if you want to submit it.

Abhishek-Dimri:
Figuring out why these checks have failed. 😢
image
image

ethanwhite:
Thanks for explaining @Abhishek-Dimri. Typically I recommend leaving the original PR open and pushing additional changes to try to fix the issue. I typically do something like:

```
git add myfile.py
git commit --amend
git push -f origin mybranch
```

This will update the PR without adding any new commits and will cause the build system to rerun all the tests.

Now, in this case the failing tests don't have anything to do with the PR. `nbsphinx` (a docs extension we use) was [broken a few days ago by an upstream change](https://github.com/spatialaudio/nbsphinx/issues/825).

So, let's go ahead and get this merged and then one `nbsphinx` gets fixed it will show up on the docs site.

ethanwhite (approved):
Docs build error is upstream https://github.com/spatialaudio/nbsphinx/issues/825 and the tests have already passed on a previous run, so looks good!

Thanks @Abhishek-Dimri!

Abhishek-Dimri:
@ethanwhite Thanks for your guidance and for merging the PR! Please share the additional improvement you mentioned. I'd be happy to incorporate it into the PR.

ethanwhite:
> Please share the additional improvement you mentioned. I'd be happy to incorporate it into the PR.

Actually I had misremembered something. Your existing solution is the right one. Thanks!

PR #936, opened 2025-02-25
add a label dict to the crop model

### LINKED ISSUE #705: CropModel should return labels not numeric values.

bw4sz:
While working on the BOEM pipeline, its quite annoying to not to be 100% sure how the numeric labels relate to the annotation labels. Get worried about ordering, better to just add an argument and be clearer. 

Closes #705.

### DIFF
--- a/src/deepforest/model.py
+++ b/src/deepforest/model.py
@@ -79,10 +79,35 @@ def simple_resnet_50(num_classes=2):
 
 
 class CropModel(LightningModule):
-    """An architecture agnostic class for classification based on crops coming
-    from the core detection models."""
+    """A PyTorch Lightning module for classifying image crops from object detection models.
 
-    def __init__(self, num_classes, batch_size=4, num_workers=0, lr=0.0001, model=None):
+    This class provides a flexible architecture for training classification models on cropped
+    regions identified by object detection models. It supports using either a default ResNet-50
+    model or a custom provided model.
+
+    Args:
+        num_classes (int): Number of classes for classification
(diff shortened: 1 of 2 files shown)


ethanwhite (changes requested):
Looks good to me. Style tests are failing. Once the style cleanup is in I'll get this merged.

PR #941, opened 2025-02-27
replaced visualise.plot_prediction with plot_results

### LINKED ISSUE #934: Image callback still using old visualization code
https://github.com/weecology/DeepForest/blob/00bcbf59ade3efa8a6b0f680042630d688b98d36/src/deepforest/callbacks.py#L61

We have moved to visualize.plot_results().

naxatra2:
Fixes #934

I have  replaced `visualise.plot_prediction` with `plot_results` in `DeepForest/src/deepforest/callbacks.py`. I don't think it requires any additional test cases so I didn't add it. But, If they are relevant then I may add them.


naxatra2:
I noticed that `DeepForest/src/deepforest/Predict.py`. also has some instances of `visualise.plot_prediction`. I didnn't change them but should we change them ?

PR #942, opened 2025-02-28
 Added batch size configuration for training, validation, and prediction

### LINKED ISSUE #889: Add seperate config for predict batch size and train batch size.
Updating model weights takes alot more GPU memory than just a forward model pass.

predict.tile is slower than it needs to be because its using trainer.predict, which inherits a dataloader with batch size set by the global config https://github.com/weecology/DeepForest/blob/3dbc8342de766f1f504d3c2da69c1fbc2443da42/src/deepforest/main.py#L348

and in train gets from load_dataset.

https://github.com/weecology/DeepForest/blob/3dbc8342de766f1f504d3c2da69c1fbc2443da42/src/deepforest/main.py#L335

the default is 1 because training is unknown size GPU (probably should be 2)

1. Make a predict_batch_size and a train_batch_size config arg
2. Update defaults to 2 for train and 8 for predict.
3. update the config doc
4. Write tests showing the dataloaders of each are yielding correct sizes.

I'm unsure about the val dataloader batch size, maybe should be higher, not clear to me the GPU memory. I think val batch size should be the predict size, since no weights are updated.

Bhavya1604:
Closes #889 

This PR introduces a configurable batch size for training, validation, and prediction in the DeepForest model. The changes allow users to modify batch sizes via the configuration file, improving flexibility for different computational requirements.

Changes made:
Added train_batch_size and predict_batch_size parameters to the configuration file.
Updated train_dataloader, val_dataloader, and predict_dataloader to use these parameters.

Testing:
Ran unit tests (pytest), and all tests passed except test_read_coco_json, which is unrelated to these changes.
Verified that batch size settings correctly update in different training modes.

Looking forward to feedback from the maintainers!

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,12 @@
 DeepForest Changelog
 ====================
 
+Version x.x.x (Date: )
+---------------------------------
+
+- **Deprecation:** ``predict_tile`` in ``deepforest/main.py``. The ``raster_path`` argument is deprecated and will be removed in a future version.
+
+- **Enhancement:** Updated the ``predict_tile()`` function to use the ``path`` argument (replacing ``raster_path``) across all code, tests, and documentation for consistency.
 
 Version 1.5.2 (Date: Feb 7, 2025)
 ---------------------------------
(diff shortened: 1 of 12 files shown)


bw4sz (changes requested):
The core idea is here, but I'm not convinced yet that this is the performance bottleneck, it may interact with workers argument.

PR #952, opened 2025-03-03
Newer versions of sphinx work with nbsphinx

ethanwhite:
We had temporarily limited the sphinx version due to compatibility issues
with nbsphinx. These issues have been addressed upstream, so we can return
to using the most recent version.

See: https://github.com/spatialaudio/nbsphinx/issues/825#issuecomment-2692944098


henrykironde (approved):
Thanks @ethanwhite for the update

PR #963, opened 2025-03-09
Fix broken image link in README.md

Samia35-2973:
This fixes an incorrect image path in the README.md file that was preventing the image from rendering correctly. The previous path used "../www/image.png", which likely caused issues with Markdown rendering on GitHub. 
The path has been updated to "www/image.png" to ensure proper display.


ethanwhite (approved):
Thanks for the contribution @Samia35-2973!

henrykironde:
Ref: Temporarily merging this code despite a possible relocation. We can move it later if necessary.

PR #966, opened 2025-03-10
Refactor CropModel to infer num_classes from checkpoint during loading

### LINKED ISSUE #960: num_classes should not be a required argument for CropModel
Title: Refactor CropModel to infer num_classes from checkpoint during loading

https://github.com/weecology/DeepForest/blob/171f55f3117ec8068f8185e667263c4b9231087f/src/deepforest/model.py#L87

## Current Behavior
Currently, `CropModel` requires `num_classes` as a mandatory argument during initialization:

```python
model = CropModel(num_classes=2)  # num_classes must be known beforehand
```

This creates problems when loading from a checkpoint because users need to know the exact number of classes the model was trained with, even though this information is stored in the checkpoint.

## Desired Behavior
`num_classes` should be optional during initialization and automatically loaded from the checkpoint when available:

```python
# Loading from checkpoint without needing to specify num_classes
model = CropModel.load_from_checkpoint("path/to/checkpoint.ckpt")
```

## General idea toward Solution

Working with copilot, I came up with:

Modify `CropModel` to use PyTorch Lightning's `on_load_checkpoint` hook to handle the `num_classes` parameter. Here's the proposed implementation:

```python
class CropModel(LightningModule):
    def __init__(self, num_classes=None, batch_size=4, num_workers=0, lr=0.0001, model=None, label_dict=None):
        super().__init__()
        self.num_classes = num_classes
        
        # Only initialize model if num_classes is provided
        if num_classes is not None:
            self._init_model(num_classes)
            
    def _init_model(self, num_classes):
        """Initialize model and metrics with given num_classes"""
        if getattr(self, 'model', None) is None:
            self.model = simple_resnet_50(num_classes=num_classes)
            
        self.accuracy = torchmetrics.Accuracy(average='none',
                                            num_classes=num_classes,
                                            task="multiclass")
        self.total_accuracy = torchmetrics.Accuracy(num_classes=num_classes,
                                                  task="multiclass")
        self.precision_metric = torchmetrics.Precision(num_classes=num_classes,
                                                     task="multiclass")
        self.metrics = torchmetrics.MetricCollection({
            "Class Accuracy": self.accuracy,
            "Accuracy": self.total_accuracy,
            "Precision": self.precision_metric
        })
            
    def on_load_checkpoint(self, checkpoint):
        if self.num_classes is None:
            # Extract num_classes from checkpoint
            self.num_classes = checkpoint['hyper_parameters']['num_classes']
            self._init_model(self.num_classes)
```

## Benefits
1. More user-friendly API when loading from checkpoints
2. Follows PyTorch Lightning's best practices for checkpoint handling
3. Maintains backward compatibility (can still specify num_classes if needed)
4. Reduces potential errors from mismatched class numbers
5. Self-documenting code that makes the model's configuration clear from the checkpoint

This needs to be tested and tests added.

naxatra2:
Fixes #960

This PR makes `num_classes` optional in the `CropModel` class and infers it from the checkpoint when loading via `load_from_checkpoint`. This improves usability, especially for users who do not know or remember the original number of classes at the time of inference.

**Changes Made**
1.  Made `num_classes` an optional argument in the CropModel constructor.
2.  In the constructor, if `num_classes` is not provided (None), we defer model/metrics initialization until on_load_checkpoint. Otherwise, we initialize them immediately.
3.  Added an `on_load_checkpoint` method that reads `num_classes` from the checkpoint if it is still `None`, then sets up the model and metrics.
4. Added tests to verify that:
   - We can load from a checkpoint without passing `num_classes`.
   - Users can still pass `num_classes` explicitly if they want.
   - Calling `forward` without having a valid model raises a clear error.


bw4sz:
This is important and should be merged, two additions. We need to use

```
self.save_hyperparameters()
```

at the end of init documented here https://lightning.ai/docs/pytorch/1.6.2/common/hyperparameters.html#lightningmodule-hyperparameters

The second is the label dict, the crop model uses torchvision dataloader that loads folders from file, that could easily not be the same as the label_dict that as user could optionally pass. We should remove label_dict as an argument. Then we should add a label_dict argument here 

https://github.com/weecology/DeepForest/blob/f1c724dcd48dbfda1fdb5a052c97ee7f59160b5c/src/deepforest/model.py#L157

I believe its under self.train_ds. class_to_idx.

https://pytorch.org/vision/main/generated/torchvision.datasets.DatasetFolder.html#torchvision.datasets.DatasetFolder

Now, the question I have is if a use creates a model, saves the hyperparameters, loads a dataset, saves a checkpoint, and then reloads the model, is the label_dict still there?

naxatra2:
If it is not urgent then can I try doing this @bw4sz, it will be a great learning opportunity. I will update my progress here.

naxatra2:
I have tried to do some changes. mainly did these things:
- [x] changed `self.save_hyperparameters()`
- [x] removed the if statement from here 
> There should not be a if statement here. CropModels should have label dicts and they should return labels, not numeric.
- [x] Test for a checkpoint with explicit `num_classes` to test backward compatibility

----

> Now, the question I have is if a use creates a model, saves the hyperparameters, loads a dataset, saves a checkpoint, and then reloads the model, is the label_dict still there?

Currently, if a user creates a model, loads a dataset (which assigns the label_dict), saves a checkpoint, and then reloads the model, the label_dict is not automatically restored, because it isn’t passed as an init argument

To make the label mapping consistent across checkpoints, we can override the checkpoint hooks to include the label_dict in the saved checkpoint and then restore them when the model is reloaded.

bw4sz:
Yes, we can use on_load_checkpoint.

Two other things I saw that generally improve the CropModel

Softmax, not sigmoid, they are equivalent in two class
https://github.com/weecology/DeepForest/blob/e3fd2d087fb8cc69a920e2489f0a5f762436cd96/src/deepforest/model.py#L211

Default value should not be true for validation shuffle.
https://github.com/weecology/DeepForest/blob/e3fd2d087fb8cc69a920e2489f0a5f762436cd96/src/deepforest/model.py#L237

bw4sz:
Also, can we make the naming of the crops more informative, 

```
                image_basename = os.path.splitext(os.path.basename(image))[0]
                img_path = os.path.join(savedir, label, f"{image_basename}_{index}.png")
```

https://github.com/weecology/DeepForest/blob/e3fd2d087fb8cc69a920e2489f0a5f762436cd96/src/deepforest/model.py#L200

naxatra2:
@bw4sz are these changes good ? I have made the necessary changes in the last commit

bw4sz:
Yes, this is excellent PR, thank you.

bw4sz (approved):

PR #975, opened 2025-03-14
Fix #545: Unified predict_tile() arguments using 'path' and deprecating 'raster_path'

### LINKED ISSUE #545: Unify args in predict_tile and predict_image
In predict_tile, the path argument is called 'raster_path' and in predict_image its 'path'. There is no reason we have more than one name for the same thing. We need to allow the current naming, add a deprecation warning and add a 'path' argument to predict_tile.

https://github.com/weecology/DeepForest/blob/3c5cbd62de8b1d6f29b980e9eb40d9932d9a57a3/deepforest/main.py#L438

Samia35-2973:
This PR updates the `predict_tile` function to use `path` as an alternative to `raster_path` while issuing a deprecation warning when `raster_path` is used. I have also tried to maintain the test cases.

Here's what I modified-
**In `main.py`:**  
- Introduced `path` as a replacement for `raster_path`.  
- Issued a `DeprecationWarning` when `raster_path` is provided.  
- Updated all references to `raster_path` to use `path`.  

**In `test_main.py`:**  Updated the test cases to use `path` instead of `raster_path`.

**In `test_visualize.py`:**  Modified `test_predict_tile_and_plot` to use `path` instead of `raster_path`.  

I did not modify the following function in `test_dataset.py`

```python
def raster_path():
    return get_data(path='OSBS_029.tif')
```
Please review this and let me know if this function should also be updated along with any further modifications.

### DIFF
--- a/docs/user_guide/13_annotation.rst
+++ b/docs/user_guide/13_annotation.rst
@@ -53,6 +53,19 @@ For long-term projects, we recommend using `Label Studio <https://labelstud.io/>
 
    Label Studio annotation platform.
 
+Exporting Annotations for DeepForest
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+To ensure compatibility with DeepForest, export your annotations from Label Studio in Pascal VOC XML format. This format is widely used for object detection tasks and can be directly read by DeepForest's ``read_pascal_voc`` function.
+
+Steps to Export in Pascal VOC Format:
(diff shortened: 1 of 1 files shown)


henrykironde:
@Samia35-2973, Thank you for your contribution! It looks like your branches haven't been updated, which has brought in work from other contributors. To keep things organized and ensure that only one person works on a given issue at a time, each PR should focus on a single issue. This helps avoid overlaps and keeps the process smooth. We appreciate your effort and look forward to your updated submission!

Issue #972, opened 2025-03-13
Docs: Add "Required Imports and Initialization" section for better onboarding


Abhishek-Dimri:
### Problem
While going through the documentation, I noticed that some pages lack clarity regarding **required imports** and **model initialization**. For example:
- Some code snippets assume prior knowledge of what needs to be imported.
- Model initialization steps are sometimes missing, making it harder for new users to follow along.

This isn’t a major issue for daily/active users or those who just want to check a specific section of code, but for newbies or those exploring the documentation for the first time, it can be confusing and time-consuming to figure out what’s missing.

### Proposed Solution
Add a small **"Required Imports and Initialization"** section at the beginning of each page or tutorial. This section should include:
1. **Required Imports**: All necessary libraries and modules.
2. **Model Initialization**: Any pre-requisite setup or initialization steps.

For example:
```python
# Required Imports
import numpy as np
import matplotlib.pyplot as plt
from deepforest import get_data
from deepforest import visualize

# Model Initialization
m = mymodule()
m.load_model("weecology/deepforest-tree") 
```

### Benefits
- **Improved Onboarding**: New users can quickly get started without guessing what to import or initialize.
- **Consistency**: All pages will follow a uniform structure, making the documentation easier to navigate.
- **Time-Saving**: Reduces the frustration of debugging missing imports or setup steps.
😊

Implementing this change might be a time-consuming process since it requires updates across multiple pages. To make it manageable, we could consider doing it page by page, prioritizing the most frequently visited or critical sections first.

ethanwhite:
I like this idea @Abhishek-Dimri and I personally find the lack of this costs me time periodically when I'm trying to do a quick test and I can't just copy-paste the relevant code block. Any improvements in this regard are definitely welcome as PRs!

naxatra2:
I was working on this issue. But I noticed that some files are in `.rst` format and some in `.md` 

![Image](https://github.com/user-attachments/assets/42670f2a-0964-4729-9826-1187e41c80b2)

it seems like their content isn’t different enough to justify using two separate formats. Wouldn’t it be better to keep them all in the same format for consistency? This would make managing the directory easier and also feel more organized.

Abhishek-Dimri:
@naxatra2, your opinion will also be valuable, so please feel free to share your thoughts on this approach!

naxatra2:
[...]

PR #981, opened 2025-03-20
Docs: Adds Required Imports and Initialization section in user_guide

naxatra2:
Adds #972
I have added the required import calls, in each code blocks that I found necessary. 

I was unsure about the prerequisite setups and model initializations, so I did not include much about them.


bw4sz:
I am ambivalent here. Maybe @henrykironde or @ethanwhite have an opinion whether each code block inside the doc is on its own reproducible.

docs/user_guide/09_configuration_file.md
```diff
@@ -104,7 +106,7 @@ Most commonly, `cpu`, `gpu` or `tpu` as well as other [options](https://pytorch-
 
 If `gpu`, it can be helpful to specify the data parallelization strategy. This can be done using the `strategy` arg in `main.create_trainer()`
 
-```
+```python
 model.create_trainer(logger=comet_logger, strategy="ddp")
```
ethanwhite, line 110: Since this is a new section we should add the import here as well and switch from `model` to `m` for consistency

docs/user_guide/09_configuration_file.md
```diff
@@ -144,13 +146,13 @@ Directory to search for images in the csv_file image_path column
 
 Learning rate for the training optimization. By default the optimizer is stochastic gradient descent with momentum. A learning rate scheduler is used based on validation loss
 
-```
+```python
 optim.SGD(self.model.parameters(), lr=self.config["train"]["lr"], momentum=0.9)
```
ethanwhite, line 150: Need to add `from torch import optim`?

docs/user_guide/09_configuration_file.md
```diff
@@ -144,13 +146,13 @@ Directory to search for images in the csv_file image_path column
...
 ```
 
 A learning rate scheduler is used to adjust the learning rate based on validation loss. The default scheduler is ReduceLROnPlateau:
 
-```
+```python
 self.scheduler = torch.optim.lr_scheduler.ReduceLROnPlateau(self.optimizer, mode='min',
```
ethanwhite, line 156: Add `import torch`?

ethanwhite (commented):
Really nice work @naxatra2! Three things to look at:

- In 01_Reading_data.md the changes related to `deepforest.utilities.read_file` have some issues. The initial changes don't work because the import `read_file` directly whereas the code that follows it runs `utilities.read_file` (i.e., the necessary import is `from deepforest import utilities`. Later changes switch to changing all code to importing `read_file` directly. The most important thing is that this should be consistent throughout. Since the PR is about the imports I lean towards using `from deepforest import utilities` throughout and leaving the calls to those functions untouched. This is consistent with submodule usage later in the docs.
- It looks like some similar changes should be added to `03_cropmodels.rst`
- I've made a bunch of inline comments related to other small fixes

docs/user_guide/11_training.md
```diff
@@ -23,6 +23,10 @@ OSBS_029.jpg,161,155,199,191,Tree
 The config file specifies the path to the CSV file that we want to use when training. The images are located in the working directory by default, and a user can provide a path to a different image directory.
 
 ```python
+import os
+from deepforest import main
```
henrykironde, line 27: Are we using main?

henrykironde (changes requested):
Thanks, @naxatra2,the changes. I've suggested a few additional places where we can make improvements.

naxatra2:
I have noticed that in the docs `m` is not used consistently. It is sometimes used for `model.py`  and sometimes for `main.py`. For example we sometimes use `m = main.deepforest()` but we are also using `from deepforest import model as m`. Having both of these represented by a single letter makes it a bit hard to understand I guess ?

ethanwhite:
Thanks @naxatra2 - this all looks good. Unfortunately I need to ask you to rebase one more time and it's going to be a little tricky due to #990 being merged. The trickiest piece is that 03_cropmodels was converted from rst to md. Sorry about that. The good news is that once this is done I think we're ready to merge!

PR #983, opened 2025-03-21
Allow predict_tile to take in a list of crop models

### LINKED ISSUE #982: Allow predict_tile to take in a list of cropmodels
Users often have nested classes they want to predict. For example, the object detector might detect birds, one classifier is for species, another for behavior. It would be useful to allow predict_tile(crop_model=[CropModel(), CropModel()]) and then iterate over models 

https://github.com/weecology/DeepForest/blob/12cd37e01a543f12b98ccfbc1f9b632110b9598d/src/deepforest/predict.py#L229

A couple thoughts jump to mind.

1. How to arrange the output dataframe, right now a new column is cropmodel_label and cropmodel_score are made. The simplest thing would be to return a list of cropmodel_labels as a value in that column, the user would need to then parse that data. Alittle ugly. Perhaps instead of a list, it could be a named dict as inputs, and then use those names as outputs?
2. Not all crops need to go to all models. It would be more efficient to specify the labels, or combination of labels that allow the crops to be passed to a model. This is probably a secondary concern, the user can always delete or filter the dataframe afterwards.

Abhishek-Dimri:
🔹 **Summary**  
This PR extends the `predict_tile` function to support multiple crop models, enabling users to pass a list of crop models (`crop_models=[...]`). The results now include separate columns for each crop model's predictions and confidence scores, enhancing flexibility and usability.

---

✨ **Key Changes**  
- **Modified `predict_tile`**:  
  - Now accepts a list of crop models and applies them iteratively.  
  - Ensures backward compatibility for single-model use cases.  
- **Updated `_predict_crop_model_`**:  
  - Dynamically handles multiple models for cleaner and more scalable code.  
- **Testing Improvements**:  
  - Updated `test_predict_tile_with_crop_model` to support multiple crop models.  
  - Added assertions to dynamically check columns (e.g., `cropmodel_label_0`, `cropmodel_score_0`, etc.).  
  - Included test cases for scenarios where predictions are empty.  

---

🔬 **Testing & Validation**  
- ✅ All existing tests passed.  
- ✅ Added new test cases for multiple crop models.  
- ✅ Verified edge cases, such as empty predictions, to ensure robustness.  

---

🖼️**Output Screenshot**
image

🔗 **Issue Fixed**  
- Fixes #982 (Support for multiple crop models in `predict_tile`).

### DIFF
--- a/docs/user_guide/03_cropmodels.rst
+++ b/docs/user_guide/03_cropmodels.rst
@@ -17,6 +17,8 @@ Why would you want to apply a model directly to each crop? Why not train a multi
 
 - **New Data and Multi-sensor Learning**: In many applications, the data needed for detection and classification may differ. The CropModel concept provides an extendable piece that allows for advanced pipelines.
 
+- **Support for Multiple Models**: You can pass a list of CropModel instances to `predict_tile`. This enables comparing predictions from multiple models, such as species classification and health assessment, in a single pass. Results include separate columns for each model's predictions and confidence scores (cropmodel_label_<i> and cropmodel_score_<i>).
+
 Considerations
 --------------
 
@@ -26,6 +28,9 @@ Considerations
 
 Usage
 -----
+
+Single Crop Model
+~~~~~~~~~~~~~~~~~
 
 Consider a test file with tree boxes and an 'Alive/Dead' label that comes with all DeepForest installations:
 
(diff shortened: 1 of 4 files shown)


bw4sz:
I appreciate the clear formatting of this PR.

bw4sz (changes requested):
Small change. Don't change the argument name to crop_models from crop_model. That will make a breaking change and require DeepForest 1.6.0, we avoid breaking changes to make users feel more comfortable. While crop_models is clearer, its not worth the break.

bw4sz (changes requested):
Great, almost complete.

1. The existing behavior should be allowed even if it means writing a bit more code. If a user passes a cropmodel, not in a list, that should be allowed. The logic on Line 618 can be updated to say, if instance is cropmodel, use that model, instead of expecting a user to pass a list of [CropModel()].
2. We need to update the docs so users know this is possible. Only needs a small amount of text and the output from the test to illustrate, and just add a line in a new subsection right after https://github.com/weecology/DeepForest/blob/main/docs/user_guide/03_cropmodels.rst#training
3. Once 1 is complete, update the docstring. Explaining that you can pass either a CropModel() or [CropModel(), CropModel()]

Thanks for your contribution!

Abhishek-Dimri:
Thanks for your appreciation, @bw4sz! I understand the changes needed and will update the logic to allow both single and multiple CropModel instances, along with updating the docs and docstring. I'll tag you once the changes are complete.

Abhishek-Dimri:
@bw4sz Maybe there is a misunderstanding. I checked and confirmed that users can already pass a single `CropModel` instance—it doesn't strictly expect a list. The existing logic ensures that both `CropModel()` and `[CropModel()]` work fine.  
### Code-
image
### Output-
image

### Improvement-
However, there might be an improvement: if a user passes only a single `CropModel` instance, instead of naming the columns `cropmodel_label_0` and `cropmodel_score_0`, we could simply name them `cropmodel_label` and `cropmodel_score`. Let me know what you think!

bw4sz:
> However, there might be an improvement: if a user passes only a single CropModel instance, instead of naming the columns cropmodel_label_0 and cropmodel_score_0, we could simply name them cropmodel_label and cropmodel_score. Let me know what you think!

I agree. My aim is here is to not have any behavioral change for current users. We want to add functionality without any potential breaking changes.

Abhishek-Dimri:
> I agree. My aim is here is to not have any behavioral change for current users. We want to add functionality without any potential breaking changes.

I understand! Can you please clarify the final changes that need to be done once again, just to ensure we're on the same page?

bw4sz:
I haven't checked the code, just restating from above, making sure that the docs and docstrings are complete, that the naming behavior of columns for a single cropmodel example doesn't change past behavior. Thanks!

Abhishek-Dimri:
Hey @bw4sz,

Sorry for the rebase issue and multiple commits. The change is done now, single crop model column names are cropmodel_label and cropmodel_score.

Screenshots:
Single CropModel-
image

Multiple CropModels-
image

I also added back the old tests for a single crop model. Additionally, I introduced a new parameter to track whether a single or multiple crop models are passed. This should be a small change and shouldn't cause any issues for users.

Thanks!

Abhishek-Dimri:
Hey Ben,

I made two key changes:

1- Added a parameter to handle column naming for single crop models (cropmodel_label and cropmodel_score instead of _0). This should be a small change and shouldn't cause issues for users.
### Screenshots-
Single CropModel- 
image

Multiple CropModels-
image

2- Restored the old tests for single crop model cases.

Let me know if this works or if any adjustments are needed!

henrykironde:
@Abhishek-Dimri can you update this PR please

bw4sz (approved):

Abhishek-Dimri:
Thanks for merging! 😊

Issue #1000, opened 2025-04-01
Let deepforest.utilities.image_to_geo_coordinates use root_dir attribute if it exists.


bw4sz:
Small improvement, since 1.4.0, the root_dir is attached to predictions as an attribute specifically for other functions to use, related to #961.

```
predictions = m.predict_tile(image=channel_last, patch_size=4000, patch_overlap=0)
>>> predictions.root_dir
'/Users/benweinstein/Downloads/data'
```

You shouldn't have to specify it when converting image_to_geo_coordinates unless it doesn't exist (this happens when user based a numpy array and not a file, so there is no way to know where it exists on disk, add to error message when not found). 

## Current behavior

```
>>> geopredictions = image_to_geo_coordinates(predictions)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: image_to_geo_coordinates() missing 1 required positional argument: 'root_dir'
```

## Desired behavior
```
>>> geopredictions = image_to_geo_coordinates(predictions)
```

Location: https://github.com/weecology/DeepForest/blob/2e0b699bb0e1d0cb418d6b5d8504675559e85496/src/deepforest/utilities.py#L565C31-L565C39

rhydham-nith:
Sir,

I had already raised a PR editing the `src/deepforest/utilities.py` file while working on issue #998. However, when I started working on this issue, I realized that it also requires modifications to the `src/deepforest/utilities.py` file. I was unsure whether I should update the existing PR with these changes or raise a new one. For now, I’ve decided to raise a new PR, as the changes are not related to each other, but I would appreciate your guidance on the ideal approach for such situations in the future.

PR #1002, opened 2025-04-01
Fix: Automatically handle root_dir in image_to_geo_coordinates function

rhydham-nith:
This PR addresses issue #1000 by modifying the `image_to_geo_coordinates` function to automatically handle the `root_dir` attribute from the predictions object. The function now checks if `root_dir` exists within the predictions and uses it if available. If `root_dir` is missing, the function raises a clear error.

Changes made:

1. Removed `root_dir` as a function parameter.
2. Implemented automatic detection of `root_dir` within the predictions object.

### DIFF
--- a/src/deepforest/utilities.py
+++ b/src/deepforest/utilities.py
@@ -542,7 +542,7 @@ def check_image(image):
                          "found image with shape {}".format(image.shape))
 
 
-def image_to_geo_coordinates(gdf, root_dir, flip_y_axis=False):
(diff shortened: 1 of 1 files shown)


ethanwhite:
Thanks for the contribution @rhydham-nith. Unfortunately it looks like we ended up merging a later PR that addressed the same issue #1011 before seeing this one. Sorry about that. We really do appreciate your work on this.

PR #1003, opened 2025-04-01
Update read_file and TreeDataset class to look at geometry column in shapefiles without geospatial reference data.

### LINKED ISSUE #992: Training model with annos throws Xmin, Ymin,... error
Hi, 

I'm running DeepForest in Google CoLab, and ran into an issue with the annotation file, I think.

To answer your User Story question, I'm a forest ecologist at Washington State University. I'm helping the City of Anacortes, WA to build a tree species map for ~3,000 acres of community owned forest land. The map will serve as the basis for a baseline assessment of forest health and a broader community led monitoring program.

Here's my issue. 

I'm trying to train the model. Attached are my annotations. Annotations were polygons. I converted them to annos using your tools. [WL_train_annotation.csv](https://github.com/user-attachments/files/19526200/WL_train_annotation.csv). I I can supply the image if needed, but I couldn't attach.

Here's my code to train the model

```
#initialize the model and change the corresponding config file
m = main.deepforest()
m.use_release()

#load the lastest release model
m.config["train"]["csv_file"] = '/content/drive/MyDrive/ACFL_Colab/Data/Imagery/model_training/WL_train_annotation.csv'
m.config["train"]["root_dir"] = '/content/drive/MyDrive/ACFL_Colab/Data/Imagery/model_training/'
m.config["score_thresh"] = 0.4
m.config["train"]['epochs'] = 2 #increase to 20 when no bugs
m.config["train"]["lr"] = 0.00001 #learning rates:currently low but vary this
m.config["train"]["fast_dev_run"] = True #for debugging

# Initialize the trainer
m.trainer.fit(m) #this line is throwing the error
```

**Here is the error that I'm getting**

KeyError: "None of [Index(['xmin', 'ymin', 'xmax', 'ymax'], dtype='object')] are in the [columns]"

I think this means that I need to have columns in my annotations file for xmin, ymin, etc... Do you agree or is there some other issue? 

Thank you!

bw4sz:
# What this PR does.

- Updates read_file to only convert to geocoordinates if the shapefile in question has a geospatial crs. 
- Updates dataset.TreeDataset to grab the annotations from the geometry column as we migrate towards multi-geometry setup.
- Adds a test to ensure behavior and replicates the errors in #992.

# Closes

closes #992 

My only lingering doubt was whether to fully delete the if statement and just insist all inputs have a geometry column, that would have been backward breaking and require 1.6.0, which didn't seem wholly needed and could be avoided.

In general this is an edge case, but points to overall themes of trying to improve data import, which is always the hardest part, and the conflict between image coordinate and geospatial coordinate system, which underly many issues.

### DIFF
--- a/src/deepforest/dataset.py
+++ b/src/deepforest/dataset.py
@@ -28,6 +28,7 @@ from rasterio.windows import Window
 from torchvision import transforms
 import slidingwindow
 import warnings
+import shapely
 
 
 def get_transform(augment):
@@ -105,8 +106,12 @@ class TreeDataset(Dataset):
             image_annotations = self.annotations[self.annotations.image_path ==
                                                  self.image_names[idx]]
             targets = {}
-            targets["boxes"] = image_annotations[["xmin", "ymin", "xmax",
-                                                  "ymax"]].values.astype("float32")
+
+            if "geometry" in image_annotations.columns:
+                targets["boxes"] = np.array([shapely.wkt.loads(x).bounds for x in image_annotations.geometry]).astype("float32")
+            else:
+                targets["boxes"] = image_annotations[["xmin", "ymin", "xmax",
+                                                      "ymax"]].values.astype("float32")
 
             # Labels need to be encoded
             targets["labels"] = image_annotations.label.apply(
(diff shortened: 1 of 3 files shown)


ethanwhite (approved):
LGTM

Issue #997, opened 2025-04-01
Improve deepforest.utilities.read_file for reading a .shp that doesn't have a image_path column


bw4sz:
Reading in data is often the hardest thing for users, we want to help them as much as possible at this step.

```
>>> ground_truth = read_file("/Users/benweinstein/Downloads/data/individual_align.shp")
m = main.deepforest()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/benweinstein/Documents/DeepForest/src/deepforest/utilities.py", line 331, in read_file
    df = shapefile_to_annotations(input, root_dir=root_dir)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/benweinstein/Documents/DeepForest/src/deepforest/utilities.py", line 184, in shapefile_to_annotations
    raise ValueError(
ValueError: No image_path column found in shapefile, please specify rgb path
```
1. We should allow users to pass an image_path to read_file. We currently have a root_dir argument, but no image_path argument.
2. This assumes that _all_ annotations in the file relate to the image_path, add a warning to make it clear.
3. The same is true for the label column, should be an argument to read_file.
4. Write tests for shapefiles, this is the common place we see this.
## Current workaround

```
raster_path = "/Users/benweinstein/Downloads/data/SPA25_nogcp_crossrotate_aligned.tif"
gdf = gpd.read_file("/Users/benweinstein/Downloads/data/individual_align.shp")
gdf["image_path"] = os.path.basename(raster_path)
gdf["label"] = "Tree"
ground_truth = read_file(gdf, root_dir = os.path.dirname(raster_path))
```
This requires two unneeded library imports (os and geopandas) and supplies the root dir. The full versus relative path is an ongoing debate in DeepForest API and should not addressed here.

SuMayaBee:
Hello, @bw4sz . I've submitted PR #1242 that addresses this issue.

**Changes:**
- Added `image_path` and `label` arguments to `read_file()`
- Added warnings when these arguments are passed
- Added 4 tests for shapefiles
- Added documentation example

**Now users can do:**
```python
ground_truth = read_file("annotations.shp", image_path="image.tif", label="Tree")
```
Instead of the previous workaround requiring `os` and `geopandas` imports.

PR #1004, opened 2025-04-01
Improving read_file function

rhydham-nith:
I've tried to address the issue #997 of `read_file`  for reading a .shp that doesn't have a `image_path` column. 

### Changes Made:

1. Added `image_path` and `label` as optional arguments in `read_file`.
2. Improved type handling using `isinstance()` for better input validation.

**Question:** Should I try to update the documentation about how to use the `read_file` Function

I would be happy to contribute more to the project and am open to any feedback or suggestions for improvement. Thank you for the opportunity to collaborate!

### DIFF
--- a/docs/user_guide/01_Reading_data.md
+++ b/docs/user_guide/01_Reading_data.md
@@ -2,6 +2,24 @@
 
 The most time-consuming part of many open-source projects is getting the data in and out. This is because there are so many formats and ways a user might interact with the package. DeepForest has collated many use cases into a single `read_file` function that will attempt to read many common data formats, both projected and unprojected, and create a dataframe ready for DeepForest functions.
 
+You can also optionally provide:
+  - `image_path`: A single image path to assign to all annotations in the input. This is useful when the input contains annotations for only one image.
+  - `label`: A single label to apply to all rows. This is helpful when all annotations share the same label (e.g., "Tree").
+
+Example:
+```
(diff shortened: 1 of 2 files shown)


rhydham-nith:
Hi @bw4sz  @ethanwhite @henrykironde ,
This PR has passed all the Continuous Integration tests successfully and is ready for review. Whenever you get a chance, could you please take a look and provide me with your feedback? Looking forward to your thoughts!

Thanks in advance!

bw4sz (changes requested):
Thanks for your contribution, great start. 

1. Update the docstrings
2. Make it clear in both docstrings and as a warning to users that by passing a label or image_path column, that is being applied to the entire dataframe. For example, if I had a dataframe containing annotations from 10 different images, I wouldn't want to pass the image_path argument, this is only in the case of a single image's worth of annotations.
3. Update deepforest docs to show this behavior, with the same text https://github.com/weecology/DeepForest/blob/cda5a907d6a349c6074689e57ad69389ba415b9c/docs/user_guide/01_Reading_data.md?plain=1#L188

rhydham-nith:
Sir, 
I’ve made all the desired updates — updated docstring, added example usage, warning, and updated the `.md` file accordingly. Please let me know the next steps. 
Thank you

bw4sz (approved):

PR #1006, opened 2025-04-02
Checking for empty dataset

### LINKED ISSUE #867: Better sanity checking to check validation is not empty.
It is possible to pass a empty dataframe to validation and get meaningless errors. Add a validation check before load_dataset so that trainer.fit cannot throw errors later on.

reddykkeerthi:
Added a validation check before load_dataset(...) so that trainer.fit(..) cannot throw errors when an empty dataframe is passed. 

Fixes #867

### DIFF
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -300,6 +300,8 @@ class deepforest(pl.LightningModule, PyTorchModelHubMixin):
                                  transforms=self.transforms(augment=augment),
                                  label_dict=self.label_dict,
                                  preload_images=self.config["train"]["preload_images"])
+        if len(ds) == 0:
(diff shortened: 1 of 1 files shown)


bw4sz:
Thanks for this, can you confirm here the functionality by making an empty .csv and asserting it raises the error.

reddykkeerthi:
> Thanks for this, can you confirm here the functionality by making an empty .csv and asserting it raises the error.

I created a `custom.csv` file that has only column names and no entries (as confirmed in the terminal output). When I used the `load_dataset()` function, it raised a `ValueError`, which was displayed in the terminal.

image

bw4sz:
Great, running CI.

bw4sz (approved):
Thank you.

PR #1011, opened 2025-04-03
Fix #1000: Improve deepforest.utilities.image_to_geo_coordinates to Use root_dir Attribute if Available

### LINKED ISSUE #1000: Let deepforest.utilities.image_to_geo_coordinates use root_dir attribute if it exists.

Samia35-2973:
This PR enhances the `deepforest.utilities.image_to_geo_coordinates` by ensuring that the `root_dir` attribute is automatically inferred from the geodataframe if it exists. If `root_dir` cannot be determined, a clear error message is provided for better debugging.

**Changes**:
- Updated `image_to_geo_coordinates()` logic in `src/deepforest/utilities.py`. Previously, users had to manually specify `root_dir`. Now, the function attempts to infer it from the geodataframe. If `root_dir` is missing, an explicit error message is raised to guide users.
- Modified test cases:
  - In `test_preprocess.py`, updated `test_split_raster_from_shp()` to avoid explicitly assigning `root_dir`, ensuring expected behavior.
  - In `test_utilities.py`, updated occurrences of `image_to_geo_coordinates()` to align with the new logic.
- Updated documentation:
  - Modified `docs/user_guide/15_Writing_data.md` to reflect the new behavior.

✅ All existing tests passed.

**Open Question:** I think in here
https://github.com/weecology/DeepForest/blob/1acab8b8169ed929e89eda80de38642b235071aa/tests/test_utilities.py#L358
a similar case occurs where the geodataframe already has a `root_dir` attribute. Should we also update `visualize.py` and other similar occurrences where `root_dir` is explicitly passed but already available in the dataframe?

- Fixes #1000

### DIFF
--- a/citation_count.json
+++ b/citation_count.json
@@ -1,6 +1,6 @@
 {
     "schemaVersion": 1,
     "label": "Citations",
-    "message": "88",
+    "message": "90",
     "color": "blue"
 }
--- a/docs/user_guide/15_Writing_data.md
+++ b/docs/user_guide/15_Writing_data.md
@@ -25,7 +25,7 @@ path_to_raster = get_data("2018_SJER_3_252000_4107000_image_477.tif")
 src = rio.open(path_to_raster)
 original = utilities.read_file(annotations)
 
-geo_coords = utilities.image_to_geo_coordinates(original, root_dir=os.path.dirname(path_to_raster))
+geo_coords = utilities.image_to_geo_coordinates(original)
 src_window = geometry.box(*src.bounds)
 
 fig, ax = plt.subplots(figsize=(10, 10))
--- a/src/deepforest/utilities.py
+++ b/src/deepforest/utilities.py
@@ -542,18 +542,29 @@ def check_image(image):
                          "found image with shape {}".format(image.shape))
 
 
-def image_to_geo_coordinates(gdf, root_dir, flip_y_axis=False):
+def image_to_geo_coordinates(gdf, root_dir=None, flip_y_axis=False):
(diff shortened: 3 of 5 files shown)


bw4sz:
This looks good and is ready to merge, just delete the added .citation_count.json, must have been in the repo? Not sure what it is. We can handle other cases in other PRs, we like small PRs, especially when there are so many open, less rebasing.

ethanwhite:
The citation count file gets rebuilt automatically when there are updates in the number of citations. Fine to delete it but I think it's also OK to leave it it for now. I'll open an new issue about how we handle it going forward.

PR #1001, opened 2025-04-01
Fix: Alpha Channel Issue in predict_tile() by Adding read_tile() Utility

rhydham-nith:
This PR addresses an issue #998 where .tif raster images containing an alpha (transparency) channel caused DeepForest to throw an error due to the model expecting only 3-band (RGB/BGR) input. 

### Key Changes:

1. Added read_tile() function to automatically remove the alpha channel (if present) and ensure a 3-band image.
2. Converted raster data from channel-first (CHW) to channel-last (HWC) format using np.moveaxis().
3. Replaced direct rio.open(path).read() calls in predict_tile() with read_tile(path).

I am open to any feedback or suggestions for improvement. Looking forward to your guidance!

PR #1016, opened 2025-04-05
Adds a simple utility to strip and convert .tif file with alpha channels

### LINKED ISSUE #998: Add a simple utility to strip and convert .tif file with alpha channels.
Lots of geoprocessing tools export .tif files with alpha channels. DeepForest models work with 3 band data (for the moment).

The error messaging is somewhat confrontational, we always want errors to lead users towards a solution.

```
>>> predictions = m.predict_tile(raster_path, patch_size=2000, patch_overlap=0)
  File "/Users/benweinstein/Documents/DeepForest/src/deepforest/main.py", line 550, in predict_tile
    ds = dataset.TileDataset(tile=image,
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/benweinstein/Documents/DeepForest/src/deepforest/dataset.py", line 164, in __init__
    raise ValueError(
ValueError: Only three band raster are accepted. Channels should be the final dimension. Input tile has shape (16868, 14148, 4). Check for transparent alpha channel and remove if present
```

let's make a deepforest.utilities.read_tile function that wraps

```
with rasterio.open(raster_path) as src:
    image = src.read()
    # Remove alpha if present (4th channel)
    if image.shape[0] == 4:
        image = image[:3]    
```

also could check that its channel last, if not.

```
channel_last = np.moveaxis(image, 0, -1)
```

There is some complexity here, since rasterio automatically moves things to channel first, but PIL doesn't, which is used in general since its faster in deepforest.dataset.

naxatra2:
Fixes #998
Added a new function, `utilities.read_tile(raster_path)`, that uses rasterio to read a raster and automatically remove an alpha channel if present.
- Reordered dimensions from channel-first to channel-last for consistency with DeepForest’s three-band input requirement.
- Created 4 Pytest tests to check that read_tile handles 3-band, 4-band, and invalid band count data properly.
- Ensured the output shape is (height, width, channels) and that it can be used directly with PIL and other NumPy-based pipelines in DeepForest.

### DIFF
--- a/src/deepforest/utilities.py
+++ b/src/deepforest/utilities.py
@@ -755,10 +755,9 @@ def project_boxes(df, root_dir, transform=True):
     return df
 
 
-def read_tile(raster_path):
+def remove_alpha_channel(raster_path):
     """Read a raster from disk, remove alpha channel if present, and reorder
-    dimensions to channel-last (H, W, C) for consistency with DeepForest input
-    expectations.
+    dimensions to channel-last (H, W, C) for consistency.
 
     Args:
         raster_path (str): Path to the GeoTIFF (or other supported raster format).
@@ -770,16 +769,11 @@ def read_tile(raster_path):
     with rasterio.open(raster_path) as src:
         image = src.read()
 
-    if image.shape[0] == 4:
-        warnings.warn(
-            "Detected an alpha channel, removing it to match DeepForest's requirement of 3 bands."
-        )
-        image = image[:3]
-
-    # Confirm we now have 3 bands
-    if image.shape[0] != 3:
-        raise ValueError(
-            f"Expected 3 bands, but got {image.shape[0]}. Please check your raster.")
+        if image.shape[0] == 4:
+            warnings.warn(
+                "Detected an alpha channel, removing it to match DeepForest's requirement of 3 bands."
+            )
+            image = image[:3]
 
     image = np.moveaxis(image, 0, -1)
 
--- a/src/deepforest/utilities.py
+++ b/src/deepforest/utilities.py
@@ -753,3 +753,34 @@ def project_boxes(df, root_dir, transform=True):
         "This function is deprecated. Please use image_to_geo_coordinates instead.")
 
     return df
+
+
+def read_tile(raster_path):
+    """Read a raster from disk, remove alpha channel if present, and reorder
+    dimensions to channel-last (H, W, C) for consistency with DeepForest input
+    expectations.
+
+    Args:
+        raster_path (str): Path to the GeoTIFF (or other supported raster format).
+
+    Returns:
+        numpy.ndarray: A 3D numpy array of shape (H, W, C).
+    """
+
+    with rasterio.open(raster_path) as src:
+        image = src.read()
+
+    if image.shape[0] == 4:
+        warnings.warn(
+            "Detected an alpha channel, removing it to match DeepForest's requirement of 3 bands."
+        )
+        image = image[:3]
+
+    # Confirm we now have 3 bands
+    if image.shape[0] != 3:
+        raise ValueError(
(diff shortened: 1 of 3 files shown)


src/deepforest/utilities.py
```diff
@@ -753,3 +753,34 @@ def project_boxes(df, root_dir, transform=True):
...
+        raster_path (str): Path to the GeoTIFF (or other supported raster format).
+
+    Returns:
+        numpy.ndarray: A 3D numpy array of shape (H, W, C).
+    """
+
+    with rasterio.open(raster_path) as src:
+        image = src.read()
+
+    if image.shape[0] == 4:
```
henrykironde, line 773: This section should be placed inside the with rasterio context. Also, the function name read_tile doesn't accurately reflect its purpose. A more suitable name might be remove_alpha_channel, which would make the docstring clearer—for example: "Removes the alpha channel and returns a 3D NumPy array with shape (H, W, C)."

src/deepforest/utilities.py
```diff
@@ -753,3 +753,34 @@ def project_boxes(df, root_dir, transform=True):
...
+        warnings.warn(
+            "Detected an alpha channel, removing it to match DeepForest's requirement of 3 bands."
+        )
+        image = image[:3]
+
+    # Confirm we now have 3 bands
+    if image.shape[0] != 3:
```
henrykironde, line 780: I don't think this is required.

src/deepforest/visualize.py
```diff
@@ -517,11 +517,11 @@ def plot_results(results,
         image_path = os.path.join(root_dir, results.image_path.unique()[0])
         image = np.array(Image.open(image_path))
 
-        # Drop alpha channel if present and warn 
+        # Drop alpha channel if present and warn
         if image.shape[2] == 4:
```
henrykironde, line 521: We should be able to use the utility we created to verify that the image exists, is not empty, and has at least 3 channels. If it has more than 3 (e.g., an alpha channel), we can use drop the extra channel.

henrykironde (changes requested):
@naxatra2 thanks for adding this utility. I have made some suggestions let me know what you think.

henrykironde:
@rhydham-nith @naxatra2 ref: https://github.com/weecology/DeepForest/pull/1001  ref: https://github.com/weecology/DeepForest/pull/1016

I recommend that both of you combine your contributions into a single PR. Please review PRs 1001 and 1016, then create one combined PR and include both of you as co-authors.

src/deepforest/utilities.py
naxatra2: I am naming the function as `remove_alpha_channel`

naxatra2:
I have created a new test case to replace the 4 tests that I created earlier, I think they are easier to read and maintain now.

naxatra2:
The CI is failing on tests that I didn't change, I think it is a server issue

rhydham-nith:
Hey @naxatra2 
I was trying to push to the Branch tif-alpha that you created. However, I am getting access denied message. Could you please check into the permissions.

Screenshot 2025-04-09 153705


Although I wasn't able to push directly to your branch due to access restrictions, I was able to clone their changes into my fork. I've made my edits and pushed them in it.
So technically, my [branch ](https://github.com/rhydham-nith/DeepForest/tree/tif_alpha) now contains the combined changes from both you and me. Let me know how shall we proceed further?

ethanwhite:
@rhydham-nith - you should be able to submit a PR to @naxatra2's branch

henrykironde:
@naxatra2 and @rhydham-nith Thanks for your contribution to this issue. It looks like the current codebase no longer exhibits the problem—Ben added a line to address it. We're closing this PR and the associated issue for now, but feel free to reopen them if the problem reappears. closes #1001 #998

PR #1018, opened 2025-04-07
add try for common rasterio error on arcgis

bw4sz:
We are seeing widespread errors that are interrupting CI, its a rare part of the codebase. I minimized the test and added a common try catch loop, which I think is worth the decreased coverage.

### DIFF
--- a/src/deepforest/utilities.py
+++ b/src/deepforest/utilities.py
@@ -313,7 +313,7 @@ def read_file(input, root_dir=None, image_path=None, label=None):
     """Read a file and return a geopandas dataframe.
 
     This is the main entry point for reading annotations into deepforest.
-    
+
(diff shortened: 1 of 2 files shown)


ethanwhite:
@bw4sz - there are tests failing on this, but I don't immediately see why

bw4sz:
Test passes locally, cannot reproduce. Rerunning jobs.

bw4sz:
Torchmetrics version update?

Locally, passes at 1.6.3, fails 1.7.2

Can we merge this and open a separate pull request and issue, since main is already 'failing' once this hits.

ethanwhite (approved):
LGTM. Errors are independent and will be fixed by @bw4sz shortly

PR #1029, opened 2025-04-15
Added default labels for each model

### LINKED ISSUE #1027: User reports that default label_dict in the livestock model still says 'tree'
Image

Can we reproduce? What does the m.label_dict look like?

reddykkeerthi:
Fixes #1027 
Previously, the `load_model()` function only had a default label set for `deepforest-bird` as `'Bird'`. For all other models, the default label `'Tree'` was displayed. By explicitly specifying the label for each model, I was able to assign appropriate labels as follows:

- `deepforest-tree`: `'Tree'`  
- `deepforest-livestock`: `'Livestock'`  
- `deepforest-bird`: `'Bird'`  
- `everglades-nest-detection`: `'Nest'`  
- `cropmodel`: default label remains `'Tree'`

After making these changes and re-running the same piece of code, I obtained the expected output.

image

### DIFF
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -146,24 +146,9 @@ class deepforest(pl.LightningModule, PyTorchModelHubMixin):
         # Load the model using from_pretrained
         self.create_model()
         loaded_model = self.from_pretrained(model_name, revision=revision)
+        self.label_dict = loaded_model.label_dict
         self.model = loaded_model.model
-
-        # Set label_dict and other configurations based on the model_name
-        if model_name == "weecology/deepforest-tree":
-            self.label_dict = {"Tree": 0}
-        elif model_name == "weecology/deepforest-bird":
-            self.label_dict = {"Bird": 0}
-            self.config['retinanet']["score_thresh"] = 0.3  # Bird-specific threshold
-        elif model_name == "weecology/deepforest-livestock":
-            self.label_dict = {"Livestock": 0}
-        elif model_name == "weecology/everglades-nest-detection":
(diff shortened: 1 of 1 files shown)


ethanwhite:
Thanks for this @reddykkeerthi. As I just noted in the associate issue, I think the correct solution here is actually to update the `label_dict` value in the `config.json` files on hugging face. Here's the file for the live stock model:

https://huggingface.co/weecology/deepforest-livestock/blob/main/config.json

Others can be found by navigating the models at:
https://huggingface.co/weecology

We would definitely welcome PR's over there for fixes and this PR could be updated to move the special settings for the bird model out of the code base and into the `config.json`.

Let me know if you have any questions about this

bw4sz:
The crop models are classification models that get cropmodel_labels, they do not interact with the label_dict. This issue should only deal with the detection models.

bw4sz:
Can this be closed in favor of what's happening on huggingface?

ethanwhite:
Thanks for your work on this @reddykkeerthi and for the improvements on huggingface! I think the history of the PR here is getting a little confusing so I'm going to outline some next steps and then close this one:

1. Regarding the removal of the bird model specific code - first we need to make sure that the additional settings (`score_thresh`) in that code are properly included in the huggingface config and test to make sure that those settings are getting properly included when bird model code is run. Once you've submitted and I've merged a huggingface PR for the bird model then open a new PR here to remove those settings.
2. As @bw4sz notes the two models you mentioned are separate and should just be left alone for now, but checking the rest of the detection models to see if the `label_dict` is correct and submitting hugging face PRs if they aren't would be great.

Issue #1030, opened 2025-04-15
Missing model initialization in documentation example


rhydham-nith:
Hi team
While working with DeepForest, I encountered an issue where calling model.config[...] without first creating an instance results in the following error:
``` python
AttributeError: module 'deepforest.model' has no attribute 'config'
```
This happens because at the line  https://github.com/weecology/DeepForest/blob/789974d744548e4eace0ca1b7fb4b34ac027de54/docs/user_guide/11_training.md?plain=1#L25 in the docs, the provided example
```python
import os
from deepforest import model as m
from deepforest import get_data

# Example run with short training
annotations_file = get_data("testfile_deepforest.csv")

m.config["epochs"] = 1
m.config["save-snapshot"] = False
m.config["train"]["csv_file"] = annotations_file
m.config["train"]["root_dir"] = os.path.dirname(annotations_file)

m.create_trainer()
```
 don't include the necessary initialization line:

``` python
model = main.deepforest()
```
This line is crucial to correctly instantiate the DeepForest model before accessing .config, .trainer, or using training/inference methods.

## Suggestion:
I propose updating the relevant examples to include the required line:
```python
import os
from deepforest import model as m
from deepforest import get_data

# Example run with short training
annotations_file = get_data("testfile_deepforest.csv")

model = main.deepforest()

m.config["epochs"] = 1
m.config["save-snapshot"] = False
m.config["train"]["csv_file"] = annotations_file
m.config["train"]["root_dir"] = os.path.dirname(annotations_file)

m.create_trainer()
```
This would help avoid confusion, especially for new users.

PR #1031, opened 2025-04-15
added model instance

rhydham-nith:
I’ve made the suggested change from this issue #1030  in this Pull Request.
If you feel it’s a good addition, you can accept the change.
Thank You!

### DIFF
--- a/docs/user_guide/11_training.md
+++ b/docs/user_guide/11_training.md
@@ -24,11 +24,14 @@ The config file specifies the path to the CSV file that we want to use when trai
 
 ```python
 import os
-from deepforest import model as m
+from deepforest import main
 from deepforest import get_data
 
 # Example run with short training
 annotations_file = get_data("testfile_deepforest.csv")
+
+# Initialize a DeepForest model instance to access configuration and training methods
+m = main.deepforest()
 
 m.config["epochs"] = 1
 m.config["save-snapshot"] = False


ethanwhite (changes requested):
Thanks for catching this issue with the docs @rhydham-nith!

I think there are a couple of things that still need to be fixed:
1. The import on Line 27 should be `from deepforest import main`
2. The name of the variable should be `m` not `model`

If you can get these fixed up and confirm that the whole block runs successfully in a fresh environment we'll be all set.

rhydham-nith:
In commit [cb1371c](https://github.com/weecology/DeepForest/pull/1031/commits/cb1371c0b625c701ea44aca399ad52fef359fefe), I have updated the model name consistently throughout the file. If this change doesn't align with the intended convention or style, please let me know. I'm happy to revise it as needed.

ethanwhite (changes requested):
I think we should keep this PR focused on fixing the bug in the docs. We typically want to fix one thing at time to make it easy to follow the history. So let's do the following. In this PR:

1. Fix the issue you originally identified, but not changing any naming throughout the doc (just use `m` in the fix since that's what is already there. 
2. Rebase the PR so that it only has one commit (let me know if it would be helpful for me to walk you through this process).

Then in a separate PR look let's homogenize the use of `model` vs. `m` throughout the the docs. I agree with you that `model` is easier to understand, so let's go with that change, but in that PR other folks will have the opportunity to chime in and we'll see what the consensus is on the naming.

rhydham-nith:
@ethanwhite 
I have made the requested changes. Looking forward to your guidance on homogenizing the use of `m` vs `model` throughout the repository in a follow-up PR. Let me know if there's anything else needed on this one!
Thank you once again for your support and guidance

docs/user_guide/11_training.md
```diff
@@ -24,18 +24,21 @@ The config file specifies the path to the CSV file that we want to use when trai
...
 
 # Example run with short training
 annotations_file = get_data("testfile_deepforest.csv")
 
-m.config["epochs"] = 1
-m.config["save-snapshot"] = False
-m.config["train"]["csv_file"] = annotations_file
-m.config["train"]["root_dir"] = os.path.dirname(annotations_file)
+# Initialize a DeepForest model instance to access configuration and training methods
+model = main.deepforest()
```
henrykironde, line 34:
```python
change to 
m = main.deepforest()
m.config["epochs"] = 1
m.config["save-snapshot"] = False
m.config["train"]["csv_file"] = annotations_file
m.config["train"]["root_dir"] = os.path.dirname(annotations_file)
.
.
......
```

rhydham-nith: @henrykironde 
I have made the suggested change. I wanted to ask—should I rebase this again into a single commit?
Please let me know if any further revisions are needed. I’m happy to help and looking forward to contributing more.

ethanwhite: @rhydham-nith yes, please rebased again into a single commit. This helps keep the history cleaner. While you're doing that let's also improve the commit message to something that fully explains what is being done "Add missing creation of model instance to docs example"

PR #1033, opened 2025-04-16
add video link

bw4sz:
For the workshop I made a walkthrough of predicting a tile with the bird detector.

### DIFF
--- a/citation_count.json
+++ b/citation_count.json
@@ -1,6 +1,6 @@
 {
     "schemaVersion": 1,
     "label": "Citations",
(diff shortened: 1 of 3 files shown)


ethanwhite (changes requested):
Can you rebase to a single commit and then this should be ready to go. The video isn't previewing properly in the readthedocs test environment, but I don't see anything different from other Loom embeds, so hopefully that's just a hiccup.

ethanwhite (approved):

Issue #1043, opened 2025-04-25
Add links to ArcGIS and QGIS tree plugins


bw4sz:
We should alert users to

https://www.arcgis.com/home/item.html?id=4af356858b1044908d9204f8b79ced99

and 

https://github.com/afruizh/TreeEyed

maybe goes above https://deepforest.readthedocs.io/en/v1.5.0/getting_started/comparison.html?

PR #1044, opened 2025-04-25
Add links to ArcGIS and QGIS tree plugins to documentation

rhydham-nith:
This PR addresses Issue #1043  and enhances the documentation by adding references to external tools and platforms.

Newly Added References:
- [ArcGIS Tree Crown Delineation Tool](https://www.arcgis.com/home/item.html?id=4af356858b1044908d9204f8b79ced99)
- [TreeEyed for QGIS](https://github.com/afruizh/TreeEyed)

Let me know if further refinements or additions are needed to move this forward!

### DIFF
--- a/docs/getting_started/comparison.md
+++ b/docs/getting_started/comparison.md
@@ -2,7 +2,30 @@
 
 There are many open-source projects for training machine learning models. 
 DeepForest aims to complement these existing tools by providing a specialized, streamlined approach tailored to ecological and environmental monitoring tasks. 
+This document also includes an overview of the models that have been built using the DeepForest framework.
 Below, we compare DeepForest with other notable tools in this space, highlighting similarities, differences, and areas of potential collaboration.
+
+## Tools Built on DeepForest
+
+DeepForest's core functionality has been successfully integrated into a variety of external platforms and tools. These implementations enable users to leverage DeepForest’s advanced tree crown detection capabilities across different GIS environments, all while maintaining the same model architecture and prediction performance.
+
+### ArcGIS Tree Crown Delineation Tool
+
+Available in the [ArcGIS Marketplace](https://www.arcgis.com/home/item.html?id=4af356858b1044908d9204f8b79ced99), this specialized plugin brings DeepForest's tree crown detection strengths into the ArcGIS Pro environment:
+
+- Provides a seamless, native ArcGIS interface for running DeepForest predictions and analyses
+- Fully integrates with Esri's comprehensive geospatial workflows, tools, and data management systems
+- Requires an active, licensed ArcGIS Pro installation to operate
+- Ideal for organizations and institutions already working extensively within the Esri ecosystem
+
+### TreeEyed for QGIS
+
+Available on [GitHub](https://github.com/afruizh/TreeEyed), this open-source QGIS plugin incorporates DeepForest together with additional models for enhanced tree crown analysis:
+
+- Offers a flexible, open-source solution for tree crown delineation tasks
+- Operates directly within the QGIS environment, providing an intuitive graphical user interface
+- Combines DeepForest with supplementary segmentation, tracking, and refinement capabilities
+- Suitable for users and researchers preferring robust, open-source GIS solutions without licensing restrictions
 
 ## Similar Tools
 


ethanwhite:
Since these tools are both DeepForest under the surface it feels a little strange to add them to the comparisons document and talk about them as if they are independent. These are  ways of running DeepForest (and in the case of TreeEyed; a couple of other models) in other environments.

I think what Ben as suggesting is a separate section to the top of this document (above "Similar Tools") named something like "Tools Built on DeepForest" where this information goes, but instead of the compare/contrast approach it describes that these are ways of running DeepForest in ArcGIS and QGIS.

Eventually we may want add this to the existing "Using DeepForest from R" page and update it's name to "Using DeepForest from R, ArcGIS, & QGIS". However this second approach could require a change since we're generating that file from another repo, so let's move forward with the first approach.

rhydham-nith:
Thanks for the feedback! I agree—I’ll restructure this as a "Tools Built on DeepForest" section instead, removing the comparisons and clarifying these are DeepForest integrations. (I was also a bit unsure initially since the original issue description was brief.)
I’ll push the updates shortly.

rhydham-nith:
### Quick question : 
Regarding the introductory paragraph under "Comparison with Other Tools" https://github.com/weecology/DeepForest/blob/8404c6569d946a75ebde88bb5499cf4138f0bd8f/docs/getting_started/comparison.md?plain=1#L5 would you like me to update it to better reflect the new structure?

**Proposed revision:**
This document covers two categories of tools:

- Tools Built on DeepForest

- Similar Tools

ethanwhite:
> would you like me to update it to better reflect the new structure?

Yes, that would be great. Thanks!

docs/getting_started/comparison.md
```diff
@@ -2,8 +2,31 @@
...
 There are many open-source projects for training machine learning models. 
 DeepForest aims to complement these existing tools by providing a specialized, streamlined approach tailored to ecological and environmental monitoring tasks. 
+This document also includes an overview of the models that have been built using the DeepForest framework.
 Below, we compare DeepForest with other notable tools in this space, highlighting similarities, differences, and areas of potential collaboration.
 
+## Tools Built on DeepForest
+
+DeepForest's core functionality has been successfully integrated into a variety of external platforms and tools. These implementations enable users to leverage DeepForest’s advanced tree crown detection capabilities across different GIS environments, all while maintaining the same reliable model architecture and prediction performance.
+
+**ArcGIS Tree Crown Delineation Tool**
```
ethanwhite, line 12: Should be a level 3 heading instead of bolded. This helps with both direct linking and accessibility.

docs/getting_started/comparison.md
```diff
@@ -2,8 +2,31 @@
...
+
+Available in the [ArcGIS Marketplace](https://www.arcgis.com/home/item.html?id=4af356858b1044908d9204f8b79ced99), this specialized plugin brings DeepForest's tree crown detection strengths into the ArcGIS Pro environment:
+
+- Provides a seamless, native ArcGIS interface for running DeepForest predictions and analyses
+- Fully integrates with Esri's comprehensive geospatial workflows, tools, and data management systems
+- Requires an active, licensed ArcGIS Pro installation to operate
+- Ideal for organizations and institutions already working extensively within the Esri ecosystem
+
+**TreeEyed for QGIS**
```
ethanwhite, line 21: Should be a level 3 heading instead of bolded. This helps with both direct linking and accessibility.

docs/getting_started/comparison.md
```diff
@@ -2,8 +2,31 @@
 
 There are many open-source projects for training machine learning models. 
 DeepForest aims to complement these existing tools by providing a specialized, streamlined approach tailored to ecological and environmental monitoring tasks. 
+This document also includes an overview of the models that have been built using the DeepForest framework.
 Below, we compare DeepForest with other notable tools in this space, highlighting similarities, differences, and areas of potential collaboration.
 
+## Tools Built on DeepForest
+
+DeepForest's core functionality has been successfully integrated into a variety of external platforms and tools. These implementations enable users to leverage DeepForest’s advanced tree crown detection capabilities across different GIS environments, all while maintaining the same reliable model architecture and prediction performance.
```
ethanwhite, line 10: Let's drop the word "reliable" here, since that's a judgement call and just say "same reliable model architecture and prediction performance"

ethanwhite (changes requested):
Nice work @rhydham-nith! Just one minor wording suggestion and a request to make a couple of things headers. Once those are done this should be ready to go

docs/getting_started/comparison.md
rhydham-nith: > just say "same **reliable** model architecture and prediction performance"

Just to confirm—do you mean **same model architecture and prediction performance**, without the word reliable?

ethanwhite: Yes, thanks for checking!

ethanwhite (approved):
Nice work @rhydham-nith! Thanks for the contribution

PR #1045, opened 2025-05-02
Add test and functionality to save the label_dict in the cropmodel

bw4sz:
I found that when you save and reload a CropModel, the label_dict isn't recovered, so you don't know the names of the classes!

- Added a test showing that this fails on main
- Added the correct hook to deepforest.models.CropModel.on_save_checkpoint
- Reloaded on load
- Tests pass.

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -169,7 +169,7 @@ def test_crop_model_load_checkpoint_with_explicit_num_classes(tmpdir, crop_model
     crop_model = model.CropModel(num_classes=num_classes)
     crop_model.create_trainer(fast_dev_run=True)
     crop_model.load_from_disk(train_dir=tmpdir, val_dir=tmpdir)
-
+    crop_model.label_dict = {0: "label1", 1: "label2", 2: "label3"}
     crop_model.trainer.fit(crop_model)
     checkpoint_path = os.path.join(tmpdir, "epoch=0-step=0.ckpt")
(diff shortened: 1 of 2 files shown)


henrykironde (approved):
LGTM 
Forwarding to @jveitchmichaelis to take a look

jveitchmichaelis:
Only thing I'd suggest is explicitly setting a different (not tree?) label dict in the unit test.

tests/test_model.py
```diff
@@ -124,6 +124,25 @@ def test_crop_model_load_checkpoint(tmpdir, crop_model_data):
...
+
+    crop_model.trainer.fit(crop_model)
+    checkpoint_path = os.path.join(tmpdir, "epoch=0-step=0.ckpt")
+    crop_model.trainer.save_checkpoint(checkpoint_path)
+    
+    # Load from checkpoint
+    loaded_model = model.CropModel.load_from_checkpoint(checkpoint_path)
+    
+    # Check that the label dictionary is maintained
+    assert crop_model.label_dict == loaded_model.label_dict
```
jveitchmichaelis, line 143: This could be an additional test case but suggest explicitly setting the dict to ensure it works with different label names (and it's not accidentally `None` for example). Could also just assert the type

bw4sz:
Updated.

bw4sz:
hit the merge when ready.

jveitchmichaelis (approved):

PR #1050, opened 2025-05-06
When reloading crop_model, regenerate the numeric_to_label_dict

bw4sz:
In #705, on reload, we didn't recreate the numeric_to_label_dict as well as the label_dict. Removed the conditional in prediction, there is no reason this shouldn't exist.

### DIFF
--- a/src/deepforest/model.py
+++ b/src/deepforest/model.py
@@ -108,11 +108,16 @@ class CropModel(LightningModule):
                  batch_size=4,
                  num_workers=0,
                  lr=0.0001,
+                 label_dict=None,
                  model=None):
         super().__init__()
         self.num_classes = num_classes
         self.num_workers = num_workers
-        self.numeric_to_label_dict = None
+        self.label_dict = label_dict
+        if label_dict is not None:
+            self.numeric_to_label_dict = {v: k for k, v in label_dict.items()}
+        else:
+            self.numeric_to_label_dict = None
         self.save_hyperparameters()
 
         if num_classes is not None:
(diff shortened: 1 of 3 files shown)


bw4sz:
ready for review.

bw4sz:
I need to rerun tests, I switched label_dict and numeric_to_label, I will squash and confirm.

src/deepforest/model.py
```diff
@@ -108,11 +108,16 @@ def __init__(self,
...
                  model=None):
         super().__init__()
         self.num_classes = num_classes
         self.num_workers = num_workers
-        self.numeric_to_label_dict = None
+        self.label_dict = label_dict
+        if label_dict is not None:
+            self.numeric_to_label_dict = {v: k for k, v in label_dict.items()}
+        else:
+            self.numeric_to_label_dict = None
```
jveitchmichaelis, line 120: What are the consequences of leaving this as `None`? I assume this will get overwritten when a checkpoint is loaded per the PR description.

src/deepforest/dataset.py
jveitchmichaelis, line 1: Should this file be included?

jveitchmichaelis:
LGTM - just a query about `dataset.py` being included here as it seems unrelated.

Otherwise the mechanics around labels being possibly `None` at construction time are fine, but I added a comment to double check when that's updated (eg loading a checkpoint?)

src/deepforest/model.py
bw4sz: a CropModel gets a label dict in one of two ways

1. As an input argument, this is mostly for testing and reloading checkpoints.
2. When you load a dataset for training.


https://github.com/weecology/DeepForest/blob/550e84288f96aaeefe50709cbd4546e8d8b337ea/src/deepforest/model.py#L177

I will add a ValueError if a user goes to predict but has not provided one of these two ways.

bw4sz:
okay, @jveitchmichaelis and @henrykironde I think this ready, responded to concerns above.

jveitchmichaelis (approved):
LGTM

henrykironde (approved):
LGTM

Issue #1051, opened 2025-05-06
Existing hydra config overrides projects that already use hydra.


bw4sz:
we need to provide guidance for projects that use hydra,

when you load a model, you initialize hydra and it looks like it correctly checks it.
https://github.com/weecology/DeepForest/blob/ff6453e98c347b31d23e1de0b1c2b47090196ab3/src/deepforest/utilities.py#L33

But if there is already a loaded hydra instance, you get that config. Example from the BOEM project, 

```
config
[...]

PR #1052, opened 2025-05-06
Simplify config loading

jveitchmichaelis:
Should remove conflicts with Hydra loading described in https://github.com/weecology/DeepForest/issues/1051

Just uses OmegaConf directly for "internal" loading of configuration but still allows overrides.


bw4sz (approved):
Yes, i think this the right choice.

PR #1065, opened 2025-06-09
Fix #1064: Auto log sample train/val images at training start

### LINKED ISSUE #1064: Auto log a few images from train and val dataset to self.log
I find myself always needing to make sure the annotations and images line up as expected, its such a common error. We should just auto log five images before training, I think the cost to the user is very low and it sets up well for tensorboard or comet or wandb.

I can imagine a on_train_start hook in main.py 

something like 

```
  n = 5 if non_empty_train_annotations.shape[0] > 5 else non_empty_train_annotations.shape[0]
  for filename in non_empty_train_annotations.image_path.sample(n=n).unique():
      sample_train_annotations_for_image = non_empty_train_annotations[non_empty_train_annotations.image_path == filename]
      sample_train_annotations_for_image.root_dir = train_image_dir
      visualize.plot_annotations(sample_train_annotations_for_image, savedir=tmpdir)
      comet_logger.experiment.log_image(os.path.join(tmpdir, filename),metadata={"name":filename,"context":'detection_train'})
```

bw4sz:
## Description

This PR implements automatic logging of sample images from training and validation datasets at the start of training, as requested in #1064.

## Changes

- Added `on_train_start` hook in `main.py` that:
  - Samples up to 5 images from training dataset
  - Samples up to 5 images from validation dataset (if available)
  - Uses `visualize.plot_annotations()` to create annotated images
  - Logs images to all available experiment loggers (Comet, TensorBoard, W&B, etc.)

## Implementation Details

- The hook checks for non-empty annotations before sampling
- Images are saved to a temporary directory
- Each image is logged with metadata including filename, context (train/val), and current step
- Compatible with all loggers that have a `log_image` method

## Testing

The implementation will automatically log images when training starts. Users will see sample images in their experiment tracking tool of choice.

Fixes #1064

### DIFF
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -1,6 +1,7 @@
 # entry point for deepforest model
 import importlib
 import os
+import tempfile
 import typing
 import warnings
 
@@ -16,7 +17,8 @@ from torchmetrics.detection import IntersectionOverUnion, MeanAveragePrecision
 from torchmetrics.classification import BinaryAccuracy
 
 from huggingface_hub import PyTorchModelHubMixin
-from deepforest import dataset, visualize, get_data, utilities, predict
+from deepforest import dataset
+from deepforest import visualize, visualize, get_data, utilities, predict
 from deepforest import evaluate as evaluate_iou
 
 from omegaconf import DictConfig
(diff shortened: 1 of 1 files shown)


bw4sz:
This is the very first autogenerated cursor background agent. Not sure yet how to tag it in a way that differentiates it from 'me' as a person. (but then how would you know 'this is me'). Besides the acerbic asides (which I guess cursor would learn to mimic).

PR #1067, opened 2025-06-09
Fix #1064: Auto log sample train/val images at training start (with tests)

### LINKED ISSUE #1064: Auto log a few images from train and val dataset to self.log

bw4sz:
## Description

This PR implements automatic logging of sample images from training and validation datasets at the start of training, as requested in #1064.

### Implementation (`src/deepforest/main.py`)
- Added `on_train_start` hook that:
  - Samples up to 5 images from training dataset
  - Samples up to 5 images from validation dataset (if available)
  - Uses `visualize.plot_annotations()` to create annotated images
  - Logs images to all available experiment loggers (Comet, TensorBoard, W&B, etc.)

### Tests (`tests/test_on_train_start.py`)
- Comprehensive test suite including:
  - Test that training images are logged correctly
  - Test that validation images are logged correctly
  - Test with multiple loggers
  - Test with empty annotations (edge case)
  - Test that correct number of images are sampled
  - Test without any loggers (should not crash)
  - Test that visualize.plot_annotations is called correctly
  - Test with no validation dataset
  - Test that parent class behavior is preserved

## Implementation Details

- The hook checks for non-empty annotations before sampling
- Images are saved to a temporary directory
- Each image is logged with metadata including filename, context (train/val), and current step
- Compatible with all loggers that have a `log_image` method
- Gracefully handles edge cases (empty datasets, no logger, etc.)

## Testing

Run the new tests with:
```bash
pytest tests/test_on_train_start.py
```

The implementation will automatically log images when training starts. Users will see sample images in their experiment tracking tool of choice.

Fixes #1064


jveitchmichaelis:
This is the one from Cursor? It does have some "tells" - for some reason Claude seems to really like mocking in unit tests (and it loves to make long test suites). I had to make a rule to discourage that.

Can the bot respond to PR comments on GitHub? It would be great if we could feedback review here and have it fix stuff.

bw4sz:
Waiting on #1079 to see if there are rebase conflicts.

henrykironde (commented):
Looks good but the tests could use some changes

tests/test_main.py
```diff
@@ -231,6 +231,11 @@ def test_train_single(m_without_release, architecture):
     m_without_release.trainer.fit(m_without_release)
 
 
+def test_on_train_start_basic(m):
+    """Test that on_train_start runs without error and logs images using the default logger."""
+    m.create_trainer(fast_dev_run=False, limit_train_batches=2, limit_val_batches=2)
+    m.on_train_start()
```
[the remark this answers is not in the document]
bw4sz: Added a mock logger to show the image count.

jveitchmichaelis:
Rebased main, there is an unrelated change in models.py which I'll revert back to main and check the tests pass.

src/deepforest/main.py
```diff
@@ -10,14 +11,11 @@
...
 from pytorch_lightning.callbacks import LearningRateMonitor
-from copy import deepcopy
-import pytorch_lightning as L
 from torch import optim
 from torchmetrics.detection import IntersectionOverUnion, MeanAveragePrecision
 from torchmetrics.classification import BinaryAccuracy
 
-from deepforest import utilities, predict
-
+from deepforest import visualize, visualize, utilities, predict
```
henrykironde, line 18: Remove the redundant import

henrykironde (changes requested):
There Is an import that needs to be removed.

Can we include some info about cleaning up the logs.

src/deepforest/main.py
henrykironde, line 18:
```suggestion
from deepforest import visualize, utilities, predict
```

bw4sz:
This PR got stale and rebased and ugly. Closing in favor of the core improvements in #1104

PR #1069, opened 2025-06-09
[WIP] Make active_learning.py

naxatra2:
Adds  #1068
---
this PR creates a new `active_learning.py` file `src/deepforest`. This file will be the main script for the basic pipeline of the active learning module. I have also created a new test file, `test_active_learning.py`, which is currently empty. I plan to add tests to it as needed in the future.


bw4sz:
Good start! 

Can we brainstorm how we imagine users entering this workflow is it like?

Something like ->
```
from deepforest import active_learning 
from deepforest import main
import glob

# start from pretrained model
m = main.deepforest()
m.load_model("Weecology/deepforest-tree")

# Get predictions from folder full of images 
images = glob.glob(folder_path + "*")
predictions = m.predict_tile(path=images)

images_to_annotate = active_learning.run_iteration(predictions)

# doesn't exist yet, maybe a deepforest[label_studio] package
label_studio.upload(images_to_annotate, project)

# Human review and then download
annotations = label_studio.download(project)

# Train new model
trainer.fit(m, annotations)

# repeat
predictions = m.predict_tile(path=images)
images_to_annotate = active_learning.run_iteration(predictions)
```

This is what I mean in starting to get the pipeline worked out to complete the circle.

jveitchmichaelis:
Thanks @naxatra2, I think the a good next step would be to demo a training + random sampling example, without label studio for now. I suggest you use images that we already have labels for (to avoid the human step). Let us know if you need any help putting that together.

bw4sz:
Related to #1100 for downloading data from label studio to complete active learning loop.

naxatra2:
https://www.kaggle.com/code/jiya1404/3-million

@jveitchmichaelis can you see this notebook. I have used the deep forest library and fine tuned it on the milliontrees dataset. The flow is similar to the one I did on the african wildlife dataset. This notebook is very different from the one that we worked in the last meet, because it had many issues that were arising while working on the format of the million trees dataset. I have created two new classes to fix this issue.  

I am obtaining a graph like this. It took a lot of time to run this, and my kernal crashed after ~2000 images, so I have only provided a graph uptil here. I will also add a baseline in this plot 
output (14)

Issue #901, opened 2025-01-27
Eliminate the custom eval code


bw4sz:
Revisiting the pull request #504, torchmetrics wasn't quite ready to be swapped in for all our evaluation, see that thread and the corresponding issue. I think it can be re-evaluated after looking at the repo and answered from

https://github.com/Lightning-AI/torchmetrics/discussions/2152#discussioncomment-11918240

Torchmetrics is standardized, testing and scalable, our code is not. For a matcher example, see the torchvision matcher class used in MillionTrees. 

https://github.com/weecology/MillionTrees/blob/202461412bdd8bc31b632f2636274ed340a631f7/milliontrees/common/metrics/all_metrics.py#L406

# Roadmap

1. Implement IOU matcher and recall/precision score for a fixed threshold
2. Assert that the two versions produce the same result
3. Test timings of two versions

bw4sz:
From the BOEM repo here is a good start

```
    def match_predictions_and_targets(self, pred, target):
        """
        Matches predicted bounding boxes with source bounding boxes using Intersection over Union (IoU).
        Args:
            pred (Tensor): A tensor containing the source bounding boxes.
            target (Tensor): A tensor containing the predicted bounding boxes.
        Returns:
            DataFrame: A dataframe containing the matched predictions and targets.
        """
        
        # Match predictions and targets
        matcher = Matcher(
            0.3,
            0.3,
            allow_low_quality_matches=False)
        
        src_boxes = pred["boxes"]
        pred_boxes = target["boxes"]

        match_quality_matrix = box_iou(
            src_boxes,
            pred_boxes)
        
        results = matcher(match_quality_matrix)

        matched_pred = []
        matched_target = []

        for i, match in enumerate(results):
            if match >= 0:
                matched_pred.append(pred["labels"][i].item())
                matched_target.append(target["labels"][match].item())
            else:
                matched_pred.append(pred["labels"][i].item())
                matched_target.append(None)
                
        matches = pd.DataFrame({"pred": matched_pred, "target": matched_target})

        return matches
            
```

this would replace https://github.com/weecology/DeepForest/blob/c2907318ada4f4a2fc17bb07891e68afd3cc88be/src/deepforest/evaluate.py#L38

martibosch:
Hello @bw4sz,

[...]

PR #1071, opened 2025-06-10
Feat torchmetrics eval

martibosch:
For #901.


martibosch:
this would be a first draft implementing what we discussed in #901.

- As mentioned there, all boxes are matched regardless of their label, which is in line with the current implementation and makes most sense for a bounding box detection + crop model classification approach. Maybe it would make sense to let the user decide whether matches should consider all boxes or only boxes of the same class (e.g., via a boolean argument) in which case we may discard the latest commit [33cdcc3](https://github.com/weecology/DeepForest/pull/1071/commits/33cdcc347da0e13d61ef85ab0050273194d75c1e) and work from there.
- Also as discussed, torchmetrics offers much more options to control the evaluation. Again this PR mimics the current evaluation API, however it may be a good idea to let users define their custom, e.g., recall thresholds, max detection thresholds and the like. This could be done via keyword arguments whose default values mimic the current evaluation, which would allow advanced users to further customize the evaluation without being too overwhelming for most users.

bw4sz:
Agreed, we can add config args (see new hydra config) to give users more control. Make sure to test for edge cases 1) all empty images, and 2) mixed empty and non-empty images. See empty_frame_accuracy. 

https://github.com/weecology/DeepForest/blob/af9458bc300e2a313ecf3834a2510bb31fe35685/src/deepforest/main.py#L97

excited! Something we have wanted for a long time but not been able to crack.

bw4sz:
@martibosch can I help here, looks like a quick rebase, and tell me if you are ready for review.

martibosch:
Hello @bw4sz, sorry for the delay on my side, I have been busy preparing materials for a conference. I will retake this tomorrow. I think I will rebase this so that the CI is run on the pull request and then we can compare results.

martibosch:
Hello! I think I fixed the conflicts, however, is there any reason why the test worfklows are not running on this PR? or they are running and I can't see them?

jveitchmichaelis:
Needs manual approval, but that should be good to run now.

martibosch:
Hello!

I compared the torchmetrics-based evaluation and it seems that there are some differences that we will have to explore. I will come back at you when I find something.

In any case, here is the zip with the data and notebook (you can see the different scores there) that I am working with
[torchmetrics-eval.zip](https://github.com/user-attachments/files/21230827/torchmetrics-eval.zip)

jveitchmichaelis:
One related issue is that `__evaluate_wrapper__` is apparently not idempotent as it modifies the dataframe and you can't run it twice interactively. Not sure if that goes away with this PR. I think we should be doing a copy under the hood there.

I think it's in evaluate_boxes somewhere, we modify `predictions`.

jveitchmichaelis:
Yeah see https://github.com/weecology/DeepForest/issues/901#issuecomment-3157614881. As long as the results are sane and repeatable, it should be OK.

> If we run a large evaluation, like on MillionTrees, is there a time difference?

My guess is probably not for matching. That runs per-image and the forward pass dominates. We could do some testing on like 10M synthetic labels just to see, but for a one-off evaluation I doubt it would be that long.

Running the COCO-style evaluation might take a long time (especially with pycocotools), but `faster-coco-eval` solved a few problems there by porting to C++ so we should definitely use that backend. I think that evaluation is still order of seconds and most of it is forming the reverse index between image ID <> annotations. Though if we have to depend on that library anyway, we could run the eval directly to get access to the matches. It feels odd to do that, but torchmetrics doesn't do anything particularly clever beyond making it easy to update/compute.

jveitchmichaelis:
@martibosch take a look at the changes [here](https://github.com/weecology/DeepForest/pull/1152). I had a look at where we could optimize without changing any of the existing interfaces, and this looks quite promising. The short answer is that most of our existing inefficiency can be removed by vectorising and avoiding lots of repeated dataframe operations.

We have switched to faster_coco_eval which is even faster than my improvements, but I wanted to see where we could get a performance boost without losing the ability to compare to older metrics. I'm also in touch with the maintainers of faster_coco_eval who seem open to feature requests. One of which could be exposing matching results so that you don't need to awkwardly inspect the coco object inside torchmetrics.

The other difference that's apparent in very large tests is that torchmetrics (really cocotools/faster_coco_eval) is single threaded. So if you have a large dataset and a beefy machine, it sits for ages spinning on a single core when most of the per-image processing can be parallelised.

martibosch:
Hello @jveitchmichaelis,

yes the ideal scenario would be to retrieve the match directly form torchmetrics, which I found to be a bit too hacky especially if the user is not bound to a specific backend. However, if the user will be bounded to faster_coco_eval, couldn't we just use the [`faster_coco_eval.mask.io` method](https://github.com/MiXaiLL76/faster_coco_eval/blob/main/faster_coco_eval/core/mask.py#L130) which essentially runs in C++? I do not know if that is faster than `linear_sum_assignment`.

In any case, it makes sense to not use torchmetrics if we do not want many of their features (e.g., maximum detection thresholds), but I am wondering to what extent it would make sense to allow the user to retrieve the whole precision/recall curve.

Finally, out of curiosity, is using the STRtree faster than a spatial join with geopandas? Also a while ago I used the [overlay](https://geopandas.org/en/stable/gallery/overlays.html) functions and I was surprised of how fast they ran over large datasets. However, I suppose that in most of DeepForest use cases, the individual image geo-data frame will most certainly be rather small, in which case most gains would indeed come from exploiting the "embarrassingly parallel" nature of the multi-image evaluation.

Let me know if you prefer to move the review to your PR. Best,
Martí

PR #1072, opened 2025-06-11
add explicit test for reloading label_dict

bw4sz:
Small PR for test coverage, if a user used trainer.save_checkpoint, and not torch.save, the label_dict wasn't being maintained. I added a hook and a test.

PR #1076, opened 2025-06-12
Warn the user that the checkpoint doesn't have a label dict

bw4sz:
before the PR #1072 we need to warn the users of pre2.0 that their checkpoint needs to update the label_dict


jveitchmichaelis:
Just needs a style fix (according to the CI), otherwise looks good

bw4sz:
@jveitchmichaelis pushed, let's merge this so I can rebase #1074 and we can be almost done this round of dev surrounding the cropmodel.

jveitchmichaelis (approved):

PR #1079, opened 2025-06-16
refactor image loading in visualize

jveitchmichaelis:
Some refactoring in `visualisation.py` to make image handling more consistent. Test coverage was a bit spotty, so added checks for the drawing functions as well.

- Renamed "plot" functions to "draw" to better indicate that they return arrays and do not invoke "show"
- Refactored image loading into a separate function that can be used in multiple places. This loader handles the various edge cases (array/str/dataframe, != 3 channels, channel ordering, etc) and returns an RGB image. Supports str, numpy array, PIL Image and dataframe.
- Simplified RGB/BGR concerns and use `cv2.cvtColor` to handle channel conversion (equivalent to rolling axes, but more explicit - and CV2 is usually fast)
- Test cases for drawing points + geometry (could be expanded)
- Test cases for loading in different guises.
- Removed unnecessary fixture calls to `m` in several test cases.
- Fixed point-no-label test which did not have empty labels 
- Aside from the plot->draw rename, user facing interfaces remain the same- 

The diff is a little funky. Suggest viewing the raw file.

### DIFF
--- a/src/deepforest/visualize.py
+++ b/src/deepforest/visualize.py
@@ -25,6 +25,7 @@ def _load_image(image: Optional[Union[np.typing.NDArray, str, Image.Image]] = No
     Args:
         image (optional): Numpy array or string
         df (optiona): Pandas dataframe
+        root_dir (optional): Root directory of the image, will override dataframe root_dir attribute
 
     Returns:
         image: Numpy array
@@ -33,21 +34,28 @@ def _load_image(image: Optional[Union[np.typing.NDArray, str, Image.Image]] = No
     if image is None and df is None:
         raise ValueError(
             "Either an image or a valid dataframe must be provided for plotting.")
-    elif df is not None:
-        if not root_dir and not hasattr(df, 'root_dir'):
-            raise ValueError(
-                "The 'root_dir' attribute does not exist in the dataframe. Please specify the 'root_dir' argument."
-            )
-        else:
+
+    if df is not None:
+        # Resolve image root
+        if hasattr(df, 'root_dir') and root_dir is None:
             root_dir = df.root_dir
             # expected str, bytes or os.PathLike object, not Series
             root_dir = df.root_dir.iloc[0] if isinstance(root_dir,
                                                          pd.Series) else root_dir
+        elif root_dir is None:
+            raise ValueError(
+                "Neither root_dir nor a dataframe with the root_dir attribute was provided."
+            )
 
         image_path = os.path.join(root_dir, df.image_path.unique()[0])
         image = np.array(Image.open(image_path))
     elif isinstance(image, str):
-        image = np.array(Image.open(image))
+        if root_dir is not None:
+            image_path = os.path.join(root_dir, image)
+        else:
+            image_path = image
+
+        image = np.array(Image.open(image_path))
     elif isinstance(image, Image.Image):
         image = np.array(image)
     elif not isinstance(image, np.ndarray):
@@ -78,7 +86,11 @@ def _load_image(image: Optional[Union[np.typing.NDArray, str, Image.Image]] = No
     return image
 
 
-def plot_points(image, points, color=None, radius=5, thickness=1):
+def plot_points(image: np.typing.NDArray,
+                points: np.typing.NDArray,
+                color: Optional[tuple] = None,
+                radius: int = 5,
+                thickness: int = 1) -> np.typing.NDArray:
     """Draw points on an image, returns a copy of the array
     Args:
         image: a numpy array in RGB order, HWC format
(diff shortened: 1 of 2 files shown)


jveitchmichaelis:
This shouldn't clobber, just moves the channel ordering check to the image loader https://github.com/weecology/DeepForest/pull/1067

bw4sz:
Thanks, my only question, is whether the renaming is worth it, breaks any existing scripts using plot_results, which I guess is fine since we are going 2.0, but I never know if its feels like the right choice, I think we had a similar discussion for predict_tile.

src/deepforest/visualize.py
```diff
@@ -10,27 +10,105 @@
...
+        image = np.array(Image.open(image_path))
+    elif isinstance(image, str):
+        image = np.array(Image.open(image))
+    elif isinstance(image, Image.Image):
+        image = np.array(image)
+    elif not isinstance(image, np.ndarray):
+        raise ValueError("Image")
+
+    # Fix channel ordering
+    if image.shape[0] == 3:
```
henrykironde, line 57: Let us hand edge cases where we have (3,3,3) as (C, H, W) and can be (H, W, C). 
```suggestion
    if image.ndim == 3 and image.shape[0] == 3 and image.shape[2] != 3:
```

src/deepforest/visualize.py
```diff
@@ -10,27 +10,105 @@
...
+        warnings.warn("Input images must be channels last format [h, w, 3] not channels "
+                      "first [3, h, w], using np.rollaxis(image, 0, 3) to invert!")
+        image = np.rollaxis(image, 0, 3)
```
henrykironde, line 60:
```suggestion
        image = np.transpose(image, (1, 2, 0))
```
Could this be more clear ? not really important

src/deepforest/visualize.py
```diff
@@ -10,27 +10,105 @@
...
+
+    # Drop alpha channel if present and warn
+    if image.shape[2] == 4:
```
henrykironde, line 63:
```suggestion
    if image.ndim == 3 and image.shape[2] == 4:
```
We can add this check` if image.ndim == 3` before accessing `shape[2]`

src/deepforest/visualize.py
```diff
@@ -10,27 +10,105 @@
...
+        warnings.warn("Image has an alpha channel. Dropping alpha channel.")
+        image = image[:, :, :3]
+
+    if image.dtype != np.uint8:
+
+        warnings.warn(f"Image is {image.dtype}. Will be cast to 8-bit unsigned")
+
+        # Images in [0,1] are allowed, but should be rescaled
+        if image.max() < 1:
```
henrykironde, line 72: Let us add a check on the lower bound because the lower bound could allow negative numbers
```suggestion
        if image.max() <= 1.0 and image.min() >= 0:
```

src/deepforest/visualize.py
```diff
@@ -10,27 +10,105 @@
...
+            warnings.warn(f"Assuming image is in [0,1, scalingg to [0,255]")
```
henrykironde, line 73:
```suggestion
            warnings.warn(f"Assuming image is in [0,1, scaling to [0,255]")
```

src/deepforest/visualize.py
```diff
@@ -10,27 +10,105 @@
...
+            image *= 255
+
+        image = image.astype("uint8")
```
henrykironde, line 76:
```suggestion
        image = np.clip(image, 0, 255).astype('uint8')
```
Let  us clip incase of over/underflows

henrykironde (changes requested):
@jveitchmichaelis, I have made a few suggestions, feel free to pick out those that should be handled.

jveitchmichaelis:
OK @henrykironde I've made those changes.

In the future we can probably standardize on either our own low level drawing functions or `supervision`. `supervision.utils.draw_x` provides all the array drawing functions, or we could drop the dependency entirely. If you look at the [source](https://supervision.roboflow.com/annotators/#boundingboxannotator) `supervision` is a very thin wrapper around OpenCV anyway.

On the one hand, it's helpful that they've packaged up those functions, on the other we have duplicated code for plot_points and plot_predictions and we have to run detections through an additional conversion to their internal format.

bw4sz:
> OK @henrykironde I've made those changes. Couple of other bugfixes - the deprecated functions I added didn't return anything, and we should swap the image channels back to RGB after using OpenCV.
> 
> In the future we should probably standardize on either our own low level drawing functions or embrace `supervision` entirely. `_plot_image_with_geometry` handles the same logic as `plot_predictions` using `supervisions`'s `Annotator` classes. Underneath it's the same OpenCV that we use, so we could equivalently drop the dependency and retain `plot_predictions` instead.
> 
> It's helpful that they've packaged up those functions and it's actively maintained. Since they also provide support for label/text annotation (e.g. confidence score + box) I think that might be useful for animal detection later where objects will be more sparse and perhaps we want to see scores. @bw4sz thoughts?
> 
> I would expose a single user-facing function `draw_predictions(image, df, color, thickness, labels)` that wraps `_plot_image_with_geometry` (on top of the utils that invoke matplotlib) + one for `draw_points`(using `DotAnnotator`) if we need it.

sure, happy to overlay text.

PR #1083, opened 2025-06-25
WIP: Refactor weight loading from deepforest:main to models

jveitchmichaelis:
This has come out of https://github.com/weecology/DeepForest/pull/1078 and some musings about how to handle checkpoints when we add new model variants to the package.

We currently store weights for trained models on HuggingFace. Models are instantiated two ways:

- `deepforest.main` calls `create_model()` on construction. This returns an instance of the desired architecture that's initialised to whatever the object returns by default, assuming no model was passed to the constructor directly.
- Alternatively/additionally we then call `load_model()` to pull in pre-trained weights.

There are a couple of issues here that I've run into when trying to add another model to the package:

- `load_model()` calls `deepforest.main.from_pretrained` which is a class method provided from the `PyTorchModelHubMixin` mixin. We attempt to set the return value to `self.model` and overwrite a few settings such as the instance label_dict, model and numeric_to_label_dict. This is an unusual way of using this function, because it builds an entirely new deepforest instance and then copies those three items from it. Whereas you'd normally do something like this to begin with:

```python
m = deepforest.from_pretrained(model_name, revision)
```

- We don't need to call `create_model` on construction unless we're training, because `load_model` will override it. This doesn't save much time, but it's a bit more efficient.
- Some parameters are hardcoded (e.g. the bird model class names) in the library, when they would be better off in the hub checkpoint.

Given what `load_model` is doing, I suggest we move the weight loading logic to the models directly.

This has an advantage that the weights stored on HF are (mostly) decoupled from the library and minimally contain the weights + label dict. I suspect there is currently a bug in the constructor at the moment where all models will open as retinanets, because of the way the config system is loaded (and the unnecessary extra call to `create_model` when you call `load_pretrained` as it calls the constructor again). Investigating that.

This PR proposes that models subclass `PyTorchModelHubMixin` where necessary. This has no effect on Lightning checkpointing as far as I'm aware, as those are separate (and store a lot more information to resume training state). This does not need to be done for transformers models as they already support the hub methods, but it would require adding the mixin to RetinaNet.

We would need re-publish weights to HF with a suitable revision (e.g. v2) and deprecate the old ones as v1. As long as the repository is updated in step with this change, this should be transparent. Otherwise we can keep a pathway to load the older weights but I've not found a clean way to get rid of the call to `create_model` this way.

```python
config = utilities.load_config(overrides= {"model": {"name": "joshvm/deepforest-tree",
                                                     "revision": "v2"}})
m = deepforest(config=config)
m.load_model()
results = m.predict_image(path="../src/deepforest/data/2018_SJER_3_252000_4107000_image_477.tif")
assert len(results) > 0
plot_results(results)
```

Config is overridden for now with test weights on HF, but we should revert that and make sure weights are under the weecology namespace.

### DIFF
--- a/src/deepforest/conf/config.yaml
+++ b/src/deepforest/conf/config.yaml
@@ -13,8 +13,11 @@ num_classes: 1
 nms_thresh: 0.05
 
 model:
-    name: 'weecology/deepforest-tree'
-    revision: 'main'
+    name: 'joshvm/deepforest-tree'
+    revision: 'v2'
+
+label_dict:
+    Tree: 0
 
 # Pre-processing parameters
 path_to_raster:
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -86,8 +86,6 @@ class deepforest(pl.LightningModule, PyTorchModelHubMixin):
         self.existing_train_dataloader = existing_train_dataloader
         self.existing_val_dataloader = existing_val_dataloader
 
-        self.create_model()
-
         # Metrics
         self.iou_metric = IntersectionOverUnion(
             class_metrics=True, iou_threshold=self.config.validation.iou_threshold)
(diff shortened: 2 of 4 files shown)


jveitchmichaelis:
@ethanwhite feel free to take a look. The gist of the PR remains the same, the save/load difference _should_ be a simple fix once we figure out what state is/isn't being reloaded (but the model is clearly working so it's not like we're loading garbage).

henrykironde:
Had a look, and it’s looking good for the moment. If we are downgrading to omega conf do we still need hydra-core or we use both ?

ethanwhite (commented):
Nice work on this @jveitchmichaelis. Everything looks good to me. I've made a couple of minor suggestions and asked a couple of questions on things I wasn't sure about.

From the tests it looks like there's probably a small bug in the changes to the logging code. If you want to pull that into a separate PR like you mentioned I have no objection, but I'm also fine with cleaning it up here. The different number of predictions is definitely super weird and I agree that we want to understand it before merging. I agree that there's nothing obvious here that should result in this change.

jveitchmichaelis:
Thanks @henrykironde, I think there is an underscore method we can use for older versions. The newer one seems to be a bugfix? https://github.com/pytorch/pytorch/blob/v2.7.0/torch/nn/modules/module.py#L2097

The problem does occur within the same version of Pytorch though, for example if you load the weights, predict and then save/reload? But yeah I'm not sure how important this is versus checking that the behaviour is generally similar.

jveitchmichaelis:
I think this is good for review. I fixed some of the label_dict bugs and added some additional test cases to verify things like single label -> multi label models and vice versa.

Also updated the docs as some sections were outdated even before this PR.

henrykironde (approved):
looks good to me

PR #1085, opened 2025-07-01
Use structured configs

jveitchmichaelis:
This PR improves config handling to use omegaconf's [structured config](https://omegaconf.readthedocs.io/en/2.1_branch/structured_config.html) for more strict type checking.

`schema.py` is a logical place to enforce other checks like which values certain options can take. This can be done as a `__post_init__` step but it gets a bit unwieldy for longer sections. We could also adopt `pydantic` https://docs.pydantic.dev/latest/.


jveitchmichaelis:
Will add some unit tests for some of the edge cases and this is good for review.

henrykironde (approved):
Looks good to me at this state. Adding some tests would be good.

jveitchmichaelis:
Rebased and will have a look at some tests, but I dunno how exhaustive we really need to be (maybe a few tests to check incorrect typing).

bw4sz (approved):
Looks good, I don't think it will affect user behavior at all.

Issue #574, opened 2023-12-12
Better error checking on label dict for multi-class


bw4sz:
When we are dealing with multi-class models, we late pretty late until a non-matching label error is thrown.

```
Exception has occurred: KeyError
Caught KeyError in DataLoader worker process 0.
Original Traceback (most recent call last):
  File "/orange/ewhite/b.weinstein/miniconda3/envs/EvergladesTools/lib/python3.11/site-packages/torch/utils/data/_utils/worker.py", line 308, in _worker_loop
    data = fetcher.fetch(index)
           ^^^^^^^^^^^^^^^^^^^^
  File "/orange/ewhite/b.weinstein/miniconda3/envs/EvergladesTools/lib/python3.11/site-packages/torch/utils/data/_utils/fetch.py", line 51, in fetch
    data = [self.dataset[idx] for idx in possibly_batched_index]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/orange/ewhite/b.weinstein/miniconda3/envs/EvergladesTools/lib/python3.11/site-packages/torch/utils/data/_utils/fetch.py", line 51, in <listcomp>
    data = [self.dataset[idx] for idx in possibly_batched_index]
            ~~~~~~~~~~~~^^^^^
  File "/orange/ewhite/b.weinstein/miniconda3/envs/EvergladesTools/lib/python3.11/site-packages/deepforest/dataset.py", line 109, in __getitem__
    targets["labels"] = image_annotations.label.apply(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/orange/ewhite/b.weinstein/miniconda3/envs/EvergladesTools/lib/python3.11/site-packages/pandas/core/series.py", line 4760, in apply
    ).apply()
      ^^^^^^^
  File "/orange/ewhite/b.weinstein/miniconda3/envs/EvergladesTools/lib/python3.11/site-packages/pandas/core/apply.py", line 1207, in apply
    return self.apply_standard()
           ^^^^^^^^^^^^^^^^^^^^^
  File "/orange/ewhite/b.weinstein/miniconda3/envs/EvergladesTools/lib/python3.11/site-packages/pandas/core/apply.py", line 1287, in apply_standard
    mapped = obj._map_values(
             ^^^^^^^^^^^^^^^^
[...]

PR #1093, opened 2025-07-15
raise error in create_trainer when there's a label mismatch

dylankershaw:
This PR addresses https://github.com/weecology/DeepForest/issues/574.


jveitchmichaelis:
The test that breaks is `test_checkpoint_label_dict` because of the ordering issue between setting the dict and creating the trainer. I feel we should move towards enforcing this at a `config` level (i.e. this should be set when you create a `deepforest` object and one should be discouraged from modifying it). So we'd prefer:

```
m = main.deepforest(config_args={"num_classes": 1},
                        label_dict={
                            "Object": 0
                        })
```

rather than overwriting the label dict after creation. The other PR would also encourage `set_labels()` which handles the reverse index as well.


```
m.create_trainer() <- test fails here
m.label_dict = {"Object": 0}
m.numeric_to_label_dict = {0: "Object"}
m.trainer.fit(m)
m.trainer.save_checkpoint("{}/checkpoint.pl".format(tmpdir))
```

would be replaced with:

```
m.create_trainer() 
m.trainer.fit(m) <- test would fail here, when Lightning calls m.train_dataloader
m.trainer.save_checkpoint("{}/checkpoint.pl".format(tmpdir))
```

jveitchmichaelis:
Yes, no problem to scoping this only to label validation. There is another open PR for box validation but the author hasn't replied in a while. If you wanted to take over that, we should be able to merge it with co-authorship.

I've requested a minor change to make this a dataset method. This also avoids calling `read_csv` twice since you can check `self.annotations`.

bw4sz:
@dylankershaw thanks for your help here.

dylankershaw:
> I've requested a minor change to make this a dataset method. This also avoids calling read_csv twice since you can check self.annotations.

Done! ✅ 

I rebased main as well so this should be good to go @jveitchmichaelis

jveitchmichaelis:
Yeah bit odd - @dylankershaw can you see if you can push without the utilities test file included? Otherwise looks good to me.

jveitchmichaelis (approved):
LGTM :)

PR #1097, opened 2025-07-29
Update dependencies for compatibility with tokenizers 

henrykironde:
Adjusting dependencies due to platform incompatibility

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ dependencies = [
     "aiolimiter",
     "aiohttp",
     "docformatter",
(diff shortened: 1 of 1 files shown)


henrykironde:
This doesn't actually resolve the hiccup we encountered. it's only working now because the tokenizer released a new version yesterday

henrykironde:
We do not have direct dependency on tokenizer, but I can include it. huggingface_hub and transformers dictate the version used for tokenizers.

ethanwhite:
Got it. In that case I think we can just close this if everything is working now.

PR #1098, opened 2025-07-29
added Double Counting algorithm for review

Bhavya1604:
Hi @henrykironde , @bw4sz and @jveitchmichaelis  
This Draft PR contains a working prototype script that implements the "predict and delete" strategy to handle the double-counting of objects in overlapping images. I've adapted the core logic from the DoubleCounting repo and integrated it

I tested the workflow on a dataset with 70-80% overlap using the "left-hand" strategy for clear visualization. The blue boxes are all initial predictions, while the pink boxes are the final, unique predictions for that image.

output : 
image

we can observe that the top predictions are in pink which are unique (new for that image) which indicates that the code is able to identify the overlap and is working fine. there were a total of 401 predictions of which 194 where detected to be unique.

This PR contains the standalone DoubleCounting.py script for review. Before I start integrating this into the main DeepForest library, I would greatly appreciate your feedback

### DIFF
--- a/DoubleCounting.py
+++ b/DoubleCounting.py
@@ -1,0 +1,279 @@
+import os
+os.environ['KMP_DUPLICATE_LIB_OK']='TRUE'
+
+import cv2
+import h5py
+import numpy as np
+import pandas as pd
+import torch
+import torchvision
+import pycolmap
+from pathlib import Path
+from matplotlib import pyplot
+import geopandas as gpd
+from shapely.geometry import box
+
+from deepforest import main
+from hloc import extract_features, match_features, pairs_from_exhaustive
+from hloc.utils.io import get_matches
+
+
+def get_matching_points(h5_file, image1_name, image2_name, min_score=None):
+    """Get matching points between two images from an h5 file."""
+    matches, scores = get_matches(h5_file, image1_name, image2_name)
+    if min_score is not None:
+        matches = matches[scores > min_score]
+    match_index = pd.DataFrame(matches, columns=["image1", "image2"])
+    
+    features_path = os.path.join(os.path.dirname(h5_file), "features.h5")
+    with h5py.File(features_path, 'r') as features_h5_f:
+        keypoints_image1 = pd.DataFrame(features_h5_f[image1_name]["keypoints"][:], columns=["x", "y"])
+        keypoints_image2 = pd.DataFrame(features_h5_f[image2_name]["keypoints"][:], columns=["x", "y"])
+        points1 = keypoints_image1.iloc[match_index["image1"].values].values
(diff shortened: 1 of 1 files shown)


bw4sz:
Thanks for your thoughts here. Your example is somewhat difficult to follow because there are alot of boxes, can you use the deepforest bird model and the data here https://github.com/weecology/DoubleCounting/tree/main/tests/data/birds. 

I _think_ that if we merge this we would need a seperate pip install since the dependencies are heavy compared to the rest of the repo. So I imagine something like

pip install deepforest[double_counting] I think this would be an extra the .toml 

```
    # pyproject.toml
    [project]
    name = "my_package"
    version = "0.1.0"

    [project.optional-dependencies]
    subpackage_extra = [
        "dependency_for_subpackage_extra_1",
        "dependency_for_subpackage_extra_2",
    ]
```
Then we would need to collect several other datasets to try to get a handle on how well it generalizes and what parameters are sensitive. These parameters would need to go in the hydra config. The general workflow would be something like

1. Make a function that takes in a list of images. 
2. use predict_tile on each image
3. Run double counting
4. Produce visualizations
5. Return a results object of unique data

All of this is the module you attached, but would need a integration into deepforest.main()

plus a documentation page with examples.

# Roadmap

[ ] example with existing test data
[ ] gather new examples to assess parameter sensitivity
[ ] package extra install for unique dependencies
[ ] documentation page
[ ] integrate a function with deepforest main (unique_predictions_images in your module, maybe call it deepforest.main.predict_unique

Bhavya1604:
@bw4sz Thank you again for the detailed feedback and the clear roadmap. The plan for optional dependencies and integrating the feature as predict_unique makes perfect sense.

As you suggested, I've rerun the workflow on the bird dataset to provide a clearer example of the result.
image

I'll start testing it with different data and let you know the sensitive parameters.

Bhavya1604:
@bw4sz i could only fine few overlapping dataset from kaggle and github. the rest all data i found was not in order or there where arial images but they where of different places without overlaps so where can i find more datasets to test this . like are there any keywords which would help me or any place i could search for these type of datasets.

bw4sz:
We have a number of datasets, let me look into this today. Can you look at that roadmap above and summarize which pieces are completed, which you plan to do, and which I can help you with. This is great stuff and I have some time this week to assist in review and get it over the finish line. Thanks for the contribution!

Bhavya1604:
@bw4sz i hope you are doing well 
I wanted to know is you have reviewed this latest commit.

ethanwhite:
Closing for now since we haven't heard back from the original contributor in a while.

If folks are interested in continuing to work on this now please feel free to reopen or open a similar PR that pulls everything into its own module.

Issue #1099, opened 2025-08-08
safetensors AssertionError when loading default model in dev version


martibosch:
**Describe the bug**

I managed to run the DETR model by installing DeepForest from github but then I am no longer able to run the prebuilt retinanet model:

```
Traceback (most recent call last):
  File "<python-input-1>", line 1, in <module>
    model = main.deepforest()
  File "/home/martibosch/miniforge3/envs/deepforest/lib/python3.13/site-packages/deepforest/main.py", line 107, in __init__
    self.create_model()
    ~~~~~~~~~~~~~~~~~^^
  File "/home/martibosch/miniforge3/envs/deepforest/lib/python3.13/site-packages/deepforest/main.py", line 238, in create_model
    self.load_model()
    ~~~~~~~~~~~~~~~^^
  File "/home/martibosch/miniforge3/envs/deepforest/lib/python3.13/site-packages/deepforest/main.py", line 144, in load_model
    self.model = model_class.Model(config=self.config).create_model(
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        pretrained=model_name, revision=revision)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/martibosch/miniforge3/envs/deepforest/lib/python3.13/site-packages/deepforest/models/retinanet.py", line 175, in create_model
    model = RetinaNetHub.from_pretrained(pretrained,
                                         revision=revision,
    ...<3 lines>...
                                         score_thresh=self.config.score_thresh,
                                         **hf_args)
  File "/home/martibosch/miniforge3/envs/deepforest/lib/python3.13/site-packages/deepforest/models/retinanet.py", line 57, in from_pretrained
    model = super().from_pretrained(pretrained_model_name_or_path, **kwargs)
  File "/home/martibosch/miniforge3/envs/deepforest/lib/python3.13/site-packages/huggingface_hub/utils/_validators.py", line 114, in _inner_fn
    return fn(*args, **kwargs)
  File "/home/martibosch/miniforge3/envs/deepforest/lib/python3.13/site-packages/huggingface_hub/hub_mixin.py", line 566, in from_pretrained
    instance = cls._from_pretrained(
        model_id=str(model_id),
    ...<7 lines>...
        **model_kwargs,
    )
[...]

PR #1106, opened 2025-08-21
set default weights to v2

jveitchmichaelis:
Quick fix for upstream `safetensors` issue. This sets the default weights to `v2`  which have tensor names that align with the refactored model classes. Older installs of DeepForest will still use `main` so this should not affect existing installs for back-compatibility. Once safetensors is fixed, `main` weights will also work because the torch load hook in the RetinaNet class will fix them.

Should fix https://github.com/weecology/DeepForest/issues/1099


jveitchmichaelis:
@henrykironde @bw4sz @ethanwhite this should be a fix for the safetensors issue, but it might be safer to pin `safetensors` to a slightly older release while we double check behavior

PR #1110, opened 2025-08-26
Refactor style

henrykironde:


henrykironde:
Yes I am working on them. I didn't expect a review until the tests pass since I added more commits

ethanwhite:
> Yes I am working on them. I didn't expect a review until the tests pass since I added more commits

:+1: I just happened to have a few minutes so I glanced at it.

henrykironde:
I will add it after this. It will still fail since there is one file expected to fail, preprocess.py.

PR #1111, opened 2025-08-28
standardize on pre-commit + Ruff, use nbQA for notebooks, and docformatter

henrykironde:
- Replace YAPF with Ruff (ruff-format + ruff --fix) for Python sources.
- Add nbQA hooks to format and lint notebooks (ruff-format + ruff).
- Run the same checks in CI via pre-commit/action@v3.0.1.


ethanwhite (approved):
LGTM!

PR #1121, opened 2025-09-02
Move from conda to UV

henrykironde:


henrykironde:
Here we are pinning the major core deps

jveitchmichaelis (approved):

PR #1122, opened 2025-09-02
refactor: modernize type annotations using PEP 585 and 604

henrykironde:
Use dict[] instead of Dict[] and | instead of Optional[]


jveitchmichaelis (changes requested):
~~If we're removing `environment` etc we should also double check docs and make sure other things like minimum python versions are respected/updated. The install.md guide in particular is structured around a conda installation, so that needs to be updated.~~

I'm guessing the deletion was unintentional? It might make sense to move install specific things to a new PR and keep this focused on typing.

henrykironde:
With the transition from conda to UV (handled in a separate PR), I’ll update the docs last. I have several ruff-related changes that I’d like to merge in small chunks. The files are ready, but I don’t want to drop everything on you at once.

Issue #1109, opened 2025-08-25
Path/permission error when logging to Comet


jveitchmichaelis:
**Describe the bug**

Seems to be some out-of-scope issue with image logging, like the file is deleted before the logger can get to it but not obvious why `tmpdir` is going out of scope. Also not clear from the error if it's actually a non-existent path, or if Comet just doesn't have permissions.

```
Training: |          | 0/? [00:00<?, ?it/s]COMET ERROR: We failed to read file '/scratch/local/10739246/tmpmeys_s15/train_2018_SJER_3_259000_4110000_image_466.png' for uploading.
Please double-check the file path and permissions
COMET ERROR: We failed to read file '/scratch/local/10739246/tmpmeys_s15/train_2019_DSNY_5_452000_3113000_image_crop_14.png' for uploading.
Please double-check the file path and permissions
COMET ERROR: We failed to read file '/scratch/local/10739246/tmpmeys_s15/train_2019_DSNY_5_452000_3113000_image_crop_193.png' for uploading.
Please double-check the file path and permissions
COMET ERROR: We failed to read file '/scratch/local/10739246/tmpmeys_s15/train_2018_SJER_3_258000_4106000_image_145.png' for uploading.
```

**To Reproduce**

Run training via CLI (e.g. `deepforest --config-name=x train`) with a Comet logger, see Dino branch.

**Environment (please complete the following information):**

Env installed via `uv` on HiperGator.

PR #1128, opened 2025-09-08
Find 2.0 deprecations

bw4sz:
# Roadmap
- [x] Remove deprecation warnings and eliminate code
- [x] Create a 2.0 migration doc
- [x] Search through docs to make sure we don't use deprecated code
- [x] Self-review on github

PR #1132, opened 2025-09-12
move image logging to callback

jveitchmichaelis:
This PR improves image visualization during training. It also includes some minor logic changes to `main.py` and `visualize.py` to support.

- Pre-train sample plotting has been moved to the same callback that plots sample images during training. This is IMO a logical place for this code to go, but does mean that the callback should be set up separately and won't be added by default.
- Images can be logged pre-train (dataset samples) and during training after the validation loop (predictions with ground truth).

Changes to `callbacks.py`
- Images are logged in `save_dir` with subdirectories for specific image categories (e.g. train_sample, val_sample and predictions). You can pass a tmpdir if you don't want anything to persist on disk, however the intent is that users will set up training to log to a specific and well-organized folder that replicates their experiment (see dinov3 CLI for example).
- Image logging during training now includes ground truth as well as the predictions. Metadata is also logged + saved to disk. This can definitely be expanded in future (e.g. plot scores/labels).
- Supports Tensorboard and/or Comet. Images are tagged so that Tensorboard will group them appropriately. Comet doesn't do this very well, but it's not a disaster as they're still sorted by step (+ group in the web UI). If both loggers are present, Tensorboard is favored so that Comet doesn't double-log. Tensorboard is an explicit dependency of DeepForest currently, so we can assume it's installed.
- One idea I had was to have the callback specifically take a logger in its constructor so that it will _only_ send to a specific one. We set up loggers + callbacks at the same time anyway. However, it wouldn't avoid double-logging unless Comet is set to not auto-detect metrics (I don't know what else this disables?). This could also be used to take advantage of things like the [Comet annotation API](https://www.comet.com/docs/v2/guides/experiment-management/log-data/images/#log-image-with-bounding-boxes). Could be a demo script so we don't need to add the dependency.

Changes to `visualization.py`:
- To plot the image, we call `plot_annotations`/`plot_results` which save the image to disk.
- Since we're plotting dataset samples, images are assumed to be transformed. It's important that we plot the transformed image.
- If you pass both an image and a dataframe to `_load_image`, the dataframe would take precedence. This would cause weird behaviour where we would take annotations from the dataset and the non-transformed image from disk.
- Refactored so that if you pass an image, it takes precedence. The plotting functions pass both to the loader, as they don't check which of image/df was provided.
- Use PIL for image saving outside of places where we're using OpenCV to avoid extra conversion between color formats.

Changes to `main.py`:
- ~~Cache `targets` as well as predictions on validation rounds. I suspect this is a bad idea.~~ I'm going to index the dataset directly instead, less ram (forgot it's a dataframe).
- I think the callback here actually is off-by-one. It should compute evaluation on `self.current_epochs + 1`. Will submit that separately.

Tests:
- Re-jigged the existing mock test to confirm that logging is called.
- Added other tests to confirm files are saved to disk.
- Added an assertion to the Checkpoint callback to confirm it fires

I've also deprecated `images_callback` in favour of `ImageCallback` and moved `savedir` to `save_dir` to align better with Lightning logger patterns (though ModelCheckpoint uses `dirpath`, go figure).

image

image

image

Fixes https://github.com/weecology/DeepForest/issues/1109

### DIFF
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -1,7 +1,6 @@
 # entry point for deepforest model
 import importlib
 import os
-import tempfile
 import warnings
 
 import geopandas as gpd
@@ -18,7 +17,7 @@ from torchmetrics.classification import BinaryAccuracy
 from torchmetrics.detection import IntersectionOverUnion, MeanAveragePrecision
 
 from deepforest import evaluate as evaluate_iou
-from deepforest import predict, utilities, visualize
+from deepforest import predict, utilities
 from deepforest.datasets import prediction, training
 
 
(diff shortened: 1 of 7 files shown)


jveitchmichaelis:
Ready for review. I've seen some sporadic test failures for some Py versions that seem unrelated to this PR (like a specific multi-batch predict test).

src/deepforest/main.py
```diff
@@ -959,15 +872,15 @@ def on_validation_epoch_end(self):
 
         if self.current_epoch % self.config.validation.val_accuracy_interval == 0:
             if len(self.predictions) > 0:
-                self.predictions = pd.concat(self.predictions)
+                predictions = pd.concat(self.predictions)
```
jveitchmichaelis, line 875: Callback order isn't guaranteed, best practice is to not mutate the state of the module (i.e. we can't rely on this callback concatenating the predictions).

jveitchmichaelis:
Would be good to add this to https://github.com/weecology/DeepForest/pull/1133 so we can remove the class alias for 2.0.

src/deepforest/main.py
bw4sz: that seems fair

src/deepforest/visualize.py
```diff
@@ -187,8 +189,8 @@ def plot_predictions(
 
 
 def draw_predictions(
```
bw4sz, line 191: These functions will all be deprecated in #1128

bw4sz (approved):
This feels good to me, it doesn't appear there is any interaction here with the deprecated viz functions in #1128

jveitchmichaelis:
Made the code example in the docs clearer, removed (now stale) changes to visualization.py. They weren't being used anyway.

PR #1140, opened 2025-09-16
Allow old configs to be used in new 2.0 checkpoints

bw4sz:


jveitchmichaelis (changes requested):
Looks fine, just need to clean the formatting (diff seems quite large?)

bw4sz:
closes https://github.com/weecology/DeepForest/issues/1142

jveitchmichaelis (changes requested):
Looks good, can you double check pre-commit on the first half of the documentation? It looks like trailing whitespace that's already in the file is being removed (which is fine), but not sure if that's intentional there or not.

ethanwhite (approved):
Those white space changes look correct to me, not sure why they were still there (maybe ruff is set to not complain about trialing line space?), but I definitely think this is good to merge.

ethanwhite:
@jveitchmichaelis - I'd say go ahead and merge

jveitchmichaelis (approved):

Issue #1135, opened 2025-09-14
OOM running evaluate on large datasets.


jveitchmichaelis:
**Describe the bug**
Running experiments for large scale pretraining, `evaluate` triggers an out of memory. Not sure if this is a leak, but it seems to be system RAM not VRAM. If only `torchmetrics` is used, training completes.

```bash
 before the concat operation.
FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
slurmstepd: error: Detected 1 oom_kill event in StepId=12725053.0. Some of the step tasks have been OOM Killed.
srun: error: c1000a-s5: task 0: Out Of Memory
```

**To Reproduce**
Run:

```
uv run deepforest train
```

With a large validation dataset, e.g. ~20k images with O(100) annotations per image.

**Environment (please complete the following information):**

HiperGator

```bash
#!/bin/bash
[...]

PR #1141, opened 2025-09-16
log torchmetrics per val loop

jveitchmichaelis:
Moved `torchmetrics` metric log command out of conditional statement, this is accumulated every validation step so no need to wait until we compute `evaluate` to log them.

### DIFF
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -957,6 +957,9 @@ class deepforest(pl.LightningModule):
         if self.trainer.sanity_checking:  # optional skip
             return
 
+        # Log epoch metrics
+        self.log_epoch_metrics()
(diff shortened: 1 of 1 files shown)


bw4sz:
I'm fine with this, should we rename the config arg 

https://github.com/weecology/DeepForest/blob/589f5fbfd1e06c4cebbc20e32e295ed69d81ab42/src/deepforest/conf/config.yaml#L88

this relates to the linger uncomfortableness with torchmetrics + custom eval code, we never got that settled in #504

jveitchmichaelis:
Should fix https://github.com/weecology/DeepForest/issues/1135

bw4sz (approved):

PR #1144, opened 2025-09-18
use faster coco eval for mAP backend

jveitchmichaelis:
This PR changes the backend for torchmetrics mAP calculation to a faster one, adds the `faster_coco_eval` dependency and drops `pycocotools`.

I used `uv add` and `uv remove`. Since the uv diff doesn't show any additional dependencies on `pycocotools`,  if tests pass I think it's safe to remove.

### More context

By default `torchmetrics` uses the `pycocotools` [backend](https://github.com/ppwwyyxx/cocoapi) for mAP calculation. This is a legacy library from FAIR which is no longer actively maintained^ and is quite slow. For a large pretraining dataset with 20k images, this trace shows CPU usage for around 40-50 mins of training, 50-10 mins of validation and 20-30 mins of metric compute.

image

This PR switches to `faster_coco_eval` which is an [alternative API-compatible backend](https://github.com/MiXaiLL76/faster_coco_eval) supported by `torchmetrics` which is actively maintained and is almost an order of magnitude faster in testing. [Torchmetrics says their testing found it to be identical](https://lightning.ai/docs/torchmetrics/stable/detection/mean_average_precision.html) (see the "important" box), and we don't use it as our gold standard metric anyway. The improved trace shows an improvement to only around 5 mins for mAP computation (a 5-6x speedup which is about what the authors claim in their own benchmarks).

image

^ I say "actively" because it looks like there have been several maintenance PRs merged recently, but none that address the core performance issues. This newer repo a fork of the [old cocoapi repo](https://github.com/cocodataset/cocoapi) which is definitely unmaintained. It's owned/maintained by the Detectron2 folks.


bw4sz:
The value remains the same?

bw4sz (approved):
Its amazing that the stack is so diverse and changing all the time.

bw4sz:
That's good enough for me, merging.

PR #1145, opened 2025-09-19
unpin torch and torchvision

bw4sz:


bw4sz:
I amended to just have minimums

jveitchmichaelis:
I'm not too worried about the bot comment. `uv` should prevent installing incompatible torch/torchvision. Otherwise uncapping it ought to be fine.

If we could support the sm_100 versions (2.8?) that'd be convenient but not urgent, uv should also cope with the other dependencies.

jveitchmichaelis (approved):
Looks good to me, we can revisit doc guidance for running on sm_100+

PR #1147, opened 2025-09-20
Ensure train and val dataset have the same index mapping for CropModel

bw4sz:
This PR is in response to the BOEM pipeline in which I see high Accuracy, but low per-class accuracy. I believe that there is an edge case that if the val and train dataset don't have the same length, there is an off-by-one error.

This PR ensures that the CropModel index order is the same regardless of whether train or validation datasets have the exact same classes.

Create a test that asserts class mappings regardless if one is a set of the other.

### DIFF
--- a/tests/test_crop_model.py
+++ b/tests/test_crop_model.py
@@ -309,7 +309,7 @@ def test_val_dataset_confusion_simple_three_class():
 
 
 class _TinyClsDataset(torch.utils.data.Dataset):
-    def __init__(self, labels, num_per_class=1):
+    def __init__(self, labels):
         self.samples = []
         for lbl in labels:
             # one dummy image per label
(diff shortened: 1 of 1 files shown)


jveitchmichaelis (changes requested):
Not clear to me from the test cases if this includes a check for the new aligned class dataset where we have e.g. (A: 0, B: 1) vs (B: 0, A: 1). This should be a valid train/val now? Since the class mapping is provided, we should also test some pathological case where a user provides a completely different `class_to_idx` like (A: 10, B: 15) and the datasets have different mappings.

bw4sz:
> Not clear to me from the test cases if this includes a check for the new aligned class dataset where we have e.g. (A: 0, B: 1) vs (B: 0, A: 1). This should be a valid train/val now? Since the class mapping is provided, we should also test some pathological case where a user provides a completely different `class_to_idx` like (A: 10, B: 15) and the datasets have different mappings.

The user doesn't set the class mappings for the cropmodel, those are set by the torchvision ImageFolder, what this PR does is make sure that the train and test have the unique set among both folders, so that the class_to_idx is the same.

jveitchmichaelis (approved):
LGTM

PR #1133, opened 2025-09-13
Remove deprecated functions and parameters

henrykironde:
Removed Functions:

**Removed Functions:**

- `xml_to_annotations()` - Use `utilities.read_pascal_voc(path)` or the general `utilities.read_file(path)`.
- `boxes_to_shapefile()` - Use `image_to_geo_coordinates()`.
- `project_boxes()` - Use `image_to_geo_coordinates()`.
- `annotations_to_shapefile` - Use `image_to_geo_coordinates()`.
- `plot_points()` - Use `plot_results`
- `draw_points()` - Use `plot_results`
- `plot_predictions()` - Use `plot_results`
- `draw_predictions()` - Use `plot_results`
- `use_release()` - Use `load_model('weecology/deepforest-tree')` instead
- `use_bird_release()` - Use `load_model('weecology/deepforest-bird')` instead

**Removed Parameters:**

- `geometry_type` and `save_dir` from `shapefile_to_annotations()`
- `num_classes` and `label_dict` from `deepforest()` constructor - Use config file instead
- `augment` parameter from all functions - Use `augmentations` parameter instead
- `raster_path` parameter from predict_tile()  - Use `path` parameter instead

Co-authored-by: bw4sz <benweinstein2010@gmail.com>

PR #1155, opened 2025-09-29
Code cov

henrykironde:


henrykironde:
This is ready, We will not be able get it working until all the precommit work is in. #1133

henrykironde:
Sample local output, not bad
Screenshot 2025-09-29 at 4 54 09 PM

PR #1165, opened 2025-10-03
constrain transformers to < 4.57 due to DETR loading

jveitchmichaelis:
As description, only difference we can identify in failing CI is a recent `transformers` release. We were running 4.56.2 and the 4.57.0 release breaks DETR tests for some reason. I am unable to replicate on OS X.

If this passes, we should probably merge it for now until we figure out if there's some additional work we need to do to maintain compatibility with > 4.57.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,7 @@ dependencies = [
     "torch>=2.2.0",
     "torchvision>=0.17.0",
     "tqdm",
-    "transformers",
+    "transformers<4.57.0",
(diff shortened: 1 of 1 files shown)


jveitchmichaelis:
I'm struggling to get a repeatable example offline, but I have been able to trigger it on occasion. Trying to get a standalone script to figure out what exactly is causing it. It seems to be related to the image processor.

ethanwhite (approved):
LGTM

PR #1178, opened 2025-10-22
Update config arguments in training docs

ethanwhite:
epochs is now in train. save-snapshot is not in the root config and I can't find it referenced anywhere else.


jveitchmichaelis (approved):
LGTM

PR #1182, opened 2025-10-23
wip: add detr keypoint architecture

jveitchmichaelis:
@bw4sz no direct training integration yet, but this is what I'm thinking for the modeling.

The implementation looks verbose, but that's because a lot of the pieces are copied from `transformers` with very minor changes - like a reference to bbox changed to point. To modify the arch we need:

`DeformableDetrKeypointConfig`: Same as object detection but explicit parameter for point loss cost.
`DeformableDetrKeypointDetectionOutput`: Mostly similar but name changes.
`DeformableDetrForKeypointDetection`: Model, changes number of decoder outputs to 2 and removes some bbox specific bits.
`DeformableDetrKeypointMatcher`: Hungarian matcher that uses L1 or L2 cost instead of IoU.
`DeformableDetrKeypointLoss`: L1 or L2.
`DeformableDetrKeypointImageProcessor`: Normalizes `keypoints` to relative pixel coords and some other bits.

I don't anticipate checking in the current test suite as the overfit test takes a minute or two to run, but it does converge.

### DIFF
--- a/src/deepforest/datasets/training.py
+++ b/src/deepforest/datasets/training.py
@@ -3,6 +3,7 @@
 import os
 
 import numpy as np
+import pandas as pd
 import shapely
 import torch
 from PIL import Image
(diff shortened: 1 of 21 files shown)


src/deepforest/datasets/training.py
```diff
@@ -106,7 +106,7 @@ def collate_fn(self, batch):
         return images, targets, image_names
 
     def load_image(self, idx):
-        img_name = os.path.join(self.root_dir, self.image_names[idx])
+        img_name = os.path.join(self.root_dir, os.path.basename(self.image_names[idx]))
```
jveitchmichaelis: Will remove this, or make it an option.

PR #1185, opened 2025-10-27
add tests and fix renaming of columns in read_file, our functions should handle any error internally and not change the users input data

bw4sz:


jveitchmichaelis (commented):
LGTM! Will merge when CI completes

PR #1195, opened 2025-10-31
Transform bounding box coordinate columns along with geometry

### LINKED ISSUE #1193: image_to_geo_coordinates updates geometry but not the bounding box position columns
`predict_tile()` returns the data in two formats:
1. a geometry column in the gdf
2. bounding box position columns (xmin, xmax, ymin, ymax)

`image_to_geo_coordinates()` correctly updates the geometry, but the bounding box position columns aren't touched leaving them in image coordinates. This produced a confusing bug in our everglades workflow.

The function should re-extract these position columns from the geometry so that the two versions of the position data are consistent.

ethanwhite:
`predict_tile()` returns bounding box data in two formats:

1. a geometry column in the gdf
2. bounding box position columns (xmin, xmax, ymin, ymax)

`image_to_geo_coordinates()` correctly updates the geometry, but the bounding box position columns aren't touched leaving them in image coordinates. This change extracts these position columns from the geometry after transformation so that the two versions of the position data are consistent.

Fixes #1193

### DIFF
--- a/src/deepforest/utilities.py
+++ b/src/deepforest/utilities.py
@@ -783,6 +783,15 @@ def image_to_geo_coordinates(gdf, root_dir=None, flip_y_axis=False):
             projected_geometry.append(geom)
 
     transformed_gdf.geometry = projected_geometry
+
+    # Update xmin, xmax, ymin, ymax columns to match transformed geometry
+    if geom_type == "box":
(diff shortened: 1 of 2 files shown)


ethanwhite:
I don't think I needed to touch the flipped y stuff here, but let me know if I'm missing something

Issue #834, opened 2024-11-11
There is no reason to ask for the height, width on plot_results.


bw4sz:
If the data is well formatted.

```
gdf.head()
   id                                            polygon  ...                  source                                           geometry
0   1  POLYGON ((79 683,78 684,77 684,76 685,75 686,7...  ...  Justdiggit et al. 2023  POLYGON ((79 683, 78 684, 77 684, 76 685, 75 6...
1   2  POLYGON ((275 323,274 324,273 324,272 324,271 ...  ...  Justdiggit et al. 2023  POLYGON ((275 323, 274 324, 273 324, 272 324, ...
2   3  POLYGON ((197 239,196 240,195 240,194 240,194 ...  ...  Justdiggit et al. 2023  POLYGON ((197 239, 196 240, 195 240, 194 240, ...
3   4  POLYGON ((996 465,995 466,994 467,993 466,992 ...  ...  Justdiggit et al. 2023  POLYGON ((996 465, 995 466, 994 467, 993 466, ...
4   5  POLYGON ((934 582,933 583,932 584,931 584,930 ...  ...  Justdiggit et al. 2023  POLYGON ((934 582, 933 583, 932 584, 931 584, ...

[5 rows x 6 columns]
```

```
gdf.root_dir
'/orange/ewhite/DeepForest/justdiggit-drone/label_sample/'
```

```
    plot_results(gdf)
Traceback (most recent call last):
  File "/blue/ewhite/b.weinstein/miniconda3/envs/MillionTrees/lib/python3.10/site-packages/deepforest/visualize.py", line 433, in plot_results
    annotated_scene = _plot_image_with_results(df=results,
  File "/blue/ewhite/b.weinstein/miniconda3/envs/MillionTrees/lib/python3.10/site-packages/deepforest/visualize.py", line 485, in _plot_image_with_results
    detections = convert_to_sv_format(df, height=height, width=width)
  File "/blue/ewhite/b.weinstein/miniconda3/envs/MillionTrees/lib/python3.10/site-packages/deepforest/visualize.py", line 323, in convert_to_sv_format
    raise ValueError(
ValueError: height and width of the mask must be provided for polygon predictions
```

to fix, just go and get the height and width within the function, everything is there.

ethanwhite:
Fixed in #1010

PR #1198, opened 2025-11-04
Remove width and height from plot_results and plot_annotations

bw4sz:
While working on MillionTrees I noticed an annoying patterns for visualizing polygon data, we asked the user to separately specify the width and height of the image, even though we had the image. This was partially fixed in #834 but not fully, since the arguments still remained. 

---

> [!NOTE]
> Simplifies visualization APIs by auto-deriving image dimensions and improving image loading/handling.
> 
> - Remove `width`/`height` params from `plot_results`, `plot_annotations`, and `_plot_image_with_geometry`; now pass the `image` to `convert_to_sv_format`
> - Update `convert_to_sv_format` to accept `image`; for polygons, get dimensions from the provided image or from `df.root_dir + image_path` and generate masks accordingly
> - Enhance `_load_image` to resolve images from path/df with `root_dir`, normalize channel order, drop alpha, and cast dtype to `uint8` (with rescaling when needed)
> - Tests updated/added: polygon plotting without dims, support for relative `image_path` with `root_dir`, and `_load_image` behaviors (PIL input, CHW→HWC, alpha drop, dtype rescale)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 920fb5c72b7f4b94c5b625cb7c289bb4eb4bf28a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/deepforest/visualize.py
+++ b/src/deepforest/visualize.py
@@ -206,30 +206,10 @@ def convert_to_sv_format(
                 raise ValueError("'image_path' column required for polygons.")
 
             image_path = df["image_path"].iloc[0]
+            full_path = os.path.join(df.root_dir, image_path)
 
-            # Resolve root_dir similar to _load_image
-            resolved_root = None
-            if hasattr(df, "root_dir"):
-                resolved_root = df.root_dir
-                resolved_root = (
-                    resolved_root.iloc[0]
-                    if isinstance(resolved_root, pd.Series)
-                    else resolved_root
-                )
-
-            full_image_path = (
-                os.path.join(resolved_root, image_path)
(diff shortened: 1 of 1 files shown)


bw4sz:
This was the same general logic and motivation for https://github.com/weecology/DeepForest/pull/1210

The plot and predict functions shouldn't have to deal with all this uncertainty in what the data looks like. All of this is going to be handled by the read_file, with the exception of the edge case where the image exists only in memory and not on disk. Updated code and we should wait until 1210 is merged.

PR #1200, opened 2025-11-07
Add tests for point annotations split_raster

bw4sz:
This PR was motivated by testing in MillionTrees where I noticed that when using a point geometry, preprocess.split_raster was not re-normalizing the annotations to the tiled output, making them offset. I added a test, which failed, and then a fix, which passed. 

I also played with adding to agents.md to make it clear that when interacting with cursor we expect it to run and show failing tests before it just goes makes a solution. Its quite annoying to revert part of the automation to show it failed beforehand and then was fixed.

I think when we work on the polygon model, this function might need to be refactored into tile_raster -> determine geometry -> select_*_annotations for points, boxes and polygons, but I think we should leave that to another time since we don't quite yet have polygons ready. 

---

> [!NOTE]
> Translate point annotations to patch-local coordinates during tiling and add a focused test plus testing guidelines doc.
> 
> - **Preprocessing**:
>   - Update `select_annotations` in `src/deepforest/preprocess.py` for point geometry:
>     - Translate `Point` annotations to window-origin coordinates.
>     - Populate/refresh `x` and `y` columns from translated geometry.
> - **Tests**:
>   - Add `test_split_raster_points_translate_and_bounds` in `tests/test_preprocess.py` to verify point translation, count preservation, and in-window bounds when `patch_overlap=0`.
> - **Docs**:
>   - Add `tests/AGENTS.md` with concise testing and `uv`-based test-running guidelines.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c970f2152d023bb1c85f3a06b6e24edd7df868ad. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,6 +1,6 @@
 # Project Overview
 
-DeepForest is a Python library for object detection in aerial images, primarily trees, but also supporting models for livestock and other objects of interest. It uses pytorch throughout and the Lightning framework for model training, prediction and evaluation.
+DeepForest is a Python library for object detection in aerial images for biodiversity applications such as tree detection and wildlife observation. It uses pytorch and Pytorch Lightning framework for model training, prediction and evaluation.
 
 Imagery is of a relatively high resolution and is generally not satellite data.
 
@@ -8,7 +8,7 @@ Imagery is of a relatively high resolution and is generally not satellite data.
 
 - We use `uv` for package management
 - NEVER edit dependencies in pyproject.toml directly, ALWAYS use `uv add`
-- To set up, use `uv sync --all-extras --dev`. You can reference the CI workflow in .github
+- To set up and run tests, use `uv sync --all-extras --dev`. You can reference the CI workflow in .github
 - Use the `deepforest` command line interface with Hydra overrides for simple tests
 - When you've finished, run `uv run pre-commit` with appropriate arguments to run formatting and linters on your work.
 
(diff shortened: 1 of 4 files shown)


jveitchmichaelis (commented):
Code looks good.

On the AGENTS - I wonder if your addition should be a top-level change. Is it a general problem when the bot powers ahead and implements something without testing first? Or is it when it's made an implementation and doesn't test it? I definitely advocate for making it run the full test suite after it's done (and possibly before).

jveitchmichaelis (changes requested):
Sounds good, I think we need a bit of de-duplication in the AGENTS file: 

- There are a couple of sections with how-to-run instructions
- Probably don't need the top-level heading or change to something like "Testing guidelines for Agents" (not explicitly cursor)
- `If permission prompts appear in Cursor (e.g., network), request and obtain them to complete installs and test runs.` -> change this to be more generic (like `if interactive prompts appear ... `)

bw4sz:
I hadn't noticed that the new Agents.md was in /tests, and not in the original, which is why there was duplicated. Fixed and combined.

jveitchmichaelis:
Ah my bad here - `tests/AGENT.md` is intentional as it indicates to agents that those rules relate to unit tests ([directory scoped rules](https://developers.openai.com/codex/guides/agents-md/#layer-project-instructions); this example is OpenAI Codex but Anthropic and copilot also support). Though I have no empirical evidence that suggests it's better or worse if we have one file vs several.

bw4sz:
okay, I move that back to a separate file.

PR #1202, opened 2025-11-07
Better support for composable configs

jveitchmichaelis:
This PR adds improved support for loading custom and named configs.

- Single level of inheritance for configs (i.e. derive from defaults/`config`). This was technically supported, but I think was not thoroughly tested.
~~- Add core "named" configs (tree, bird, livestock)~~
~~- main now clearly defaults to `config='tree'`~~
- Users can provide a config path directly to `deepforest.main`
- Removed `MISSING` from the schema as not necessary
~~- Adds a `task` key in preparation for keypoint/polygon modeling~~
- A few more tests.

Notes:

- Using `MISSING` indicates a required field that is likely custom, like a `csv_file`. However, this forces derived config files to set all of those fields. In practice it doesn't make a difference because we have error handling for missing paths.
- We discussed using Hydra throughout, but that gets messy when calling DeepForest from another pipeline that uses it. It's cleaner to keep Hydra purely for the CLI.
- However, Hydra can do more complicated things (like multiple layers of inheritance)
- OmegaConf doesn't actually support inheritance, it's managed by `.merge`. We do this by merging in the following order:
  - base (always)
  - supplied config (optional built-in yaml file or path)
  - overrides (optional)
- Handling label_dict is awkward because of how OmegaConf merges dictionaries, so there is still a hack to make sure that if the class list is updated, it's properly overriden.

### DIFF
--- a/src/deepforest/conf/schema.py
+++ b/src/deepforest/conf/schema.py
@@ -1,7 +1,5 @@
 from dataclasses import dataclass, field
 
-from omegaconf import MISSING
-
 
 @dataclass
 class ModelConfig:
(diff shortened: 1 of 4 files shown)


jveitchmichaelis:
Fixed-up to remove the named configs for now, but support is in there if we want to add later.

PR #1015, opened 2025-04-05
Added validation check for bounding boxes

reddykkeerthi:
Fixes #1014 
Implemented vectorized validation using Pandas operations to check *all bounding boxes per image simultaneously* for boundary violations (negative coords, out-of-image bounds, inverted boxes). Replaced row-wise iteration with batched `(xmin >= 0) & (xmax <= width)`-style checks on entire annotation subsets, improving efficiency for large datasets. Validation errors aggregate all issues per image for clearer debugging.

PR #1211, opened 2025-11-15
Add validation checks for bounding boxes

### LINKED ISSUE #1014: Add validation check to ensure that all bounding boxes are valid for training/evaluation
Periodically we see issues where something is working for a user performing training and it turns out to be that there are bounding boxes that extend beyond the edges of the images they are associated with (#989 is the most recent example).

We should add a validation step to ensure that all bounding boxes associated with an image are within the size bounds of that image and report clearly that this is the issue to make it easier for the user to figure this out.

ethanwhite:
Checks to see if bounding boxes occur outside of image boundaries and clearly communicates to user when they do.

This is a version of #1015 that accounts for changes in the codebase over the last year. This was good work by @reddykkeerthi that I didn't want to see get lost.

Closes #1014

### DIFF
--- a/src/deepforest/datasets/training.py
+++ b/src/deepforest/datasets/training.py
@@ -1,5 +1,6 @@
 """Dataset model for object detection tasks."""
 
+import math
 import os
 
 import numpy as np
@@ -71,6 +72,7 @@ class BoxDataset(Dataset):
         self.preload_images = preload_images
 
         self._validate_labels()
+        self._validate_coordinates()
 
         # Pin data to memory if desired
         if self.preload_images:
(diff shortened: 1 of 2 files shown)


jveitchmichaelis:
I guess that's in the dino branch? Struggling to see where that commit belongs - it looks like I added it when I was testing the pretrain data and then squashed. So we can keep this one if it didn't make it into main.

jveitchmichaelis (commented):
The coverage checker is requesting failure tests for image path and geometry, might not be a bad idea to add those.

`test_validate_dataset_missing_image` and `test_validate_dataset_bad_geometry` etc.

jveitchmichaelis:
Yeah it's not clear to me how you'd fail that. Unless it's possible to store a bad box in WKT format and Shapely blindly loads the bounds? But otherwise I think the only way to trigger that error would be to have more (or less) than 4 bounding coordinates. Or if the shape doesn't have `bounds` defined - like a point?

ethanwhite:
> Yeah it's not clear to me how you'd fail that. Unless it's possible to store a bad box in WKT format and Shapely blindly loads the bounds? But otherwise I think the only way to trigger that error would be to have more (or less) than 4 bounding coordinates. Or if the shape doesn't have `bounds` defined - like a point?

Should I just go ahead and pull that try/except block then?

ethanwhite:
Thanks @jveitchmichaelis - changes made. Apologies for the linter changes in the test file. We should probably be running ruff on those so they don't get behind like this this.

henrykironde (approved):
I think this works for the cases we have.

henrykironde:
Thanks @reddykkeerthi and @ethanwhite . LGTM

PR #1229, opened 2025-12-10
Allow user to expand cropmodel boxes for predict_tile

bw4sz:
While working on the BOEM pipeline I noticed a strong performance difference between how annotators make detections and the deepforest detection models, leading to weaker cropmodel classification performance. This PR allows the user to specify in config to expand the detection both by x pixels on all sides to capture a wider extent. 

Agent statement: I conceived of the change and wrote the initial function. Cursor wrote the test which I then showed failed and that my fix improved. Cursor added the edge cases that the bounding box expansion goes off the side of the image.

### DIFF
--- a/src/deepforest/conf/schema.py
+++ b/src/deepforest/conf/schema.py
@@ -107,6 +107,7 @@ class CropModelConfig:
     scheduler: SchedulerConfig = field(default_factory=SchedulerConfig)
     balance_classes: bool = False
     resize: list[int] = field(default_factory=lambda: [224, 224])
+    expand: int = 0
 
 
 @dataclass
--- a/src/deepforest/datasets/cropmodel.py
+++ b/src/deepforest/datasets/cropmodel.py
@@ -49,12 +49,23 @@ class BoundingBoxDataset(Dataset):
         transform: Optional transform function
         augmentations: Augmentation configuration
         resize: Optional list of [height, width] for resizing. Defaults to [224, 224]
+        expand: expand: number of context pixels to add to input bounding boxes
+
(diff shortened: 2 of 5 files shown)


src/deepforest/conf/schema.py
```diff
@@ -107,6 +107,8 @@ class CropModelConfig:
     scheduler: SchedulerConfig = field(default_factory=SchedulerConfig)
     balance_classes: bool = False
     resize: list[int] = field(default_factory=lambda: [224, 224])
+    # Number of pixels to expand bbox crops for additional context (not padding)
```
jveitchmichaelis: Drop this comment for consistency with the rest of the file

src/deepforest/datasets/cropmodel.py
```diff
@@ -49,12 +49,23 @@ class BoundingBoxDataset(Dataset):
         transform: Optional transform function
         augmentations: Augmentation configuration
         resize: Optional list of [height, width] for resizing. Defaults to [224, 224]
+        expand: Number of pixels to expand the bounding box on all sides before
```
jveitchmichaelis: Suggest:

```
expand: Number of pixels to expand bounding box windows in source image, boxes will be expanded in all directions and clipped to image extent.
```

jveitchmichaelis (changes requested):
Added a few suggestions + request a test for boxes that would expand out of bounds.

jveitchmichaelis:
Looks good to me, just needs a squash

jveitchmichaelis (approved):

PR #1230, opened 2025-12-11
Replace Albumentations with Kornia

jveitchmichaelis:
Supersedes https://github.com/weecology/DeepForest/pull/1169

Seems like a lot of changes, but a good chunk is new docs + re-implementations of some augmentations that `kornia` doesn't support out of the box.

Fully replaces `albumentations` with `kornia`. On the user end, very little changes except to removing `RandomSizedBBoxSafeCrop`, but we have a mostly equivalent crop operation and we can handle empty training inputs.

One gotcha is `kornia` doesn't do any output filtering after transforming. It composes its pipeline into a transformation that's applied to all geometries, but we need to check if those lie within the image or not. There's a new dataset function to do this, though perhaps in the future could be rolled into the recent box validation function instead (we need to return to filter and not just flag).

I also put some notes about zoom transforms, as `DownScale` is poorly named in Albumentations, with some guidelines for reproducing the effect through a double `Rescale` if it's important.

transforms

Example transformations included in the commit.


jveitchmichaelis:
re. codecov:

- We can ignore the diff in augmentations as it refers to an inverse transform which I copied from the relevant `kornia` class, but I don't think we use it.
- I can add some tests for when all boxes are removed, just to check that training doesn't outright fail

PR #1241, opened 2025-12-18
Image callback should handle empty images

jveitchmichaelis:
With stronger augmentation it's possible that images are transformed to exclude bounding boxes. We handle this in the dataset, but the callback will fail because it attempts to construct an annotation dataframe for viz. This PR adds a check and plots the image on its own if no annotations are found.


bw4sz:
I needed this too.

bw4sz (approved):

Issue #1246, opened 2025-12-23
Logged images don't seem to respect the minimum score in config when starting from loaded model.


bw4sz:
<img width="1647" height="1019" alt="Image" src="https://github.com/user-attachments/assets/f45249fb-057e-4c97-abd6-28f57151f5b9" />

Note the min score is around 0.1, which is below the config. On inspection, the retinanet model (I don't know if this is a structure that is in DETR), has the older value.

```
m.config.score_thresh
0.25
m.model.score_thresh
0.1
```
We need to update the score after load_model?

bw4sz:
fixed #1263

Issue #1245, opened 2025-12-22
Don't log class scores if there is just one class.


bw4sz:
Its confusing to have 'bird recall' and 'bird precision' equal to 1, reported from model training, the user might confuse this with the box recall and precision. That's the class scores, and of course its 1 there is only one class...

<img width="1528" height="597" alt="Image" src="https://github.com/user-attachments/assets/875620ed-2d38-4377-8caa-493bc0f4e195" />

Issue #1244, opened 2025-12-19
Validate annotations takes a very long time for large datasets.


bw4sz:
I waited nearly an hour before receiving: 

```
[rank0]: Traceback (most recent call last):
[rank0]:   File "/blue/ewhite/b.weinstein/src/DeepForest/src/deepforest/scripts/train_birds.py", line 207, in <module>
[rank0]:     run()
[rank0]:   File "/blue/ewhite/b.weinstein/src/DeepForest/src/deepforest/scripts/train_birds.py", line 181, in run
[rank0]:     m.trainer.fit(m)
[rank0]:   File "/blue/ewhite/b.weinstein/src/DeepForest/.venv/lib/python3.12/site-packages/pytorch_lightning/trainer/trainer.py", line 584, in fit
[rank0]:     call._call_and_handle_interrupt(
[rank0]:   File "/blue/ewhite/b.weinstein/src/DeepForest/.venv/lib/python3.12/site-packages/pytorch_lightning/trainer/call.py", line 48, in _call_and_handle_interrupt
[rank0]:     return trainer.strategy.launcher.launch(trainer_fn, *args, trainer=trainer, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/blue/ewhite/b.weinstein/src/DeepForest/.venv/lib/python3.12/site-packages/pytorch_lightning/strategies/launchers/subprocess_script.py", line 105, in launch
[rank0]:     return function(*args, **kwargs)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/blue/ewhite/b.weinstein/src/DeepForest/.venv/lib/python3.12/site-packages/pytorch_lightning/trainer/trainer.py", line 630, in _fit_impl
[rank0]:     self._run(model, ckpt_path=ckpt_path, weights_only=weights_only)
[rank0]:   File "/blue/ewhite/b.weinstein/src/DeepForest/.venv/lib/python3.12/site-packages/pytorch_lightning/trainer/trainer.py", line 1079, in _run
[rank0]:     results = self._run_stage()
[rank0]:               ^^^^^^^^^^^^^^^^^
[...]

Issue #1238, opened 2025-12-18
main.evaluate() gives incorrect results when validation images are not all the same size. 


bw4sz:
```
# Bug Report: `evaluate()` ignores `config.validation.size`

## Description

The `evaluate()` method does not use `config.validation.size` even when it is set, causing a RuntimeError when images have different dimensions. The size parameter must be passed explicitly to `evaluate()`, which is inconsistent with how other methods respect config settings.

## Expected Behavior

When `config.validation.size` is set, the `evaluate()` method should use this value by default (similar to how training and validation dataloaders use it).

## Actual Behavior

`evaluate()` ignores `config.validation.size` and requires the `size` parameter to be passed explicitly. This causes the following error when evaluating on images with different dimensions:

```
RuntimeError: Images in batch have different dimensions. Set validation.size in config.yaml to resize all images to a common size.
```

## Steps to Reproduce

```python
from deepforest import main

m = main.deepforest()
m.load_model("weecology/deepforest-tree")

# Set validation size in config (as documented)
m.config["validation"]["size"] = 800

# This fails even though config.validation.size is set
results = m.evaluate(csv_file="test.csv", root_dir="./data")
# RuntimeError: Images in batch have different dimensions...
```

## Current Workaround

Pass `size` explicitly to `evaluate()`:

```python
# Must pass size explicitly (redundant with config setting)
results = m.evaluate(csv_file="test.csv", root_dir="./data", size=800)
```

## Minimal Reproducible Example

See `reproduce_evaluate_size_issue.py` for a complete working example.

## Suggested Fix

Modify the `evaluate()` method in `main.py` to use `config.validation.size` as default when `size` parameter is `None`:

```python
def evaluate(self, csv_file, iou_threshold=None, root_dir=None, size=None, batch_size=None, predictions=None):
    # ...
    if size is None:
        size = self.config.validation.size
    
    if predictions is None:
        predictions = self.predict_file(csv_file, root_dir, size=size, batch_size=batch_size)
    # ...
```

## Environment

- DeepForest version: [current main branch]
- Python version: 3.12
- OS: Linux

## Additional Context

This issue affects training scripts that set `config.validation.size` and then call `evaluate()` for pre/post training evaluation. The inconsistency makes the config system less intuitive.

```

jveitchmichaelis:
@bw4sz I put a larger comment in the discussion. But the key points here:

- `size` is ignored entirely during training, because `evaluate` is provided the results from the validation loop.
[...]

PR #1247, opened 2025-12-24
Update the bird model 

bw4sz:
## Description

This PR makes a script to organize and train a new bird detector. It uses data from the original Weinstein et al. 2022 paper, adds in data from the Drones for Ducks and other datasets from lila.science.

I added blank white images to test the performance and can confirm it no longer predicts in blank images with an empty frame accuracy of 100%.

# Next steps

- [x] Update docs
- [x] Compare performance to old detector
- [ ] Update weights on huggingface
- [ x] Check tiling sensitivity and optionally add more zoom augmentations.
- [x] reach out to community for other images to test against. Add these to docs.
- [x] Add a couple non-bird images to test as well.
- [x] Quick comparison about segment anything 3, just a screenshot of the browser.
- [x ] Check tradeoff in precision and recall for different score-thresholds.

# Other issues.

~There is an issue that needs to be documented in which model.evaluate() needs a size argument (below), but more importantly doesn't give the same results as within the training loop. They may be related. Let's wait until #1238 is solved and confirm. I saw the performance drop completely.~ fixed in #1238  

~I am quite confused about the CPU memory (@jveitchmichaelis did you see this in other model training). It just doesn't jive with my expectations and back of the envelope calculations. If you have 6 workers, and an average image size of 10MB, and a prefetch of 2 and batch size of 20 = 6 * 2 * 10 * 20 ~ 3GB. We are seeing HUGE memory usage, and it seems like it's more within the model.train loop, not in the dataloader. I am concerned about kornia.~ Fixed in https://github.com/weecology/DeepForest/pull/1304

Screenshot 2025-12-24 at 10 37 00 AM

## Related Issue(s)

I've made a number of issues during this PR

#1246 #1245 #1244 

## AI-Assisted Development

- [x ] I used AI tools (e.g., GitHub Copilot, ChatGPT, etc.) in developing this PR
- [ x] I understand all the code I'm submitting
- [ x] I have reviewed and validated all AI-generated code


jveitchmichaelis:
Yeah there are possibly some memory leaks. I've been trying to hunt this down with the DINO branch. You can try aggressively clearing the cache + running `gc.collect()` at the end of each epoch. It's hard to tell on HPG because you're shown the entire system RAM and not only your own process. Best to debug that locally if you can.

Another one was that the `losses` should be detached when logging, but not when being returned. Also making sure we don't return things from hooks that should not be called directly, metrics are all reset. Let me gather up my changes and PR.

image

Here is my trace for a long run after I made an effort to stop this happening, you need to ignore the "background" level. You should see it tick up (caching?) and then flatten out).

bw4sz:
Comparison to old detector show much improved performance on a 90/10 split.

```

Box Precision:
  Checkpoint:  0.8492
  Pretrained:  0.7495
  Difference:  +0.0997 (+13.30%)

Box Recall:
  Checkpoint:  0.8662
  Pretrained:  0.4645
  Difference:  +0.4017 (+86.47%)

Empty Frame Accuracy:
  Checkpoint:  1.0000
  Pretrained:  0.0000
  Difference:  +1.0000
```

This is from trainer.validate(model), the results from main.evaluate() feel muddled with #1238. We need to fully understand that issue before merging this PR. 

To do is a zero-shot comparison with a new dataset, I am asking the community for a couple images atleast.

bw4sz:
This is the huggingface PR for the model weights. https://huggingface.co/weecology/deepforest-bird/discussions/5

PR #1248, opened 2025-12-25
allow dictionary format in augmentation config schema

### LINKED ISSUE #1234: How should I list the augmentations in the config.
I'm not sure the way I should list augmentations in the new structures. 
 
```
 m.config["train"]["augmentations"] = [
      {"RandomResizedCrop": {"size": (800, 800), "scale": (0.5, 1.0), "p": 0.3}},
      {"Rotate": {"degrees": 15, "p": 0.5}},
      {"HorizontalFlip": {"p": 0.5}},
      {"VerticalFlip": {"p": 0.3}},
      {"RandomBrightnessContrast": {"brightness": 0.2, "contrast": 0.2, "p": 0.5}},
      {"HueSaturationValue": {"hue": 0.1, "saturation": 0.1, "p": 0.3}},
      {"ZoomBlur": {"max_factor": (1.0, 1.03), "step_factor": (0.01, 0.02), "p": 0.3}},
  ]
```
but that yields.

```
omegaconf.errors.ValidationError: Cannot convert 'dict' to string: '{'RandomResizedCrop': {'size': (800, 800), 'scale': (0.5, 1.0), 'p': 0.3}}'
    full_key: train.augmentations[0]
    reference_type=Optional[List[str]]
    object_type=list
```

musaqlain:
### Description
The current schema in `conf/schema.py` was restricted to `List[str]`. This caused `omegaconf.errors.ValidationError` when users tried to provide detailed augmentation configurations (dictionaries with parameters), as described in the issue.

### Changes Made
- Updated `augmentations` type hint in `TrainConfig` and `ValidationConfig` from `List[str]` to `List[Any]`.
- Added a new regression test `test_augmentation_schema_validation` in `tests/test_augmentations.py` to ensure dictionary configurations are accepted.

### Testing
- Added `test_augmentation_schema_validation` which explicitly initializes a model with a complex dictionary augmentation list (e.g., `RandomResizedCrop` with params) and asserts it is stored correctly without raising a ValidationError.
- Ran `pytest tests/test_augmentations.py` to ensure existing string-based augmentations (like `["HorizontalFlip"]`) still pass.
- Verified locally with a reproduction script.

### Linked Issue
Closes #1234

### AI-Assisted Development

### DIFF
--- a/tests/test_augmentations.py
+++ b/tests/test_augmentations.py
@@ -41,6 +41,34 @@ def test_load_dataset_without_augmentations():
     image, target, path = next(iter(train_ds))
     assert len(train_ds.dataset.transform) == 0
 
+def test_augmentation_schema_validation():
+    """
+    Test that the schema accepts a mixed list of dictionaries and strings for augmentations,
+    and that they are correctly applied to the dataset pipeline.
+    """
+    augmentations = [
+        {"RandomResizedCrop": {"size": (800, 800), "scale": (0.5, 1.0), "p": 0.3}},
+        "HorizontalFlip"
+    ]
+
+    m = main.deepforest(config_args={"train": {"augmentations": augmentations}})
+
+    # Verify Schema stored it correctly
+    assert m.config.train.augmentations == augmentations
+
+    csv_file = get_data("example.csv")
+    root_dir = os.path.dirname(csv_file)
+
+    train_ds = m.load_dataset(csv_file, root_dir=root_dir, augmentations=augmentations)
+
+    transforms = train_ds.dataset.transform
+
+    has_resized_crop = any(isinstance(t, K.RandomResizedCrop) for t in transforms)
+    has_hflip = any(isinstance(t, K.RandomHorizontalFlip) for t in transforms)
+
+    assert has_resized_crop
+    assert has_hflip
+
 """
 Augmentation parsing tests:
 """
--- a/src/deepforest/conf/schema.py
+++ b/src/deepforest/conf/schema.py
@@ -82,7 +82,7 @@ class TrainConfig:
     fast_dev_run: bool = False
     preload_images: bool = False
     validate_coordinates: bool = True
-    augmentations: list[str] | None = field(default_factory=lambda: ["HorizontalFlip"])
+    augmentations: list[Any] | None = field(default_factory=lambda: ["HorizontalFlip"])
 
 
 @dataclass
@@ -104,7 +104,7 @@ class ValidationConfig:
     iou_threshold: float = 0.4
     val_accuracy_interval: int = 20
     lr_plateau_target: str = "val_loss"
-    augmentations: list[str] | None = field(default_factory=lambda: [])
+    augmentations: list[Any] | None = field(default_factory=lambda: [])
 
 
 @dataclass


jveitchmichaelis:
Thanks for this, please could you provide test cases that would cover the changes? e.g. configs with (dict) augmentation listings that would currently fail.

henrykironde (changes requested):
@musaqlain, We expect some tests for changes in logic. 
Thank you for the contribution

jveitchmichaelis (changes requested):
Thanks for adding that! In the test case, please pass the dict to utilities.load_config or as a config_arg to main.deepforest.

The current test assigns the dict after instantiation and I'm not sure if the schema is re-validated? Maybe it does, but it's a bit clearer if we check how the code would normally be used (ideally we should set up the config once and not modify it).

musaqlain:
Thanks for the feedback! You are right, testing it during initialization is much safer to ensure validation triggers correctly. I have updated it.

jveitchmichaelis (commented):
@musaqlain thanks, I will take a closer look this week. I added a couple more suggestions, while we're adding this fix we could take a moment to improve the tests more.

The test should cover some other variations that we might typically expect; for example it's valid to mix strings ("HorizontalFlip", to use defaults) and dicts (to customize parameters). Note that we do test the parser for this behaviour earlier in the file. 

Another suggestion is to enforce stricter typing here, as `Any` is very broad. I need to check how OmegaConf handles unions, but we could try `list[union[str, dict]]` or something similar. I'm not sure if we would also have to add a specific type for the dict.

tests/test_augmentations.py
```diff
@@ -347,6 +347,22 @@ def test_geometric_augmentation_filters_boxes():
         # Labels should match box count
         assert len(labels) == len(boxes), f"Label count mismatch: {len(labels)} labels, {len(boxes)} boxes"
 
+def test_augmentation_schema_validation():
```
jveitchmichaelis: Please could you move this to the top of the file with the other integration tests, since we already have a couple for passing augmentations to main.deepforest.

musaqlain:
Thanks for the detailed review! I've updated the PR with the following changes:
**1. Robust Integration Test:**
I completely refactored `test_augmentation_schema_validation` as suggested. It now:
* Defines a **mixed list** of augmentations (String + Dictionary) to ensure both formats work together.
* Explicitly loads a dataset using these augmentations.
* inspects the actual `transforms` pipeline to assert that the correct Kornia objects (e.g., `K.RandomResizedCrop`) are instantiated.
* also moved this test function at the top of file along with other integration tests.

**2. Schema Typing:**
I attempted to enforce stricter typing using `list[str | dict[str, Any]]` (or `Union`) as requested. However, **OmegaConf** (Structured Configs) currently raises a `ValidationError: Unsupported value type` when using Unions for mixed lists.
To ensure stability, I have reverted the schema to `list[Any]`, but I added the integration test above to guarantee that valid dictionary configurations are processed correctly.

tests/test_augmentations.py
musaqlain: Done

ethanwhite:
@jveitchmichaelis / @henrykironde - looks like this PR got lost in the pile a few months ago and is probably close to going in if one of you has a few minutes to check @musaqlain's updates to the tests

musaqlain:
Hi @ethanwhite , I am currently taking my final university exams. I noticed that you marked a few of my PRs as `Awaiting author contribution`. I will be able to address the feedback and update all of them this Friday once my exams are completely finished, after that my semester will be over, and I will be completely free. Thank you for your support! 🙏 
cc @henrykironde @jveitchmichaelis

ethanwhite:
Hi @musaqlain - Good luck on your exams! No rush on our end. That's just a label that we use to help the core maintainers keep track of what we need to deal with. We know if that label is there we don't need to do anything and as soon as you comment on the PR it gets automatically removed so we know that we should take a look.

ethanwhite:
@jveitchmichaelis - looks like this one should be close to merging if you have a few minutes to give it another check

ethanwhite:
@musaqlain - it looks like we have a conflict with some recent changes. Can you please rebase against main

musaqlain:
PTAL

PR #1249, opened 2025-12-25
Test: Assert specific evaluation scores for sample data

### LINKED ISSUE #1233: Assert the outputs of sample data
We need to have tests that assert that on release, the sample data produces the same output through time for the same model. Right now we don't specify whether the model outputs themselves have changed. There is ground truth with each of the same data, so we can assert the eval score.

musaqlain:
### Description
Previously, `test_evaluate` only checked if the execution completed without errors, but did not assert the quality of the results. This PR adds assertions to ensure the model produces consistent precision and recall scores on the sample data.

### Changes Made
- Relaxed Unit Test: I modified test_evaluate in tests/test_main.py to use looser bounds (> 0.7) so it acts as a sanity check for future models.
- New Benchmark Test: I created tests/test_benchmark.py which loads the specific model revision (SHA) using m.load_model(..., revision=...). This test asserts the exact precision/recall values (approx 0.80 / 0.72) to ensure reproducibility for this specific release.

### Testing
- Ran `pytest tests/test_main.py::test_evaluate` locally.
- Result: Passed.

### Linked Issue
Closes #1233

### AI-Assisted Development

### DIFF
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -1,0 +1,18 @@
+import pytest
+from deepforest import main, get_data
+
+def test_benchmark_release():
+    """
+    Benchmark test to ensure the specific release version of the model
+    produces consistent results.
+    """
+    release_sha = "cc21436bc5d572dde8ff5f93c1e71a32f563cace"
+
+    m = main.deepforest()
+    m.load_model("weecology/deepforest-tree", revision=release_sha)
+
+    csv_file = get_data("OSBS_029.csv")
+    results = m.evaluate(csv_file, iou_threshold=0.4)
+
+    assert results["box_precision"] == pytest.approx(0.8, abs=0.01)
+    assert results["box_recall"] == pytest.approx(0.7213, abs=0.01)
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -559,11 +559,11 @@ def test_evaluate(m):
     df = pd.read_csv(csv_file)
     results = m.evaluate(csv_file)
 
-    # Metrics are sane
-    assert np.round(results["box_precision"], 2) > 0.5
-    assert np.round(results["box_recall"], 2) > 0.5
+    # Check that precision and recall don't regress below reasonable baselines
+    assert results["box_precision"] > 0.7
+    assert results["box_recall"] > 0.5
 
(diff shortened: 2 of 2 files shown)


tests/test_main.py
```diff
@@ -511,12 +511,16 @@ def test_predict_tile_from_array(m, path):
...
 
-def test_evaluate(m, tmpdir):
+def test_evaluate(m):
     csv_file = get_data("OSBS_029.csv")
     results = m.evaluate(csv_file, iou_threshold=0.4)
 
-    assert np.round(results["box_precision"], 2) > 0.5
-    assert np.round(results["box_recall"], 2) > 0.5
+    # Relaxed assertions (Sanity Check only)
+    # Allows model improvements without breaking tests
```
ethanwhite: I'd change these comments to something like:

> Check that precision and recall don't regress below reasonable baselines

ethanwhite (changes requested):
This looks really close to me. Just one recommendation for clarifying a code comment. @jveitchmichaelis - do you see anything else?

tests/test_benchmark.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+from deepforest import main, get_data
+
+def test_benchmark_release():
+    """
+    Benchmark test to ensure the specific release version of the model
+    produces consistent results.
+    """
+    # Load the model using a SPECIFIC revision (Commit SHA)
```
jveitchmichaelis: We can remove this comment too

jveitchmichaelis:
Looks like we also need to rebase, and there is another comment I would suggest removing as it doesn't provide much. I think that in the long run, we _probably_ want to use tags or branches rather than opaque commit hashes but that's not an issue right now.

musaqlain:
it is ready for review, I have resolved the conflict, squashed the commit messages into 1, and also remove those comments. 
PTAL now.

tests/test_benchmark.py
musaqlain: done

ethanwhite:
Sorry it's taken so long for us to get to this @musaqlain. It looks good to me from a quick read, but we have failing tests. Are they passing for you locally?

PR #1250, opened 2025-12-27
added `project` argument to predict_tile for geo-coordinates

### LINKED ISSUE #608: Outputs should attach geometry if available
Currently the output method is a simple plot and `boxes.head()` will give `x_min` `x_max` `y_min` `y_max` bounding boxes around detections. It would be great if, in the case the input image has geometry (i.e., .tiff images), the output also contain the geometry.

This way the image and detections can be referenced on a map easily without the need for post process georeferencing.

musaqlain:
addresses Issue #608. Previously, `predict_tile` returned pixel coordinates even if the input was a georeferenced raster. This PR adds an optional `project=False` argument to allow users to request map coordinates directly.

- updated `predict_tile` in `src/deepforest/main.py` to accept `project` (bool).
- if `project=True`, the results are converted to a GeoDataFrame using `utilities.image_to_geo_coordinates`.
- added `test_predict_tile_projected` to `tests/test_main.py` to verify that setting `project=True` returns a GeoDataFrame with valid CRS and map coordinates (UTM).

Closes #608

### AI-Assisted Development

### DIFF
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -6,6 +6,7 @@ import warnings
 import numpy as np
 import pandas as pd
 import pytorch_lightning as pl
+import rasterio
 import torch
 import torchmetrics
 from omegaconf import DictConfig, OmegaConf
@@ -577,6 +578,7 @@ class deepforest(pl.LightningModule):
         iou_threshold=0.15,
         dataloader_strategy="single",
         crop_model=None,
+        project=False,
     ):
         """For images too large to input into the model, predict_tile cuts the
         image into overlapping windows, predicts trees on each window and
@@ -593,9 +595,10 @@ class deepforest(pl.LightningModule):
                 - "batch" loads the entire image into GPU memory and creates views of an image as batch, requires in the entire tile to fit into GPU memory. CPU parallelization is possible for loading images.
                 - "window" loads only the desired window of the image from the raster dataset. Most memory efficient option, but cannot parallelize across windows.
             crop_model: a deepforest.model.CropModel object to predict on crops
+            project (bool): If True, return a geopandas.GeoDataFrame with geometry column projected to the image CRS.
 
         Returns:
-            pd.DataFrame or tuple: Predictions dataframe or (predictions, crops) tuple
+            pd.DataFrame, geopandas.GeoDataFrame, or tuple: Predictions dataframe, geopandas.GeoDataFrame, or (predictions, crops) tuple.
         """
         self.model.eval()
         self.model.nms_thresh = self.config.nms_thresh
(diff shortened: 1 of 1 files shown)


src/deepforest/main.py
```diff
@@ -531,9 +532,10 @@ def predict_tile(
                 - "batch" loads the entire image into GPU memory and creates views of an image as batch, requires in the entire tile to fit into GPU memory. CPU parallelization is possible for loading images.
                 - "window" loads only the desired window of the image from the raster dataset. Most memory efficient option, but cannot parallelize across windows.
             crop_model: a deepforest.model.CropModel object to predict on crops
+            project (bool): If True, return a geopandas.GeoDataFrame with geometry column projected to the image CRS. Defaults to False.
```
ethanwhite, line 535: It's not standard to declare the defaults in docstring so let's remove "Defaults to False." here.

src/deepforest/main.py
```diff
@@ -531,9 +532,10 @@ def predict_tile(
...
 
         Returns:
-            pd.DataFrame or tuple: Predictions dataframe or (predictions, crops) tuple
+            pd.DataFrame or tuple: Predictions dataframe or (predictions, crops) tuple. If project=True, returns a geopandas.GeoDataFrame.
```
ethanwhite, line 538: Change to "Predictions dataframe,  geopandas.GeoDataFrame, or (predictions, crops) tuple." since I don't think it's standard to describe logic here, and we already have two possible types and we're not describing the difference between them

ethanwhite (changes requested):
Sorry this got lost in the queue @musaqlain. Everything looks really good. I've requested a couple of small changes to the docstring.

My one bigger question is what happens if someone sets `project=True` and passes a path to a non-georeferenced file. It would be good to make sure that we fail informatively in this case, which might happen automatically or might require a new check either in the new `if project:` block or in `image_to_geo_coordinates`

src/deepforest/main.py
musaqlain: that makes sense

musaqlain:
> It would be good to make sure that we fail informatively in this case

Thanks for pointing this out! I have added a check in the `if project:` block that opens the image and verifies if src.crs is None.
If so, it now raises an informative ValueError and will be advising the user to either use a georeferenced raster or set project=False. Thank you and you can review this too. 👍

src/deepforest/main.py
```diff
@@ -593,9 +595,10 @@ def predict_tile(
                 - "batch" loads the entire image into GPU memory and creates views of an image as batch, requires in the entire tile to fit into GPU memory. CPU parallelization is possible for loading images.
                 - "window" loads only the desired window of the image from the raster dataset. Most memory efficient option, but cannot parallelize across windows.
             crop_model: a deepforest.model.CropModel object to predict on crops
+            project (bool): If True, return a geopandas.GeoDataFrame with geometry column projected to the image CRS.
 
         Returns:
-            pd.DataFrame or tuple: Predictions dataframe or (predictions, crops) tuple
+            pd.DataFrame, geopandas.GeoDataFrame, or tuple: Predictions dataframe, geopandas.GeoDataFrame, or (predictions, crops) tuple.
```
ethanwhite, line 601: Probably a little more cleanly: "pd.DataFrame, geopandas.GeoDataFrame, or tuple: Predictions dataframe with or without geometry or (predictions, crops) tuple.

src/deepforest/main.py
```diff
@@ -746,6 +749,28 @@ def predict_tile(
...
         formatted_results.root_dir = root_dir
 
+        if project:
+            if paths[0] is None:
+                raise ValueError(
+                    "project=True requires a file path, not an in-memory image array."
+                )
+
+            if root_dir is None:
+                root_dir = os.path.dirname(paths[0])
```
ethanwhite, line 759: Does the code starting on Line 723 already cover this?

ethanwhite (changes requested):
Looks really good. One additional docstring change suggestion and a question about whether there is some duplicate code being added

ethanwhite:
Replaced by #1423

PR #1253, opened 2025-12-30
Simplify evaluation and prediction to mirror training datasets and create tests to assert expected eval behavior

bw4sz:
* This PR favors trainer.validate over main.evaluate since it closely matches pytorch lightning. I considered renaming main.evaluate to main.__evaluate__ or deprecating it to hide it from users, I welcome input on that.
*  We want one standard evaluation path and then assert the behavior of this matches the inference path. I started on this road by adding tests to check parity between main.evaluate and trainer.validate. 
* I also simplified the outputs of trainer.validate since if this is going to be primary evaluation vehicle, it has too many map products for the average user.
* I removed the size and batch size argument from main.evaluate and had it routed through the config, in general we want to config to control as much as possible #1240 as we discussed here. 
* I simplified the prediction dataset workflow to just accept lists instead of tensors. This is faster, cleaner and easier to read. There is one additional complication for the MultiImage dataset since the batch comes from several images and we need to keep track of which in order to put it all back together.
* I wrote several tests to assert constancy of actions between evaluate and predict making the whole system more cohesive.

## AI-Assisted Development

- [x ] I used AI tools (e.g., GitHub Copilot, ChatGPT, etc.) in developing this PR
- [ x] I understand all the code I'm submitting
- [ x] I have reviewed and validated all AI-generated code

I used cursor planning mode to help structure the tests, which i then edited and simplified. 

---

> [!NOTE]
> Major refactor to streamline augmentation, evaluation, prediction, and I/O with extensive tests.
> 
> - Migrate augmentations to `kornia` (remove `albumentations`), add `ZoomBlur` and `RandomPadTo`, and switch dataset transforms to `AugmentationSequential`
> - Standardize evaluation via `trainer.validate` with simplified mAP logging; deprecate `main.evaluate` (internal `__evaluate__` retained)
> - Simplify prediction datasets: use list-based batches, track sub-batch/window indices for `MultiImage` and tiled rasters, and update `predict`/`predict_tile` postprocessing
> - Overhaul `utilities.read_file` and geospatial handling with `DeepForest_DataFrame`, explicit `image_path`/`label`/`root_dir` assignment, and improved COCO/shape conversions
> - Update visualization to infer dimensions from `image`/`root_dir` (remove explicit `width`/`height`), and harden callbacks for empty annotations
> - Config/schema: add `log_root`; crop model: add bbox `expand` and dataset support
> - Adjust datasets (`training`, `prediction`, `cropmodel`) for new transforms, normalization, and box filtering; ensure 3-channel checks
> - Update docs (user guide, HISTORY), README link fix, and add `kornia` dependency; tweak `codecov.yml` to mark patch status informational
> - Add comprehensive tests for augmentations, datasets, callbacks, CLI, evaluation parity, crop model, DETR, and prediction batching
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 71728ea351b46f9ce28adeaac9cefe1b26b1a0c3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -82,6 +82,7 @@ class deepforest(pl.LightningModule):
         self.create_trainer()
 
         self.model = model
+        self.original_batch_structure = []
 
         if self.model is None:
             self.create_model()
@@ -475,7 +476,10 @@ class deepforest(pl.LightningModule):
         return result
 
     def predict_file(
-        self, csv_file, root_dir, crop_model=None, size=None, batch_size=None
+        self,
+        csv_file,
+        root_dir,
+        crop_model=None,
     ):
         """Create a dataset and predict entire annotation file CSV file format
         is .csv file with the columns "image_path", "xmin","ymin","xmax","ymax"
@@ -492,8 +496,8 @@ class deepforest(pl.LightningModule):
             df: pandas dataframe with bounding boxes, label and scores for each image in the csv file
         """
 
-        ds = prediction.FromCSVFile(csv_file=csv_file, root_dir=root_dir, size=size)
-        dataloader = self.predict_dataloader(ds, batch_size=batch_size)
+        ds = prediction.FromCSVFile(csv_file=csv_file, root_dir=root_dir)
+        dataloader = self.predict_dataloader(ds, batch_size=self.config.batch_size)
         results = predict._dataloader_wrapper_(
             model=self,
             crop_model=crop_model,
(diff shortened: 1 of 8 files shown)


src/deepforest/main.py
```diff
@@ -866,28 +888,18 @@ def predict_step(self, batch, batch_idx):
...
-        if split_results:
-            results = []
-            batch_size = len(batch_results) // original_list_length
-            for i in range(original_list_length):
-                start_idx = i * batch_size
-                end_idx = start_idx + batch_size
-                results.append(batch_results[start_idx:end_idx])
-            return results
+        if isinstance(batch, dict):
+            images = batch["batch"]
```
jveitchmichaelis: May be clearer to label the "batch" key -> batch["images"]?

jveitchmichaelis (changes requested):
Only a few comments for readability + needs rebase, but looks good. I using Lightning-y methods is the right direction to avoid confusion. Once this is in, I'll rebase #1256 which should make this even tidier.

ethanwhite (approved):

PR #1255, opened 2026-01-01
Improve CLI and separate into sub-scripts

jveitchmichaelis:
## Description

This is a cleaned up version of the training script I used for LIDAR pre-training. I took some inspiration from the bird branch as well to add some more comet features. Docs are updated for all three scripts.

My goal is that users should never _need_ to write scripts and we can be opinionated about defaults that we commonly use ourselves.

- Split tools into sub scripts for easier organization and testing
- Improvements to the basic training script to support common training situations:
  - Comet/Tensorboard logging, optional
  - Sensible defaults for log directories, including using comet's experiment key if generated/set and tags
  - Defaults for callbacks (image + checkpoints)
  - Resume support
  - Optional tracing
- Predict script that supports single/tiled/CSV
- Evaluate script that takes a prediction CSV from above, or runs on validation.csv_file
- Better test coverage
- Reset the default log root directory to `./lightning_logs` (as opposed to $PWD).
- Small fixup in checkpoint tests where we were using a non-standard key for the log directory
 
## AI-Assisted Development

Some use of Claude Code for assisted testing (e.g. checking it can start training jobs).


bw4sz (approved):
I agree. There are some options that I never personally use, like tensorboard, but I don't think that should limit merging. The whole point here is that the user is not looking at the script, they are using the CLI, so I care less about the large amount of code, since the user doesn't engage with it directy.

PR #1256, opened 2026-01-02
add torchmetrics wrapper for evaluate_boxes

jveitchmichaelis:
## Description

- Use `torchmetrics` to collect results and call `evaluate_boxes` during validation
- All metrics now respect `validate_on_epoch`, but should be fast enough even with large datasets to run with `n=1`. Metrics are always reset regardless.
- Simpler empty frame tracking during `validation_step`
- Much cleaner logging in `main`. Non-loggable metrics are dropped in the torchmetric class.
- Metric set up is delegated to when the trainer is created. See next point.
- If the user changes anything like validation data etc after initialization, we need to also re-init the metrics because precision/recall needs to know about label_dict and the annotation file when it calls `evaluate` under the hood. There are some guards in the code to check this (e.g. reload metrics in create_trainer and on model load).
- There is also a guard to not init metrics unless a validation set is defined, which is required for our internal metric (since it needs to know about the gt dataframe to run). Case when the user sets validation csv/root after init, instead of a config arg. I would suggest we think about decoupling this in evaluate for training (ie the core eval methods should only care about pairs of pred/target and not worry about image paths.
- The metric has a `task` arg which is set to `box` for now, as I anticipate that we will modify it to support different prediction types.
 
## Related Issue(s)

https://github.com/weecology/DeepForest/issues/901 (step towards this)
https://github.com/weecology/DeepForest/pull/1254 
https://github.com/weecology/DeepForest/issues/1245
Supports https://github.com/weecology/DeepForest/pull/1253

## AI-Assisted Development

Claude Code to do some bug hunting and experiments with gpu sync that turned out to be not required.

### DIFF
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -3,12 +3,10 @@ import importlib
 import os
 import warnings
 
-import geopandas as gpd
 import numpy as np
 import pandas as pd
 import pytorch_lightning as pl
 import torch
-from lightning_fabric.utilities.exceptions import MisconfigurationException
 from omegaconf import DictConfig, OmegaConf
 from PIL import Image
 from pytorch_lightning.callbacks import LearningRateMonitor
@@ -19,6 +17,7 @@ from torchmetrics.detection import IntersectionOverUnion, MeanAveragePrecision
 from deepforest import evaluate as evaluate_iou
 from deepforest import predict, utilities
 from deepforest.datasets import prediction, training
+from deepforest.metrics import RecallPrecision
 
 
 class deepforest(pl.LightningModule):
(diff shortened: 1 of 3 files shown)


bw4sz:
Is this ready for review? I'm seeing multi-gpu core dumps currently after 1 epoch, so i'm pretty sure its something about eval. Are you able to test this PR on 2 gpus? Not sure if its related, could be logging images, or logging predictions. Not sure which.

bw4sz:
Let me see if a monkey merge it into train_birds just a moment, run that branch, which shouldn't have any changes to src and then see if it helps the core dump.

bw4sz (approved):
I am approving these changes with the belief that the multi_gpu errors i'm seeing in bird_training are unrelated. The fact that these tests pass is indication that we have improved the validation setup and it will help us with points and polygons.

Issue #1260, opened 2026-01-11
Consider replacing DeformableDetr with AutoModelForObjectDetection


jveitchmichaelis:
**Please describe A feature you would like to be added.**
Currently we have a model class that supports DeformableDetr. With a few changes in another branch, I'm at a point where I can train models fairly reliably, albeit unsure if they're getting the performance we want (even with more recent varients, convergence seems quite slow). In theory we can support any other object detection model in `transformers` that has a similar API via `AutoModel`. The challenge is how to specify which model should be loaded. We currently ask for a model.name, model.revision and there is a high-level "architecture".

The proposal is to replace the existing class with one called `transformers_object_detection` or something similar (since we will likely want to support other models for segmentation/instance later). The structure is identical, except it would use `AutoModelForObjectDetection` and rely on the mostly uniform API that transformers exposes.

The basic usage scenarios are as follows:

1. Training:
- Provide a `model.name` which is a path or repo_id. Here, the path must be a HuggingFace compatible checkpoint. This is not the same as a Lightning checkpoint. `transformers` figures out which architecture to load based on the name.
- Training proceeds as usual and at the end we have two checkpoints: a Lightning .ckpt file which captures the state of the run completely, and a HuggingFace checkpoint created by `model.save_pretrained` which is a safetensors + json in a folder.
2. Resuming or extended training
- Use `main.deepforest.load_checkpoint('path.ckpt', config_args={})` in some form.
- For resuming, usually we would load `main.deepforest(config)` as usual, then call `model.fit(ckpt_path= ... )`. This picks up the training state.
- For modifications to training, say you modified the number of epochs. You could still do this by modifying the `config_args` which will set up the trainer as desired.
3. Fine-tuning
- This is identical to (1) except `model.name` is the path to the HF weights, and `transformers` will load the model as desired.

Retinanet works in fundamentally the same way, since it has `save_pretrained` via the HF hub mixin. 

**Describe the solution you'd like**

Here are the test examples, you can see a working version [here](https://github.com/jveitchmichaelis/DeepForest/blob/dino-eval-fix/tests/test_main_checkpointing.py) where I've tried to capture the various workflows we're likely to use. I've been using this workflow for the large scale pretraining experiments and it feels like the most pragmatic approach: checkpoints for resume, huggingface weights for fine-tuning.

Train and resume:
```
def test_train_and_resume(config):
    """Test resume training from checkpoint"""

    m = main.deepforest(config=config)

    # Train and save model
[...]

PR #1263, opened 2026-01-13
More consistent checkpoint handling.

jveitchmichaelis:
## Description

This PR fixes a some usability issues in config and checkpoint loading.

It adds a more thorough test suite for training, checkpoints, resuming/continuing and pretain/fine-tune workflows. The tests are parametrized over Retinanet and DeformableDETR and I've moved them into a separate file as `test_main` is getting quite long. This is the bulk of the PR in terms of line count, the logic changes are pretty small.

The main changes are:

- `label_dict` and `num_classes` are undefined by default. The model classes will obtain label mappings on load and will populate the config.
- If the user supplies a label_dict, it will override the model if it differs. This is expected for fine-tuning (e.g. tree -> bird). If the label dict matches the loaded model, nothing happens.
- If nothing is passed and the user tries to create a from-scratch model, this will fail. And there are some basic checks for mismatches between num_classes/label_dict. We can probably plan to deprecate num_classes from the config as it's fully defined by label_dict anyway.
- Exposes `save_pretrained` on DeformableDetr. Note that this is the expected way to store models for re-use/distribution as it's a different format to Lightning and is what goes in `model.name` (works locally too). TODO update docs on this.
- The tests include usage of checkpoints (`save_checkpoint()`), HF weights (`save_pretrained`) and check that the correct modified labels are recovered after training. We also have tests for `ckpt_path` on `fit` (resuming interrupted training) and continuing training from `load_checkpoint` which is roughly equivalent.
- Fixes the bug where config_args aren't respected by `load_checkpoint`.

TODO: update docs on best practices for checkpoints.

## Related Issue(s)

https://github.com/weecology/DeepForest/issues/1260
https://github.com/weecology/DeepForest/issues/1246

## AI-Assisted Development

Claude + GitHub Copilot assisted writing tests and figuring out some of the more obscure features in HuggingFace (e.g. that we need to store model config as `_hub_mixin_config`). The test suite was largely written by hand to try and mimic the kinds of workflows we actually use.

### DIFF
--- a/docs/user_guide/09_configuration_file.md
+++ b/docs/user_guide/09_configuration_file.md
@@ -18,7 +18,6 @@ batch_size: 1
 
 # Model Architecture
 architecture: 'retinanet'
-num_classes: 1
 nms_thresh: 0.05
 score_thresh: 0.1
 
@@ -29,9 +28,16 @@ model:
 
 # If this label dict is specified, and it differs
 # from the model downloaded from the hub, the model
-# will be updated to reflect the new class list.
+# will be updated to reflect the new class list. By
+# default, this is blank and is populated when the
+# model is loaded.
+#
+# label_dict:
+#   Tree: 0
+# num_classes: 1
+#
 label_dict:
-    Tree: 0
+num_classes:
 
 # Pre-processing parameters
 path_to_raster:
--- a/docs/user_guide/11_training.md
+++ b/docs/user_guide/11_training.md
@@ -77,7 +77,7 @@ which will create an initialized RetinaNet model with 3 classes, ready for train
 
 ## Custom datasets with other classes
 
-If you need to re-train a "tree" detection model to work in your specific survey area or ecosystem, you don't need to do anything. Models themselves have no understanding of the label "tree", they output predictions corresponding to numerical class IDs (starting at 0). The default config sets `{"Tree": 0}`.
+If you need to re-train a "tree" detection model to work in your specific survey area or ecosystem, you don't need to do anything. Models themselves have no understanding of the label "tree", they output predictions corresponding to numerical class IDs (starting at 0). By default, the `label_dict` in the config is populated when the model is pulled from the hub (or a local checkpoint), which is `{ Tree: 0 }`.
 
 However, if you want to train on multiple classes or detect something that isn't a tree (for example we host a model for multiple Everglades bird species), you need to specify:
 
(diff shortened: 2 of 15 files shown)


jveitchmichaelis:
Tests pass locally, but looks like the CI is out of space. Will try to do something about that in another PR - I see this even locally with all the logs that are created.

jveitchmichaelis:
@bw4sz appreciate your input on the additional test cases, especially any train/reload scenarios that I missed here.

henrykironde:
Testing this PR

henrykironde (approved):
This solves the issue of not taking the default label from the model. 
for example: 

```Python
from deepforest import main
import numpy as np
import pandas as pd

model_name =  "weecology/deepforest-bird"
m = main.deepforest()
m.load_model(model_name=model_name)

# Create a white image (uint8 RGB)
white = np.full((2048, 2048, 3), 255, dtype=np.uint8)
m.config.score_thresh = 0.2
m.model.score_thresh = 0.2
res = m.predict_tile(
    image=white,
    patch_size=128,
    patch_overlap=0,
    iou_threshold=0,
)

print(res)
```

Before PR: 
```j
GPU available: False, used: False
TPU available: False, using: 0 TPU cores
Predicting DataLoader 0: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 256/256 [04:06<00:00,  1.04it/s]
256 predictions in overlapping windows, applying non-max suppression
256 predictions kept after non-max suppression
No image path provided, root_dir of the output results dataframe will be None, since either images were directly provided or there were multiple image paths
       xmin    ymin    xmax    ymax label     score                                           geometry
0       4.0     0.0   117.0   123.0  Tree  0.284699      POLYGON ((117 0, 117 123, 4 123, 4 0, 117 0))
1       4.0   128.0   117.0   251.0  Tree  0.284699  POLYGON ((117 128, 117 251, 4 251, 4 128, 117 ...
```

With PR patch:

```j
GPU available: False, used: False
TPU available: False, using: 0 TPU cores
Predicting DataLoader 0: 100%|█████████████████████████████████████████████████| 256/256 [05:05<00:00,  0.84it/s]
256 predictions in overlapping windows, applying non-max suppression
256 predictions kept after non-max suppression
No image path provided, root_dir of the output results dataframe will be None, since either images were directly provided or there were multiple image paths
       xmin    ymin    xmax    ymax label     score                                           geometry
0       4.0     0.0   117.0   123.0  Bird  0.284698      POLYGON ((117 0, 117 123, 4 123, 4 0, 117 0))
1       4.0   128.0   117.0   251.0  Bird  0.284698  POLYGON ((117 128, 117 251, 4 251, 4 128, 117 ...
2       4.0   256.0   117.0   379.0  Bird  0.284698  POLYGON ((117 256, 117 379, 4 379, 4 256, 117 ...
3       4.0   384.0   117.0   507.0  Bird  0.284698  POLYGON ((117 384, 117 507, 4 507, 4 384, 117 ...
```

henrykironde:
For the future, we need to add tests that use non-default models in the config; I’ll open an issue for GSoC contributors. This slipped through because the default and our tests were both using the tree model.

PR #1267, opened 2026-01-14
Add predict_tile profiling script

Rishiiiii097:
Add profiling script for predict_tile

Summary
This PR updates the existing `profile_predict_tile.py` script to improve profiling of
`predict_tile` performance.

Changes
- Modified `tests/profile_predict_tile.py` to add/adjust profiling logic
- No changes to production code or unrelated files

Notes
- Scope is intentionally limited to profiling only

- [ ] I used AI tools (e.g., GitHub Copilot, ChatGPT, etc.) in developing this PR
- [ ] I understand all the code I'm submitting
- [ ] I have reviewed and validated all AI-generated code

### DIFF
--- a/tests/profile_predict_tile.py
+++ b/tests/profile_predict_tile.py
@@ -1,114 +1,101 @@
+# NOTE: Added profiling for predict_tile performance
 import gc
-import glob
 import os
 import time
-
 import numpy as np
-import psutil
-import torch
 from tabulate import tabulate
-
 from deepforest import main
 
+try:
+    import torch
+    import psutil
+except ImportError:
+    pass  # Optional if psutil or torch not available
 
-def get_memory_usage():
-    """Get current memory usage in MB"""
-    process = psutil.Process(os.getpid())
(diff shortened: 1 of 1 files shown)


henrykironde:
@Rishiiiii097, can you paste your output from local testing of your work

Rishiiiii097:
Update after further local testing:

I corrected the predict_tile API usage (using `path` instead of `paths`)
and reran the profiling locally with the existing test image
tests/data/NY.png.

Single strategy now runs successfully and returns predictions
(9 predictions kept after non-max suppression).

Batch strategy fails with:
ValueError: Expected 8-bit 3-channel RGB, got P, 1 channels

This occurs because batch mode uses prediction.MultiImage, which reloads
images from disk and ignores the provided `image` argument, so palette-mode
PNGs cannot be converted to RGB beforehand. This appears to be internal to
the batch prediction path rather than the profiling logic itself.

Please let me know how you’d like me to proceed .

PR #1269, opened 2026-01-15
Replace tmpdir with tmp_path in tests

jveitchmichaelis:
## Description

This PR refactors tests to use `tmp_path` fixtures instead of `tmpdir` as the latter is deprecated by `pytest`. In addition, it should allow us to use more aggressive cleanup on the CI so we don't run out of disk space during testing. Either by default, or through an additional option `-o "tmp_path_retention_policy=none"`

This is _mostly_ a find + replace job, but also allows us to use path syntax instead of os/glob/etc.

See https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures

We could also enforce `-p no:legacypath` to auto-fail the CI if tmpdir is used.

## AI-Assisted Development

A lot of tab auto-complete with Copilot which saved time with replacing strings.

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -54,7 +54,7 @@ jobs:
         run: uv sync --all-extras --dev
 
       - name: Run pytest with coverage
-        run: uv run pytest -v --cov=src/deepforest --cov-report=xml --cov-report=term-missing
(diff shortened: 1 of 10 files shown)


jveitchmichaelis:
@henrykironde would appreciate your input on the refactor here

PR #1271, opened 2026-01-16
[WIP] Update the docs to favor trainer.validate since it more aligns with p…

ethanwhite:
…ytorch lightning and add tests to check parity between main.evaluate and trainer.validate. Simplify the outputs of trainer.validate if this is going to be primary evaluation vehicle, it has too many map products for the average user


bw4sz:
can this be closed?

Issue #1274, opened 2026-01-19
PadIfNeeded augmentation is poorly named. It crops as well as pads.


bw4sz:
<img width="843" height="654" alt="Image" src="https://github.com/user-attachments/assets/ef73c825-559d-4163-b320-2142ab961367" />

If you have an image that is larger than the padding size, it crops the image to that size. The real underlying issue is https://kornia.readthedocs.io/en/latest/augmentation.module.html#kornia.augmentation.PadTo. I think we can write our own augmentation? I think this can be marked good first issue unless @jveitchmichaelis thinks there is some complexity here.

# Desired functionality

An augmentation that pads the image to a given size if it smaller than that size. If it is larger, do nothing.

proofbyhuman:
Hi @bw4sz @ethanwhite 👋

I'd like to work on this issue!

My proposed approach:
1. Update/implement the `PadIfNeeded` augmentation logic in `src/deepforest/augmentations.py` so that it only applies padding when the image dimensions are smaller than the target size, and leaves larger images untouched (preventing unintended cropping).
2. Add comprehensive unit tests in `tests/test_augmentations.py` to verify behavior on both smaller and larger images.

Looking forward to your feedback!

(AI assistance: I used an AI assistant to help analyze the issue and draft this approach.)

PR #1275, opened 2026-01-20
Fix padding ordering issues in augmentation pipeline

jveitchmichaelis:
## Description

Fixes some unexpected behavior during augmentation.

1. If a padding operation is performed before other geometric transforms, the final labels are not transformed correctly. It seems that the geometric transform is correct, though probably not what the user would want, but the labels are definitely wrong. Here for example, if we pad and then flip, the image is correctly vflipped. The labels however are only vflipped in _original_ image coordinates. From what I can see it's not as simple as resizing the labels, this should be handled by `kornia`.

This may be a limitation of how `kornia` constructs/chains the transformation matrices for different operations.

image

3. `kornia` transforms input boxes/points but it doesn't do any filtering afterwards. If the source image is rotated for example, it will may result in boxes going out of bounds. Currently, we clip to the bounds of the augmented image, but if padding is forced to be the last operation, we should clip to the _source_ image and then pad to the desired size. In the other direction, if the image is _cropped_, then we want to bound the labels to the transformed image.

image

This PR makes two changes:

1. If the pipeline contains a `PadTo` operation, move it to the final step. This ensures that the resulting labels are correct. This is opinionated, but until we figure out a way to solve the label transform problem, it's a safe compromise and _should_ always yield correct labels even if the transform pipeline is specified out of order by mistake.
3. Modify box filtering to clamp to the minimum of the source/augmented image width and height.

Typical corrected sample:

image
image

## Related Issue(s)

https://github.com/weecology/deepForest/issues/1274 - related, but this does not change the behavior of PadTo (unexpected cropping).

## AI-Assisted Development

Auto-complete only for updating filter_boxes

### DIFF
--- a/src/deepforest/augmentations.py
+++ b/src/deepforest/augmentations.py
@@ -251,6 +251,14 @@ def get_transform(
             aug_transform = _create_augmentation(aug_name, aug_params)
             transforms_list.append(aug_transform)
 
+    # Enforce ordering to avoid issues with padding and cropping
+    _crop_types = (K.RandomCrop, K.RandomResizedCrop)
+    _pad_types = (K.PadTo, RandomPadTo)
+    standard = [t for t in transforms_list if not isinstance(t, _crop_types + _pad_types)]
+    crops = [t for t in transforms_list if isinstance(t, _crop_types)]
+    pads = [t for t in transforms_list if isinstance(t, _pad_types)]
+    transforms_list = standard + crops + pads
+
     # Create a sequential container for all transforms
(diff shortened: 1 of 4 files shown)


jveitchmichaelis:
Updated this with an additional test and filtering for points + polygons too. I think this is good to go as it's a footgun that's easy to fix.

ethanwhite:
Looks like there's a failing test to resolve on this one

jveitchmichaelis:
@bw4sz I think this is good, we can re-assess if we run into any more edge cases.

bw4sz (approved):
Thanks for fixing this.

PR #1296, opened 2026-02-02
Fix #1274: Rename and fix PadIfNeeded to only pad (no crop)

### LINKED ISSUE #1274: PadIfNeeded augmentation is poorly named. It crops as well as pads.

abhayrajjais01:
### Summary
This PR fixes Issue #1274 where `PadIfNeeded` was incorrectly cropping images larger than the target size because it was aliased directly to `Kornia.PadTo`.

### Changes
*   **New `PadIfNeeded` Class**: Implemented a custom `PadIfNeeded` augmentation (inheriting from `GeometricAugmentationBase2D`) that:
    *   Checks if the input image dimensions are smaller than the target size.
    *   Pads *only* if needed.
    *   Returns the identity (unchanged image) if the image is already larger than the target size.
*   **Added `PadTo`**: Added a new key `"PadTo"` to `_SUPPORTED_TRANSFORMS` that maps to the original `Kornia.PadTo` behavior. This preserves backward compatibility for users who relied on the cropping behavior.
*   **Tests**: Added a new regression test `test_pad_if_needed` in `tests/test_augmentations.py` to verify:
    *   Small images are padded correctly.
    *   Large images are NOT cropped (the fix).
    *   `PadTo` still works as expected (legacy behavior).

### Verification
Ran tests via Docker:
`docker compose run --rm deepforest pytest tests/test_augmentations.py`
Result: All 28 tests passed.

I have taken help of ai to understand the codebase


abhayrajjais01:
@henrykironde can you please review and tell me if changes needed??

PR #1298, opened 2026-02-03
Fix low recall when limit_val_batches is set

### LINKED ISSUE #1232: Evaluation reports spuriously low recall if limit_batches is set
Was doing some test runs with a limited batch size and the results look much better than the box recall suggests (< 0.03, which is low, even for a bad model). Evaluate loads the ground truth CSV file and doesn't use the dataloader, so it will run over the whole val dataset and assume missing predictions for images that were skipped. (I suspect this anyway).

This would probably fix it:

```python
        if self.trainer.limit_val_batches != 1.0 and self.val_dataloader() != []:
            val_loader = self.val_dataloader()
            valid_image_paths = set()
            for batch in val_loader:
                _, _, image_names = batch
                valid_image_paths.update(image_names)
            ground_df = ground_df[ground_df.image_path.isin(valid_image_paths)]
```

but seems wasteful to do a full loop over the dataloader.

This is useful for easily testing the effect of dataset size on model performance, but we would probably want to generate specific subset train and val files to be more rigorous.

vickysharma-prog:
## Description
When `limit_val_batches` is set (e.g., 0.1 for 10%), evaluation loads the full ground truth CSV but predictions only cover the limited images. This makes recall look very low because of "missing" predictions for images that were never processed.

Added a check in `__evaluate__` that trims `ground_df` based on `limit_val_batches` value. Uses `ceil(limit_val_batches * n_images)` as suggested in the issue.

Also added a test case to verify the fix.

## Related Issue(s)
Fixes #1232

## AI-Assisted Development

Used for initial research and understanding the codebase structure

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -10,11 +10,13 @@ import torch
 import tempfile
 import copy
 import importlib.util
+import math
 
 from deepforest import main, get_data, model
 from deepforest.utilities import read_file, format_geometry
 from deepforest.datasets import prediction
 from deepforest.visualize import plot_results
+from deepforest.metrics import RecallPrecision
 
 from pytorch_lightning import Trainer
 from pytorch_lightning.callbacks import Callback
@@ -1167,10 +1169,47 @@ def test_predict_file_mixed_sizes(m, tmp_path):
     preds = m.predict_file(csv_file=csv_path, root_dir=str(tmp_path))
 
     assert preds.ymax.max() > 200  # The larger image should have predictions outside the 200px limit
+
+def test_recall_not_lowered_by_unprocessed_images():
+    """This test checks that recall is only computed for images that were
+    passed to the metric and ignores unprocessed images in the ground truth
+    dataframe."""
+
+    label_dict = {'Tree': 0}
+    metric = RecallPrecision(label_dict=label_dict)
+
+    # Simulate limit_val_batches: only 2 images processed
+    # Using different boxes for each image to catch matching bugs
+    preds = [
+        {'boxes': torch.tensor([[10, 10, 50, 50]], dtype=torch.float32),
+        'labels': torch.tensor([0]), 'scores': torch.tensor([0.9])},
+        {'boxes': torch.tensor([[60, 60, 100, 100]], dtype=torch.float32),
+        'labels': torch.tensor([0]), 'scores': torch.tensor([0.85])}
+    ]
+
+    targets = [
+        {'boxes': torch.tensor([[10, 10, 50, 50]], dtype=torch.float32),
+        'labels': torch.tensor([0])},
+        {'boxes': torch.tensor([[60, 60, 100, 100]], dtype=torch.float32),
+        'labels': torch.tensor([0])}
+    ]
+
+    metric.update(preds, targets, ['img1.jpg', 'img2.jpg'])
+    results = metric.compute()
+
+    # Verify only 2 images were processed (not affected by any external ground truth)
+    assert metric.num_images == 2
+
+    # With perfect matches, recall should be 1.0 (2/2 images matched)
+    assert math.isclose(results['box_recall'], 1.0, rel_tol=1e-5), (
+        f"box_recall={results['box_recall']:.2f}, expected 1.0"
+    )
+
 def test_custom_log_root(m, tmpdir):
     """Test that setting a custom log_root creates logs in the expected location"""
     custom_log_dir = tmpdir.join("custom_logs")
     m.config.log_root = str(custom_log_dir)
+
     m.config.train.fast_dev_run = False
 
     m.create_trainer(limit_train_batches=1, limit_val_batches=1, max_epochs=1)


jveitchmichaelis (changes requested):
Thanks for your contribution, here's some comments:

- Please scope the PR to only the issue (remove .gitignore changes, we could include that in another submission).
- Please test with non-deprecated eval calls. Use m.create_trainer() with limit batches set as an argument and then call m.trainer.validate(m) or m.trainer.fit(m). You may need to set the validation interval to 1. This would better reflect a training scenario.
- As above, this code should work with .validate() - `__evaluate__` is not called during training.
- The code for `main` is a little defensive. The trainer is created on `init` so it's almost impossible to run this function without self.trainer existing.
- The test case does not adequately check the behavior. For example you have asserted non-negative recall, but this does not prove that recall accurately reflects the limited dataframe.
- Please remove the LLM-inserted "fixes" and issue number from the comment in main (L1023), this is unnecessary.
- Is this code correct in multi-GPU environments? I don't think my suggestion in the issue is correct in this case.

It's probably best for this logic to go in the RecallPrecision metric. You can filter ground_df by the image_paths that the metric was called on which is more reliable in multi-GPU.

vickysharma-prog:
Pushed the updates:
- Removed .gitignore changes
- Updated comment in main.py
- Updated test to use `create_trainer()` + `trainer.validate()`
Note: ReadTheDocs build seems to be failing on dependency install (`uv sync --extra docs`) - this appears unrelated to my changes. Let me know if I need to do anything on my end.
Still working on understanding the RecallPrecision metric location for the architectural refactor you suggested.

vickysharma-prog:
Rebased on latest main all checks passing now!
Let me know if there's anything else to address.

vickysharma-prog:
Moved the fix to `RecallPrecision.compute()` in `metrics.py` as suggested - now filtering `ground_df` to only include images that were actually predicted before calling `__evaluate_wrapper__`.
Removed the old fix from `main.py` and updated the test.

vickysharma-prog:
Hi @jveitchmichaelis,
Just a gentle follow-up on this PR.  
I’ve addressed all the requested changes and rebased on the latest main, and CI is passing.
Please let me know if there’s anything else I should update.  
Thanks again for the review!

vickysharma-prog:
@jveitchmichaelis Updated the test to demonstrate the fix!
**Test verification:**
- **Main branch:** `FAILED - box_recall=0.50, expected > 0.7`
- **This PR:** `PASSED - box_recall=1.00`
The test creates 4 images in ground truth but makes predictions for only 2 (simulating `limit_val_batches`).
- Before fix: ground_df has 4 images → recall = 2/4 = 0.50 → test fails
- After fix: ground_df filtered to 2 images → recall = 2/2 = 1.00 → test passes
Ready for review!

tests/test_main.py
```diff
@@ -1131,18 +1131,42 @@ def test_predict_file_mixed_sizes(m, tmp_path):
...
-    """Test that setting a custom log_root creates logs in the expected location"""
-    custom_log_dir = tmpdir.join("custom_logs")
-    m.config.log_root = str(custom_log_dir)
-    m.config.train.fast_dev_run = False
 
-    m.create_trainer(limit_train_batches=1, limit_val_batches=1, max_epochs=1)
-    m.trainer.fit(m)
+def test_recall_not_lowered_by_unprocessed_images(tmp_path):
+    """Regression test for #1232."""
+    import pandas as pd
```
jveitchmichaelis: No inline imports please

tests/test_main.py
```diff
@@ -1131,18 +1131,42 @@ def test_predict_file_mixed_sizes(m, tmp_path):
...
-def test_custom_log_root(m, tmpdir):
-    """Test that setting a custom log_root creates logs in the expected location"""
-    custom_log_dir = tmpdir.join("custom_logs")
-    m.config.log_root = str(custom_log_dir)
-    m.config.train.fast_dev_run = False
 
-    m.create_trainer(limit_train_batches=1, limit_val_batches=1, max_epochs=1)
-    m.trainer.fit(m)
+def test_recall_not_lowered_by_unprocessed_images(tmp_path):
+    """Regression test for #1232."""
```
jveitchmichaelis: Remove references to the issue number throughout.

"This test checks that recall is only computed for images that were passed to the metric and ignores unprocessed images in the ground truth dataframe."

jveitchmichaelis (changes requested):
Thanks for the update, some further changes and please squash your PR to a single commit.

vickysharma-prog:
@jveitchmichaelis Thanks for the detailed guidance throughout this PR!

Done! Changes made:
- Removed .gitignore edits
- Moved imports to top of file  
- Different bounding boxes for img3/img4 to catch edge cases
- Added `len(metric.image_indices) == 2` assertion
- Using `math.isclose()` for recall = 1.0
- Updated docstring per your suggestion
- Squashed to single commit
Ready for review whenever you get a chance.

vickysharma-prog:
@jveitchmichaelis Just a gentle ping - all changes addressed and checks passing. Let me know if anything else needs updating!

vickysharma-prog:
pre-commit.ci autofix

vickysharma-prog:
@jveitchmichaelis All changes addressed, conflicts resolved, and CI passing now. Ready for re-review when you get a chance!

vickysharma-prog:
@jveitchmichaelis All tests are passing. I’ve rebased and squashed the branch into a single commit with all prior feedback addressed.
Summary of changes:
* `metrics.py`: +4 lines — filter `ground_df` to predicted images
* `test_main.py`: +43 lines — added regression test

Please let me know if any further adjustments are needed.

vickysharma-prog:
Hi @jveitchmichaelis , Let me know if any further changes needed for the test

jveitchmichaelis (approved):

PR #1302, opened 2026-02-12
Block incompatible transformers versions

jveitchmichaelis:
This PR adds some version exclusions for transformers to prevent a couple of bugs from causing the CI to fail. These are known to be 4.57.x and 5.1.0, currently. Fixes for underlying issues have been proposed or merged upstream, so alternatively we could wait for the next release cycle.


ethanwhite (approved):
LGTM!

PR #1303, opened 2026-02-12
added test and fix to multiImage using the same indices during batch

bw4sz:
I found a bug in the MultiImage dataset predict_tile strategy. It was using the same index for each batch, leading to incorrect matching with the filename! High priority in the sense that this is a bug, low priority in the sense that a think few users are using the high performance workflow piece. 

For example this passes

dataloader_strategy="single"

```
def test_predict_tile_batch_uses_global_image_indices(m, tmp_path):
    """Batch strategy must assign image_path using global dataset indices, not batch position.
    """
    source = get_data("OSBS_029.png")
    num_images = 5
    paths = []
    for i in range(num_images):
        dest = tmp_path / f"image_{i}.png"
        shutil.copy(source, dest)
        paths.append(str(dest))
    m.config.train.fast_dev_run = False
    m.create_trainer()
    prediction = m.predict_tile(
        path=paths,
        patch_size=300,
        patch_overlap=0,
        dataloader_strategy="single",
    )
    unique_paths = prediction.image_path.unique().tolist()
    assert len(unique_paths) == num_images
    expected_basenames = sorted(os.path.basename(p) for p in paths)
    assert sorted(unique_paths) == expected_basenames
```
but this fails

dataloader_strategy="batch"
```
def test_predict_tile_batch_uses_global_image_indices(m, tmp_path):
    """Batch strategy must assign image_path using global dataset indices, not batch position.
    """
    source = get_data("OSBS_029.png")
    num_images = 5
    paths = []
    for i in range(num_images):
        dest = tmp_path / f"image_{i}.png"
        shutil.copy(source, dest)
        paths.append(str(dest))
    m.config.train.fast_dev_run = False
    m.create_trainer()
    prediction = m.predict_tile(
        path=paths,
        patch_size=300,
        patch_overlap=0,
        dataloader_strategy="batch",
    )
    unique_paths = prediction.image_path.unique().tolist()
    assert len(unique_paths) == num_images
    expected_basenames = sorted(os.path.basename(p) for p in paths)
    assert sorted(unique_paths) == expected_basenames
```

It was a fairly simple fix, you need to keep the global index appended to the list. Cursor and I disagreed what was more pythonic, I did it as an iterator and it recommended a subclass. I don't think it matters much.

# Agentic statement

I found the error and the cause. I asked cursor to create a test to see if it failed. I implemented a fix, which I asked cursor to check. We argued about what fix was more pythonic and I relented to a subclass approach. 

---

> [!NOTE]
> **Medium Risk**
> Changes how `predict_tile(..., dataloader_strategy="batch")` tracks image/window indices across dataloader batches, which can affect how predictions are attributed to image paths. Scope is limited to the batch tiling workflow and is covered by a new regression test.
> 
> **Overview**
> Fixes a bug in `predict_tile`’s `dataloader_strategy="batch"` where predictions could be assigned the wrong `image_path` because batch collation used *batch position* rather than the *global dataset index*.
> 
> `MultiImage` now wraps each image’s crop list in `_IndexedCrops` to carry the dataset index through `collate_fn`, emits `batch_indices` instead of `sublist_lengths`, and `postprocess` uses those indices to format results with the correct image/window metadata. A new test asserts that batching across multiple identical images yields one unique `image_path` per input file.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 55cb4c9eabc6d5e4ad28cdd96738e9c0da679a8d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/deepforest/datasets/prediction.py
+++ b/src/deepforest/datasets/prediction.py
@@ -244,6 +244,15 @@ class FromCSVFile(PredictionDataset):
         return result
 
 
+class _IndexedCrops(list):
+    """List of crops with the dataset index attached for correct batch
+    collation."""
+
+    def __init__(self, index: int, crops: list):
+        super().__init__(crops)
+        self.index = index
+
+
 class MultiImage(PredictionDataset):
     """Take in a list of image paths, preprocess and batch together.
 
@@ -260,7 +269,7 @@ class MultiImage(PredictionDataset):
         self.paths = paths
         self.patch_size = patch_size
         self.patch_overlap = patch_overlap
-        self.sublist_lengths = []
+        self.batch_indices = []
 
         image = self.load_and_preprocess_image(image_path=self.paths[0])
         self.image_height = image.shape[1]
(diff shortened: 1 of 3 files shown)


bw4sz:
pushed, no idea how I even did that, I had the commit the change names, rebased main, somehow managed to push an empty commit, which I didn't even know was possible (@ethanwhite). When you squash and merge it will just be combined with the others. Or I can interactive rebase.

PR #1311, opened 2026-02-17
Add keypoint dataset class

jveitchmichaelis:
## Description

This PR adds a keypoint dataset that returns either (x, y) coordinates or a 2D density mask as target. The dict entry `labels` is consistent with `transformers` semantic segmentation models. This should cover the majority of the models that we might want to train. The structure is essentially the same as BoxDataset.

The diff is better than it looks. I've pulled out some duplicate logic for the box dataset and made base class which will be used for polygons as well. Most of the additions are in the test suite.

Models now have a `task` attribute which is used to load the correct dataset type (e.g. all current models have `task = box`). I think this is cleaner than having the user specify this via config, since it should be obvious by the choice of architecture.

Training is not supported, but I'd like to add that separately to reduce review load.

image

## Related Issue(s)

Supersedes https://github.com/weecology/DeepForest/pull/1182 as features from there are migrated.

## AI-Assisted Development

Some suggest tests and refactoring.

Claude Code, Gemini


bw4sz (approved):
I'm happy with the /src. I always am unsure about the amount of tests. For example, do we need flip tests? I think it's fine and they are pretty cheap to run. I'm going to approve, but something for others to weigh in on as we think about the size of the codebase, which I sense is going to swell a lot over the next year.

jveitchmichaelis:
All good, first rebase had a small conflict

PR #1314, opened 2026-02-19
Expose detections_per_img and topk_candidates in config

### LINKED ISSUE #1309: Feature request: allow users to change retinanet candidate images.
# Expose RetinaNet `detections_per_img` and `topk_candidates` to user config

## Problem

The torchvision `RetinaNet` model has two important parameters that control how many detections are returned per image:

- **`detections_per_img`** (default: `300`) — the maximum number of detections returned after NMS
- **`topk_candidates`** (default: `1000`) — the number of top-scoring candidates considered *before* NMS

Neither of these is currently exposed in the DeepForest config. They are hardcoded at the torchvision defaults. For dense scenes — such as large bird colonies or very dense tree canopies — 300 detections per image may be insufficient, leading to missed objects that the model would otherwise detect.

### Example

While running predictions on dense bird imagery (file `C2_L2_F52_T20230910_134201_350.jpg` from flight `JPG_20230910_133900`), we appear to be hitting the detection cap in very dense regions. The model scores objects above the `score_thresh` but they get silently dropped because they exceed `detections_per_img`.

We also see this in the Bird Fine-Tuning example notebook, which already warns:

> UserWarning: Encountered more than 100 detections in a single image. This means that certain detections with the lowest scores will be ignored...

### Where the limit is applied

In `src/deepforest/models/retinanet.py`, `RetinaNetHub.__init__` passes `**kwargs` through to `torchvision.models.detection.retinanet.RetinaNet`, which accepts `detections_per_img` and `topk_candidates`. However, `Model.create_model()` never passes these parameters, so the torchvision defaults are always used:

```python
# current code in create_model — no detections_per_img or topk_candidates
model = RetinaNetHub(
    backbone_weights="COCO_V1",
    num_classes=self.config.num_classes,
    nms_thresh=self.config.nms_thresh,
    score_thresh=self.config.score_thresh,
    label_dict=label_dict,
)
```

## Proposed solution

1. Add `detections_per_img` (default `300`) and `topk_candidates` (default `1000`) to the DeepForest config schema and `config.yaml`, so users can override them.
2. Pass them through from `Model.create_model()` → `RetinaNetHub()` → `RetinaNet.__init__()`.

## Discussion: Can we just increase the default?

Since we already have `score_thresh` filtering low-confidence detections, one might ask: *why not just set `detections_per_img` very high and rely on `score_thresh` alone?*

Things to consider:

- **Performance**: Higher `detections_per_img` means more boxes survive to NMS, which increases compute cost — though this is likely modest for most use cases.
- **Memory**: More detections means larger output tensors, which could matter on GPU-constrained setups or with very large batch sizes.
- **Interaction with `topk_candidates`**: `topk_candidates` acts as a pre-NMS filter. If `detections_per_img` is raised but `topk_candidates` is left at 1000, the improvement may plateau. Both should be tunable together.
- **Backwards compatibility**: Changing the default could alter existing users' results. Exposing the parameter without changing the default preserves backwards compatibility while giving power users control.

The safest first step is to expose both parameters in the config with their current torchvision defaults.

## Questions

1. @jveitchmichaelis does DETR have similar limitations?
2. What are the downsides, this threshold must exist for a reason, is it slower?

vickysharma-prog:
This PR exposes `detections_per_img` and `topk_candidates` as configurable parameters so users can adjust them for dense scenes.

Currently, these values are hardcoded to the torchvision defaults (300 / 1000). In very dense imagery (e.g., large bird colonies or dense tree canopies), this limit can truncate valid detections.

### Changes
- Added `detections_per_img` and `topk_candidates` to `config.yaml` with current defaults.
- Passed these parameters through `Model.create_model()` → `RetinaNetHub()` → torchvision RetinaNet.
- Added tests in `test_main.py` to verify the parameters are correctly applied to the model.

### Backwards Compatibility
Defaults remain unchanged (300 / 1000), so existing workflows are unaffected.

### Benchmarks
See issue #1309 for details. On CPU, increasing the limit showed only modest overhead (~3–8%). The sample images in the repo did not reach the 300-detection cap, so I was not able to demonstrate detection differences locally. Happy to test further with denser imagery if available.

Fixes #1309

---

- [x] Used AI tools for guidance
- [x] I understand all submitted code
- [x] I reviewed and validated all AI-generated suggestions

### DIFF
--- a/src/deepforest/conf/config.yaml
+++ b/src/deepforest/conf/config.yaml
@@ -11,7 +11,8 @@ batch_size: 1
 architecture: 'retinanet'
 nms_thresh: 0.05
 score_thresh: 0.1
-
+detections_per_img: 300
+topk_candidates: 1000
 # Set model name to None to initialize from scratch
 model:
     name: 'weecology/deepforest-tree'
--- a/src/deepforest/conf/schema.py
+++ b/src/deepforest/conf/schema.py
@@ -144,6 +144,8 @@ class Config:
 
     nms_thresh: float = 0.05
     score_thresh: float = 0.1
+    detections_per_img: int = 300
+    topk_candidates: int = 1000
     model: ModelConfig = field(default_factory=ModelConfig)
 
     log_root: str = "./"
--- a/src/deepforest/models/retinanet.py
+++ b/src/deepforest/models/retinanet.py
@@ -201,6 +201,8 @@ class Model(BaseModel):
                 nms_thresh=self.config.nms_thresh,
                 score_thresh=self.config.score_thresh,
                 label_dict=label_dict,
+                detections_per_img=self.config.detections_per_img,
+                topk_candidates=self.config.topk_candidates,
             )
         else:
             # Pre 2.0 compatibility, the score_threshold used to be stored under retinanet.score_thresh
@@ -216,6 +218,8 @@ class Model(BaseModel):
                 label_dict=label_dict,
                 nms_thresh=self.config.nms_thresh,
                 score_thresh=self.config.score_thresh,
+                detections_per_img=self.config.detections_per_img,
+                topk_candidates=self.config.topk_candidates,
                 **hf_args,
             )
 
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1222,6 +1222,7 @@ def test_custom_log_root(m, tmpdir):
     version_dir = version_dirs[0]
     assert version_dir.join("hparams.yaml").exists(), "hparams.yaml not found"
 
+
 def test_huggingface_model_loads_correct_label_dict():
     """Regression test for #1286:
     HuggingFace models should load correct label_dict from config.json.
@@ -1243,6 +1244,7 @@ def test_huggingface_model_loads_correct_label_dict():
 
     actual = set(m.label_dict.keys())
     assert actual == expected, f"Expected {expected}, got {actual}"
+
 
 def test_existing_dataloader_end_to_end(tmp_path_factory):
     """Regression test for #1369 — verify training and validation
@@ -1283,3 +1285,24 @@ def test_existing_dataloader_end_to_end(tmp_path_factory):
 
     m.trainer.fit(m)
     m.trainer.validate(m)
+
+def test_detections_per_img_and_topk_candidates_config():
+    """Test that detections_per_img and topk_candidates can be configured
+    and are passed through to the underlying model."""
+    m = main.deepforest()
+
+    # Check default values
+    assert m.config.detections_per_img == 300
+    assert m.config.topk_candidates == 1000
+
+    # Test custom values
+    m.config.detections_per_img = 500
+    m.config.topk_candidates = 2000
+
+    assert m.config.detections_per_img == 500
+    assert m.config.topk_candidates == 2000
+
+    # Verify values are passed to actual model
+    m.create_model()
+    assert m.model.detections_per_img == 500
+    assert m.model.topk_candidates == 2000


vickysharma-prog:
@bw4sz Test results with the shared image:

**Setup:**
- Image: 4852 × 6464 pixels
- Model: weecology/deepforest-bird (HuggingFace)
- predict_tile: patch_size=400, patch_overlap=0.25

**Results:**
| detections_per_img | Detections | Time |
|--------------------|------------|------|
| 300 (default)      | 396        | ~29 min |
| 1000               | 396        | ~25 min |

This image shows 719 raw predictions per tile, reduced to 396 after NMS — doesn't appear to hit the per-patch limit with this model.
You mentioned using a different model. Would you like me to test with your checkpoint? That might show the limit being hit.

PR #1314 is ready whenever you'd like to review!

vickysharma-prog:
Rebased and added `detections_per_img` and `topk_candidates` to the config schema — all CI checks passing now.

Testing with the shared image didn't hit the per-tile 300 cap. If you'd like, I can test with your alternate checkpoint to demonstrate the limit in a denser scenario.
Ready for review!

jveitchmichaelis (commented):
The config + passing to the model is fine.

There is a small issue that is likely to pop up in future as we introduce more architectures. This is a specific argument for retinanet. DETR-based models have the same sort of config option, but it is a very different mechanic (changing the number of object queries will require completely retraining because of how the detection head works). And we'll encounter the same with other models I imagine.

I think the naming is confusing for users. How many raw detections does the model produce (which is related to anchors) vs how many outputs we pass to NMS (topk_candidates) and how many we cut after NMS (detections_per_image). This is from retinanet, but we can map it to whatever we like.

For example, in DETR (and family) "num_queries" defaults to 300. We also have an optional NMS stage which seems required for some datasets where the object count varies a lot.

One option would be to expose configs specifically for different architectures, which makes the config a little larger but is at least unambiguous.

@bw4sz any thoughts here?

vickysharma-prog:
@jveitchmichaelis @bw4sz That these are fundamentally different kinds of parameters. detections_per_img and topk_candidates are inference-time filters for RetinaNet topk_candidates controls how many anchor boxes survive to NMS, detections_per_img 
caps the final output after NMS. Neither touches model weights, so users can change them freely without retraining.

DETR's num_queries is a different story it defines the number of learned object queries in the detection head itself. 
Change it and the weight dimensions change, which means retraining from scratch. Putting both under the same flat config namespace would be genuinely confusing for users switching between architectures.

The nested approach makes sense to me:
retinanet:
  detections_per_img: 300
  topk_candidates: 1000

detr:
  num_queries: 300

Implementation would mean a RetinaNetConfig dataclass in schema.py, updating both init paths in retinanet.py to use 
self.config.retinanet.detections_per_img, and adjusting the test accordingly. All straightforward once direction is confirmed.

Happy to implement just waiting on @bw4sz's thoughts before touching anything.

bw4sz:
that's fine with me. per-model config. We only have a couple models at the current model, when we get to AutoModel, that may make this more complicated? I think this is pressing enough lets get this in.

bw4sz:
Okay we discussed and pre-model in the config is fine and we want to get this in and then that will lead us to a larger conversation about config files for each model and a per-model documentation. This is too big of an idea to relate to this small and important PR.

vickysharma-prog:
@jveitchmichaelis rebased onto latest main, all checks passing. Ben confirmed the current config approach is fine for now. Ready for your approval!

bw4sz (approved):
jumping back in here, looks good. @jveitchmichaelis I'm trying to take some weight off of you I feel like you are the reviewer for a lot right now.

PR #1326, opened 2026-02-26
disable PIL DecompressionBombError for large rasters in predict_tile

### LINKED ISSUE #1324: PIL DecompressionBombError when processing large images
The main error where predict_tile fails when attempting to open very large images
PIL's safety mechanism blocks images exceeding a certain pixel size

musaqlain:
When processing large orthomosaics via `predict_tile`, DeepForest triggers a `PIL.Image.DecompressionBombError` (or warning). This occurs because Pillow hardcodes a safety limit of ~89.5 million pixels to prevent DoS attacks, which standard drone imagery effortlessly exceeds.

**Technical Implementation**
* Added `Image.MAX_IMAGE_PIXELS = None` to the module initialization block in `src/deepforest/main.py`.

**AI usage acknowledgement**
- AI is used to pull knowledge from internet only. Fixes #1324.

### DIFF
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -11,6 +11,9 @@ import torch
 from lightning_fabric.utilities.exceptions import MisconfigurationException
 from omegaconf import DictConfig, OmegaConf
 from PIL import Image
+
+Image.MAX_IMAGE_PIXELS = None
+
 from pytorch_lightning.callbacks import LearningRateMonitor
 from torch import optim
 from torchmetrics.classification import BinaryAccuracy


musaqlain:
Thanks again for the feedback, @vickysharma-prog! I want to elaborate on the architectural reasoning behind the module-level approach for the maintainers' consideration.

After evaluating a scoped approach (like a context manager or a per-function toggle), I found a few architectural realities that strongly reinforce the global module-level implementation:

1. **Existing Codebase Precedent:** [`download.py`](https://github.com/weecology/DeepForest/blob/8d9c2dfdbe9379e187ffe82b6cb6367dfa21a7ee/src/deepforest/download.py#L22) already sets `Image.MAX_IMAGE_PIXELS = None` at the module level (line 22). Applying this change in `main.py` follows the established pattern, keeping the codebase consistent rather than introducing a conflicting approach.
2. **The Blast Radius of `Image.open()`:** This function is called across 7+ modules, not just within `predict_tile`. A scoped approach would require injecting boilerplate into:
   * `prediction.py` ([`PredictionDataset.load_and_preprocess_image`](https://github.com/weecology/DeepForest/blob/8d9c2dfdbe9379e187ffe82b6cb6367dfa21a7ee/src/deepforest/datasets/prediction.py#L54))
   * `preprocess.py` ([`split_raster`](https://github.com/weecology/DeepForest/blob/8d9c2dfdbe9379e187ffe82b6cb6367dfa21a7ee/src/deepforest/preprocess.py#L192))
   * `training.py` (`BoxDataset` in both `__init__` and `__getitem__`)
   * `main.py` (`predict_image`)
   * As well as `visualize.py`, `utilities.py`, and `callbacks.py`.

Touching every call site introduces a massive maintenance burden. If even one site is missed, users will hit the exact same crash on a different code path.
3. **Preserving Zero-Config UX:** A configurable flag would break the current zero-config experience. Users processing standard drone orthomosaics (which routinely exceed 100M+ pixels) would be forced to learn and set a new security flag just to use `predict_tile` as documented. This degrades the UX for the most common use case.
4. **Contextual Threat Modeling:** DeepForest is a scientific compute library executed on researcher-provided geospatial rasters in local/HPC/notebook environments. Pillow's decompression bomb guard is designed to protect public web servers accepting untrusted uploads, a threat model that simply doesn't apply here.

Ultimately, this is a one-line, backward-compatible fix that aligns with the [existing approach in `download.py`](https://github.com/weecology/DeepForest/blob/8d9c2dfdbe9379e187ffe82b6cb6367dfa21a7ee/src/deepforest/download.py#L22) .

@henrykironde @bw4sz , I am happy to make any adjustments if you'd prefer a different direction!

bw4sz (approved):
LGTM

PR #1332, opened 2026-03-03
Update loaded GitHub Actions and pin to hash

ethanwhite:
This updates the GitHub Actions we load to the newest release and pins those releases to a specific hash. Pinning to a specific hash is considered a security best practice for avoiding upstream compromises of the actions we use:

https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

This PR eliminates the High threats indicated by zizmor

## AI-Assisted Development

None


henrykironde (approved):
LGTM

PR #1334, opened 2026-03-05
Add spatial-temporal metadata embeddings to CropModel

bw4sz:
## Summary

- Adds optional late-fusion spatial-temporal metadata (lat, lon, date) to CropModel for improved species classification
- Sinusoidal encoding of coordinates + day-of-year through a small MLP (~1.5% of image feature vector), concatenated with ResNet-50 features before the classifier
- Fully backward compatible — metadata is optional at both training and inference time; model gracefully degrades to image-only predictions when metadata is absent

## Key changes

- **`SpatialTemporalEncoder`**: Sinusoidal positional encoding + MLP (6 → configurable dim, default 32)
- **`CropModel`**: Splits into backbone + metadata encoder + classifier when `use_metadata=True`
- **`MetadataImageFolder`**: Wraps ImageFolder with a CSV sidecar (`filename,lat,lon,date`) for training
- **`BoundingBoxDataset`**: Supports per-crop metadata dict for inference
- **`predict_tile`**: Accepts `metadata={"lat": ..., "lon": ..., "date": "YYYY-MM-DD"}`
- **Config**: `use_metadata`, `metadata_dim`, `metadata_dropout`
- **19 new tests** covering encoder, model, datasets, checkpoint save/load, and full training integration
- **Documentation** added to CropModel user guide and configuration reference

## Test plan

- [x] All 19 new metadata tests pass (`pytest tests/test_metadata_cropmodel.py`)
- [x] All 8 existing crop model tests pass (`pytest tests/test_crop_model.py`)
- [x] Full test suite passes on CI
- [x] Manual smoke test with real field data

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Touches core CropModel training/inference data flow and changes batch/forward signatures to support metadata; defaults remain unchanged, but incorrect metadata wiring or shape handling could break crop-model prediction/training in edge cases.
> 
> **Overview**
> Adds optional spatial-temporal metadata (lat/lon/date) support to `CropModel`, including a new `SpatialTemporalEncoder` and a backbone+metadata+classifier path when `cropmodel.use_metadata` is enabled, while keeping image-only behavior as the default.
> 
> Training can now load per-image metadata from a sidecar CSV via `CropModel.load_from_disk(..., metadata_csv=...)`, and inference can pass image-level metadata through `deepforest.predict_tile(..., metadata=...)` which is expanded to per-crop tensors and threaded through `predict`/`BoundingBoxDataset`.
> 
> Updates config/schema/docs to expose `use_metadata`, `metadata_dim`, and `metadata_dropout`, and adds a dedicated test suite covering encoder behavior, training/predict steps, dataset wrappers, and checkpoint save/load.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c763c179b8baa1e57fdde91e3107e1c6e4861df9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


bw4sz:
I changed something small, I don't agree with bugbot.

bw4sz:
@jveitchmichaelis can you skim this. There was a lot of cursor generated code at the beginning and I'm starting to take ownership over it as it seems to be effective and well balanced when I tested it in the BOEM repo. 

image

There are implementation pieces I'm not sure about, and then there are scientific things that merit further review. Like what happens if we see some that is not supposed to occur in a new case, I want to manually change a few test items to see if the metadata overrides the vision model. It's a small portion of the concat features but I will test it. In terms of engineering and our workflows, I am playing with this process of working with the agents to prototype something, check it out on a different project and then assume ownership of the PR for DeepForest. That hand-off is probably awkward. I want to see which things strike you as needing developer time. @ethanwhite just keeping you abreast as a trial pattern. I'm taking it out of WIP, but not triggering review yet?

src/deepforest/model.py
```diff
@@ -104,6 +105,23 @@ def create_crop_backbone(
     return m
 
 
+def create_crop_feature_backbone(
```
bw4sz, line 108: @jveitchmichaelis this is the line i'm most worried about and how it relates to any DETR or generalizing to other models. I am working my way through this branch and this is the only piece that needs your attention.

src/deepforest/model.py
```diff
@@ -403,14 +516,24 @@ def normalize(self):
...
-        if self.model is None:
+    def forward(self, x, metadata=None):
+        if self.backbone is not None:
+            image_features = self.backbone(x)
+            if metadata is not None:
+                meta_features = self.metadata_encoder(metadata)
+            else:
+                meta_dim = self.classifier.in_features - image_features.shape[1]
+                meta_features = torch.zeros(
+                    x.shape[0], meta_dim, device=x.device, dtype=x.dtype
```
bw4sz: @jveitchmichaelis let's discuss the use of x.device here, flagging this as something that worries me with #1391

bw4sz:
Okay, I have taken control over this branch. I am happy with the commits with the exception of the two comments i've flagged. I want to make sure this makes sense if we every want a transformer backbone to the crop classification model (not detection) and whether we learned anything about pytorch lightning handles the .device internally within single/multi-gpu. Other than that, testing on the BOEM pipeline shows this works nicely and I think users will appreciate this as a first step towards greater ecological realism. Getting this merged will allows us to begin to experiment with pretraining and the grain and intensity of the metadata prior.

src/deepforest/model.py
jveitchmichaelis: I think for now I wouldn't worry about it and we can fix it up after.

src/deepforest/model.py
jveitchmichaelis: The device move here could be to the device that `self.backbone` is on, not to `x` (although if `x` is incorrect, we would fail on L519 anyway).

But this comes through `predict_step` which _should_ handle that? So mirroring `.x` should be fine.

src/deepforest/model.py
```diff
@@ -465,20 +586,27 @@ def val_dataloader(self):
         return val_loader
 
     def training_step(self, batch, batch_idx):
-        x, y = batch
-        outputs = self.forward(x)
+        if len(batch) == 3:
+            x, y, metadata = batch
```
jveitchmichaelis, line 590: In future we could probably have this more cleanly as `batch.x`, `batch.y`, `batch.metadata` and have the collation handle it. As opposed to a conditional here and having to set `metadata` anyway.

jveitchmichaelis (commented):
I think this looks OK. There are a couple of style nits that would make the train/predict steps a bit cleaner, but nothing I would necessarily/urgently change here.

If you're happy with the device placement then I think this is good to go.

src/deepforest/model.py
bw4sz: updated to self.backbone.device

src/deepforest/model.py
bw4sz: Let's do this in a separate PR if needed, since it's not related to the behavior of the current collate.

bw4sz:
Picking this back up, once it passes, @jveitchmichaelis merge if you are ready.

docs/user_guide/03_cropmodels.md
```diff
@@ -2,137 +2,214 @@
...
+)
+
+# Train
+crop_model.trainer.fit(crop_model)
+
+# Validate
+crop_model.trainer.validate(crop_model)
+
+# Save checkpoint
+crop_model.trainer.save_checkpoint("model.ckpt")
```
jveitchmichaelis, line 68: Do we have an example for loading? The difference between lightning checkpoint, huggingface model + config in a folder, and a plain torch checkpoint can be confusing for users.

docs/user_guide/03_cropmodels.md
```diff
@@ -2,137 +2,214 @@
...
+## Spatial-Temporal Metadata
 
-You can also pass multiple crop models to `predict_tile`. Each model's predictions and confidence scores will be stored in separate columns.
+In biodiversity monitoring, species distributions vary by location and season. The CropModel supports an optional spatial-temporal metadata embedding that provides location and date context alongside image features to improve classification. The metadata signal is by default "gentle" — it contributes only a small portion of the feature vector. This means the model still classifies primarily from visual appearance but can use location/season as a soft prior. When metadata is not provided at inference time, the model gracefully reverts to image-only classification.
+
+### How It Works
+
+When `use_metadata=True`, the CropModel:
+
+1. Encodes `(lat, lon, day_of_year)`
```
jveitchmichaelis, line 92: Add a bit more detail to how the encoding works (e.g. position/rotation)

docs/user_guide/03_cropmodels.md
```diff
@@ -2,137 +2,214 @@
...
+crop_model = CropModel(config_args={"use_metadata": True})
+crop_model.load_from_disk(train_dir="path/to/train", val_dir="path/to/val",
+                          metadata_csv="metadata.csv")
+crop_model.create_trainer(max_epochs=10)
+crop_model.trainer.fit(crop_model)
+
+result = m.predict_tile(
+    path="image.tif",
+    crop_model=crop_model,
+    metadata={"lat": 35.2, "lon": -120.4, "date": "2024-06-15"}
```
jveitchmichaelis, line 117: Should probably specify somewhere what requirements are for this - are all necessary, only some?

docs/user_guide/03_cropmodels.md
```diff
@@ -2,137 +2,214 @@
...
+```text
+filename,lat,lon,date
+bird_001.png,35.2,-120.4,2024-06-15
+bird_002.png,35.2,-120.4,2024-06-15
+mammal_001.png,40.1,-105.3,2024-07-20
 ```
 
-A `CropModel` is a PyTorch Lightning object and can also be used like any other model.
+- `filename` matches the image basename
+- `date` is an ISO format string
```
jveitchmichaelis, line 135: What format/convetion lat/lon?

docs/user_guide/03_cropmodels.md
```diff
@@ -2,137 +2,214 @@
...
+  --species "Morus bassanus" \
+  --dates 2024-04-15 \
+  --bounds -98 18 -55 48 \
+  --cell-degrees 1.0 \
+  --output-dir outputs/metadata_prior_maps
 ```
 
-### Sampler
+The map below shows a relative metadata prior for Northern Gannet
+(`Morus bassanus`) on April 15, 2024. It reflects the learned metadata branch,
```
jveitchmichaelis, line 198: Can we remove the LLM prose here please (also unclear to me what "It reflects the learned metadata branch, not image evidence" actually means)

docs/user_guide/03_cropmodels.md
```diff
@@ -2,137 +2,214 @@
...
-The `CropModel` makes very few assumptions about the architecture and simply provides a container to make predictions at each detection. To specify a custom CropModel, use the `model` argument.
+## Balance classes during training
 
-```python
-from deepforest.model import CropModel
-from torchvision.models import resnet101
-backbone = resnet101(weights='DEFAULT')
-crop_model = CropModel(num_classes=2, model=backbone)
-```
+Many classification tasks have imbalanced data, meaning that one class appears many more times than others. This leads to the model often choosing this class, regardless of visual appearance. To reduce this effect, a weighted_sampler randomly chooses images to show in training weighted by their inverse frequency. This means that rarer crops are shown more often to offset the common classes. This leads to better performance for rare classes, but can reduce performance on common classes by a small amount. To active the sampler, set the config: cropmodel -> sampler -> 'weighted_random'.
```
jveitchmichaelis, line 210: Does `weighted_sampler` need backticks?

docs/user_guide/03_cropmodels.md
```diff
@@ -444,6 +384,20 @@ class CustomCropModel(CropModel):
         return loss
 ```
 
+## Benefits
+
+Why would you want to apply a model directly to each crop? Why not train a multi-class object detection model?
+While that approach is certainly valid, there are a few key benefits to using CropModels, especially in common use cases:
+
+- **Flexible Labeling**: Object detection models require that all objects of a particular class be annotated within an image, which can be impossible for detailed category labels. For example, you might have bounding boxes for all 'trees' in an image, but only have species or health labels for a small portion of them based on ground surveys. Training a multi-class object detection model would mean training on only a portion of your available data.
+- **Simpler and Extendable**: CropModels decouple detection and classification workflows, allowing separate handling of challenges like class imbalance and incomplete labels, without reducing the quality of the detections. Two-stage object detection models can be finicky with similar classes and often require expertise in managing learning rates.
```
jveitchmichaelis, line 393: + class weighting

docs/user_guide/03_cropmodels.md
```diff
@@ -444,6 +384,20 @@ class CustomCropModel(CropModel):
...
+- **New Data and Multi-sensor Learning**: In many applications, the data needed for detection and classification may differ. The CropModel concept provides an extendable piece that allows for advanced pipelines.
+
```
jveitchmichaelis, line 395: I would also add that training a new cropmodel is much easier (in theory) than training a new detector, for example if your priorities change and you want to add an additional class.

Plus, a cropmodel can be trained on additional data that is not in the detection corpus (which relates to the incomplete labeling argument), such as isolated crops from another dataset.

docs/user_guide/03_cropmodels.md
```diff
@@ -444,6 +384,20 @@ class CustomCropModel(CropModel):
...
+## Considerations
+
+- **Efficiency**: Using a CropModel will be slower, as for each detection, the sensor data needs to be cropped and passed to the detector. This is less efficient than using a combined classification/detection system like multi-class detection models. While modern GPUs mitigate this to some extent, it is still something to be mindful of.
+- **Lack of Spatial Awareness**: The model knows only about the pixels inside the crop and cannot use features outside the bounding box. This lack of spatial awareness can be a major limitation. It is possible, but untested, that multi-class detection models might perform better in such tasks. A box attention mechanism, like in [this paper](https://arxiv.org/abs/2111.13087), could be a better approach. See the {ref}`spatial-temporal-metadata` section for an optional way to incorporate location and season information.
```
jveitchmichaelis, line 399: This is correct, but don't we have the option to increase the visual field through expanded cropping for this reason?

docs/user_guide/09_configuration_file.md
```diff
@@ -332,3 +319,15 @@ crop_model = CropModel()
 # Or use custom resize dimensions
 crop_model = CropModel(config_args={"resize": [300, 300]})
 ```
+
+### use_metadata
+
+Boolean flag to enable spatial-temporal metadata fusion. When `True`, the model accepts `(lat, lon, date)` alongside image crops and learns a small embedding that is concatenated with image features. Default is `False`. See {ref}`spatial-temporal-metadata` for usage details.
```
jveitchmichaelis, line 325: As above, note that these are the only supported fields (possibly that we may add more/options in future).

src/deepforest/datasets/training.py
```diff
@@ -887,3 +879,64 @@ def _classes_in(root):
     )
 
     return train_ds, val_ds
+
+
+class MetadataImageFolder(Dataset):
+    """Wrapper that adds spatial-temporal metadata to an ImageFolder dataset.
+
+    Expects a CSV file with columns: filename, lat, lon, date.
```
jveitchmichaelis, line 887: Same as other comments - need to specify the lat lon format (e.g. decimal degrees). I suppose it doesn't matter as long as the column is consistent.

src/deepforest/datasets/training.py
```diff
@@ -887,3 +879,64 @@ def _classes_in(root):
...
+    The date column should be an ISO format string (e.g., "2024-06-15")
```
jveitchmichaelis, line 888: Would be better to use `fromisoformat` which is more flexible: `https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat`

src/deepforest/datasets/training.py
```diff
@@ -887,3 +879,64 @@ def _classes_in(root):
...
+
+    def __len__(self):
+        return len(self.image_folder)
+
+    def __getitem__(self, idx):
+        image, label = self.image_folder[idx]
+        filepath = self.image_folder.samples[idx][0]
+        filename = os.path.basename(filepath)
+
+        if filename in self._meta_lookup:
```
jveitchmichaelis, line 920: Should the fallback be separate for each field? e.g. if you have lat/lon but no date. 

I suggest storing the metadata (per file) as a `NamedTuple` which is also a bit more defensive against getting the order wrong.

src/deepforest/model.py
```diff
@@ -210,21 +291,44 @@ def create_model(self, num_classes: int, architecture: str | None = None):
             }
         )
 
-        self.model = create_crop_backbone(
-            architecture=architecture,
-            num_classes=num_classes,
-        )
+        use_metadata = self.config["cropmodel"].get("use_metadata", False)
+
+        if use_metadata:
```
jveitchmichaelis, line 296: It's probably transparent to the user, but feels awkward to have backbone/model/classifier and then have metadata models not have `model` defined and vice versa. I am OK with this for now, but I would consider sub-classing the `CropModel` to avoid having conditional statements everywhere.

tests/test_metadata_cropmodel.py
```diff
@@ -0,0 +1,76 @@
+"""Tests for spatial-temporal metadata embeddings in CropModel."""
```
jveitchmichaelis, line 1: Probably need a few tests that loading data with metadata columns are parsed correctly + we get sensible embeddings out (generally, add tests for the modified dataset class)

jveitchmichaelis (changes requested):
I had a re-read:

- Some comments about docs where things could be expanded upon / also docstrings
- A few comments about the architecture of the changes. The modeling looks fine, but it's getting a bit messy with conditions everywhere. I would just make a separate subclass or a wrapper as we do with the dataset. Let's look at that later.
- Needs some tests for the dataset + check that embeddings are computed in range etc?

jveitchmichaelis:
(also needs squash)

PR #1343, opened 2026-03-07
refactor recall_precision metric for better DDP support

### LINKED ISSUE #1073: box_recall and box_precision don't match inside training_loop and outside
I have not been able to recreate this issue locally yet, but I want to document in case a user sees it. I believe this will only occur on the pre2.0 tag on the git repo and not on any release. 

In https://www.comet.com/bw4sz/boem/f3c687c8824f4b9fafa4bafbbe3fad8b?compareXAxis=step&experiment-tab=panels&prevPath=%2Fbw4sz%2Fboem%2Fview%2Fnew%2Fexperiments&showOutliers=true&smoothing=0&xAxis=step
the box_recall and precision is 0.8 and 0.15 on the logged metrics, but afterwards when running main.evaluate on the same code you get the expect high score of 0.91 and 0.92. A couple posibilities

This likely relates to the on_validation_epoch_end hook which has been a consistant challenge because of the inability to use trainer.predict inside a hook that is called during trainer.fit, which is well documented within pytorch lightning. 

# Potential ideas.

1. The metrics aren't being logged currently
2. The predictions aren't being formatted correctly.
3. The metrics and predictions are being passed correctly, but the model isn't being called correctly, batch_norm and any batch statistics.

### LINKED ISSUE #1232: Evaluation reports spuriously low recall if limit_batches is set

jveitchmichaelis:
## Description

- Refactors the RecallPrecision metric to use (mostly) the same `update` signature as other metrics.
- Ground truth/pred matching is performed eagerly during `update` so that the final reduction stage doesn't time-out the sync between ranks during DDP joining.
- This should be faster on multi node machines, because each rank will only perform 1/N of the work. There are maybe some tiny benefits from not having to search the dataframe for image_path <> preds/gt. The `it/s` during eval might be a bit slower, but there won't be a big chunk of compute at the end.
- The metric has to keep track of the matching results so it can be accessed later, which is a little awkward, but I've tried to keep it clean (at least to users...)
- I moved empty frame accuracy here too, since we have to handle those checks during evaluation anyway and it make main a bit cleaner.
- The metrics are calculated against the dataset targets directly (instead of loading the CSV and passing that raw). This has the benefit that adding geometric augmentation to the validation dataset (if we did that for some reason) wouldn't break eval.

Tests:
- Added a sanity check in `test_main::test_evaluate` to confirm that the output dataframe is what we expect + some comments on the assertions.
- One fixup in tests for the new metric.update signature.

## Related Issue(s)

- Fixes https://github.com/weecology/DeepForest/issues/1073
- Should also fix https://github.com/weecology/DeepForest/issues/1232

## AI-Assisted Development

Rewrote most of the metric and then got CC to check some of the logic and the test suite.

Claude Code Opus/Sonnet 4.6

### DIFF
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -12,7 +12,6 @@ from PIL import Image
 from pytorch_lightning.callbacks import LearningRateMonitor
 from pytorch_lightning.loggers import CSVLogger
 from torch import optim
-from torchmetrics.classification import BinaryAccuracy
 from torchmetrics.detection import IntersectionOverUnion, MeanAveragePrecision
 
 from deepforest import predict, utilities
@@ -99,11 +98,8 @@ class deepforest(pl.LightningModule):
         )
         self.mAP_metric = MeanAveragePrecision(backend="faster_coco_eval")
 
-        # Empty frame accuracy
-        self.empty_frame_accuracy = BinaryAccuracy()
-
         self.precision_recall_metric = RecallPrecision(
-            csv_file=self.config.validation.csv_file,
+            iou_threshold=self.config.validation.iou_threshold,
             label_dict=self.label_dict,
         )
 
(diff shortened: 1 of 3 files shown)


bw4sz (commented):
This looks right, i'm going to check it out and submit a BOEM detection multi-gpu that was previous failing and see.

bw4sz (approved):
Agreed. I'm glad you checked out a different branch. I'm not surprised there is something specific about the bird detector.

PR #1352, opened 2026-03-14
Ensure pre-commit tests by adding actions 

henrykironde:
Add GitHub Actions to enforce pre-commit tests alongside existing app tests

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -31,6 +31,8 @@ jobs:
     steps:
       - name: "Checkout code"
         uses: "actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd"
+        with:
+          fetch-depth: 0
 
       - name: Install opencv dependencies
         run: |
@@ -52,6 +54,11 @@ jobs:
 
       - name: Install package
         run: uv sync --all-extras --dev
+
+      - name: Run pre-commit (diff vs origin/main)
+        run: |
+          git fetch origin main --prune
+          uv run pre-commit run --from-ref origin/main --to-ref HEAD
 
       - name: Run pytest with coverage
         run: uv run pytest -v --cov=src/deepforest --cov-report=xml --cov-report=term-missing -o tmp_path_retention=none tests/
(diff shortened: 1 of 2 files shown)


vickysharma-prog:
@henrykironde Following up on your tiled raster check suggestion — I've looked deeper into the implementation:

The current global `MAX_IMAGE_PIXELS = None` (#1326) works but disables PIL's protection entirely. Your original idea of a pre-flight check is cleaner and safer:

```python
import rasterio
import warnings

# PIL default: int(1024 * 1024 * 1024 // 4 // 3) = 89,478,485
PIL_PIXEL_LIMIT = 89_478_485

def _check_large_image(raster_path):
    """Pre-flight check: detect large images and route to optimal strategy."""
    try:
        with rasterio.open(raster_path) as src:
            total_pixels = src.width * src.height
            is_tiled = src.profile.get('tiled', False)
            
            if total_pixels > PIL_PIXEL_LIMIT:
                if is_tiled:
                    warnings.warn(
                        f"Large tiled raster ({total_pixels:,} pixels). "
                        f"Auto-switching to 'window' strategy."
                    )
                    return "window"
                else:
                    raise ValueError(
                        f"Image ({total_pixels:,} pixels) exceeds PIL limit "
                        f"and is not tiled. Convert with:\n"
                        f"  gdal_translate -of GTiff -co TILED=YES input.tif output.tif"
                    )
    except rasterio.errors.RasterioIOError:
        # Non-raster format (PNG/JPG) - continue with default strategy
        pass
    
    return None
```

**Edge cases I've identified:**

Non-raster fallback: rasterio.open() fails on PNG/JPG — needs graceful handling (not just crash)
predict_tile dual input: Accepts both raster_path (file) and image (numpy array) — check only applies to file paths
Consistency: download.py (line 22) also sets MAX_IMAGE_PIXELS = None — both should use the same approach
Dual PIL thresholds: Warning at ~89.5M pixels, hard error at ~178M — should we handle these differently?
This would be placed at the start of predict_tile, before any dataset creation — zero memory overhead since rasterio.open() only reads metadata.

I'd like to implement this as a follow-up PR to #1326 with tests. Happy to get started if this direction works!

PR #1356, opened 2026-03-19
add mixed precision training/matmul32 precision

jveitchmichaelis:
## Description

Adds config options to enabled mixed precision training via Lightning, and allow setting matmul32 precision on CUDA systems. This suppresses a warning from `torch` when CUDA cores are available, and also provides a decent speedup.

The defaults are `auto` equivalents to avoid any issues with existing models, but I've seen pretty good speedups using `16-mixed/high`

## AI-Assisted Development

Checks to make sure configs are consistent.

### DIFF
--- a/src/deepforest/conf/config.yaml
+++ b/src/deepforest/conf/config.yaml
@@ -17,6 +17,13 @@ model:
     name: 'weecology/deepforest-tree'
     revision: 'main'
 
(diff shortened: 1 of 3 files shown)


bw4sz (approved):
This looks good. At some point we need to revisit the docs, we have all these PyTorch lightning trainer args that are basically passthroughs to the trainer.

PR #1304, opened 2026-02-14
[WIP]: Improve predict_tile memory use and windowed raster inference 

henrykironde:
## AI-Assisted Development

- [x] I used AI tools (e.g., GitHub Copilot, ChatGPT, etc.) in developing this PR
- [x] I understand all the code I'm submitting
- [x] I have reviewed and validated all AI-generated code

**AI tools used (if applicable):**

---

> [!NOTE]
> **Medium Risk**
> Touches core inference data loading/normalization and raster IO lifecycle, which can affect prediction correctness and performance on large tiles. Added guards/tests reduce risk but edge cases around raster windowing and multiprocessing still warrant review.
> 
> **Overview**
> Improves `predict_tile`/prediction datasets by making image preprocessing stricter and more consistent (shared RGB/CHW + dtype/range normalization) and preventing crop views from being mutated in-place (`SingleImage.get_crop()` now clones).
> 
> Hardens windowed raster inference (`TiledRaster`) by keeping a single raster handle open with explicit `close()`/pickle-safe reopen, warning (not failing) on non-tiled rasters, filtering/guarding against out-of-bounds/zero-size windows, and ensuring the handle is always closed via a `try/finally` in `predict_tile`.
> 
> Adds a warning when CUDA is available/requested but the Lightning `Trainer` isn’t actually using a GPU accelerator, plus new GPU/HPC-only regression tests for CUDA usage and multi-GPU DDP training; also bumps supported Python to `<3.14`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d847701d5c380e8d785bd63beda8a4accec9ad7c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #1367, opened 2026-04-03
Multi server multi gpu

henrykironde:
## AI-Assisted Development


bw4sz:
Can you confirm what SLURM script you used to check this so i can match that?

bw4sz (commented):
I want to approve this, @jveitchmichaelis any objections? I have spoken to comet and they agree that the lack of multi-node GPU utilization graph is probably on their end and not anything wrong here.

bw4sz:
remove references to torchrun, we can use srun alone.

bw4sz:
To do, is @henrykironde comparing this with #1304 and decide if both are needing, but we want to get this done because it's broad and could make rebasing harder.

src/deepforest/distributed.py
```diff
@@ -0,0 +1,90 @@
...
+    if trainer is not None:
+        return bool(getattr(trainer, "is_global_zero", True))
+
+    if not is_distributed():
+        return True
+
+    return dist.get_rank() == 0
+
+
+def should_sync(trainer: Any | None = None) -> bool:
```
jveitchmichaelis, line 29: Is this required? I thoughtorchmetrics handles syncing?

henrykironde: We do not need this for tensor metrics. TorchMetrics artifacts already sync those across ranks. For non-tensor results (pandas DataFrames), we handle distribution explicitl. Without gathering predictions across ranks, each GPU keeps its own DataFrame, which led to duplicated or inconsistent outputs.

ethanwhite: @jveitchmichaelis - can you take a look at this response and either merge if @henrykironde's response clears things up or the two of you get together to figure it out so we can get this one merged. Thanks!

bw4sz (approved):
This looks good, related to the question of when and how we should launch multi-gpu tests to verify future releases.

PR #1370, opened 2026-04-12
Fix erroneous checks when using existing dataloader 

### LINKED ISSUE #1369: Erroneous checks when using existing dataloader
---

## Issue 1: `on_fit_start` raises when `existing_train_dataloader` is set

**Title:** `on_fit_start` raises `AttributeError` when using `existing_train_dataloader`

**Body:**

When passing a custom dataloader via `existing_train_dataloader=` the guard in `on_fit_start` still raises:

```python
def on_fit_start(self):
    if self.config["train"]["csv_file"] is None:
        raise AttributeError("Cannot train with a train annotations file ...")
```

`config["train"]["csv_file"]` is `None` (correct — there is no CSV), but training should proceed because `existing_train_dataloader` is set. The check should be:

```python
if self.config["train"]["csv_file"] is None and self.existing_train_dataloader is None:
```

**Suggested fix:** change the condition in `on_fit_start` to also allow training when `existing_train_dataloader is not None`.

---

## Issue 2: `create_trainer` silently disables validation when `existing_val_dataloader` is set

**Title:** `create_trainer` disables validation even when `existing_val_dataloader` is provided

**Body:**

`create_trainer` decides whether to enable validation by checking only `config["validation"]["csv_file"]`:

```python
if not self.config["validation"]["csv_file"] is None:
    limit_val_batches = 1.0
    num_sanity_val_steps = 2
else:
    limit_val_batches = 0   # validation silently disabled
    num_sanity_val_steps = 0
```

If a user sets `existing_val_dataloader=` but leaves `config["validation"]["csv_file"]` as `None`, validation is silently disabled — no warning, no error, val metrics never appear. The condition should also check `self.existing_val_dataloader`:

```python
has_val = (self.config["validation"]["csv_file"] is not None
           or self.existing_val_dataloader is not None)
if has_val:
    limit_val_batches = 1.0
    num_sanity_val_steps = 2
else:
    limit_val_batches = 0
    num_sanity_val_steps = 0
```

vickysharma-prog:
## Summary

Adds a regression test for #1369 to ensure training and validation work correctly when `existing_train_dataloader` or 
`existing_val_dataloader` is passed instead of a CSV file.

---
## Background

Issue #1369 reported two silent bugs when using existing dataloaders:

**Bug 1:** `on_fit_start` raised `AttributeError` when `existing_train_dataloader` was set but `csv_file` was `None` 
even though training should proceed via the existing dataloader.

**Bug 2:** `create_trainer` silently disabled validation when `existing_val_dataloader` was set but `csv_file` was `None` 
no warning, no error, val metrics never appeared.

Both bugs were caused by guards checking only the `csv_file` config key, without considering the existing dataloader alternatives.

This PR fixes both guards and adds regression tests to prevent future regressions.

---

## Changes

**Fix 1 — `on_fit_start`:**  
Raise only when **both** `csv_file` and 
`existing_train_dataloader` are missing.

**Fix 2 — `create_trainer`:**  
Enable validation when **either** `csv_file` or 
`existing_val_dataloader` is present.

---

## Test Coverage

| Test | Covers |
|------|--------|
| `test_on_fit_start_existing_train_dataloader` | Bug 1 — no raise when existing train dataloader set |
| `test_create_trainer_existing_val_dataloader` | Bug 2 — validation enabled when existing val dataloader set |

No custom dataloader tests existed before these fill that gap.

---

## Backwards Compatibility

- [x] `csv_file` based training — unchanged
- [x] `csv_file` based validation — unchanged
- [x] Only new paths added for existing dataloader support
---
## Related Issue(s)

Closes #1369

---
## AI-Assisted Development

AI tools used for initial research and understanding the codebase better. All code was written and verified manually.


bw4sz:
I think the test here needs to genuinely build a dataloader, run one pass of train, one pass of validate to demonstrate that the workflow actually passes. I think mocking this just hides the error.

vickysharma-prog:
Updated the PR based on your feedback.

Replaced the mock-based tests with a single end-to-end test that genuinely builds real dataloaders, runs one train pass and one validation pass via fast_dev_run=True.

Also fixed setup_metrics which was returning early when csv_file=None, preventing precision_recall_metric from initializing when existing_val_dataloader was set. This was the root cause of the on_validation_epoch_end crash you reported.

Three fixes total:
- on_fit_start
- create_trainer  
- setup_metrics

All three are covered passes locally.

bw4sz:
Can you assert that the label_dict is correctly created, so that

```
self.label_dict["Tree"] = 0
```
I think by setting fast_dev_run you are turning off pieces that could fail, might be safer to do limit_train_bathes=1 or limit_val_batches=1 and run both trainer.fit() and trainer.validate(). That might run into #1298 but I think should be fine.

vickysharma-prog:
Thanks! Replacing fast_dev_run with limit_train_batches=1 + limit_val_batches=1, adding trainer.validate() and asserting 
label_dict will push shortly.

bw4sz (approved):
Looks good. Thank you.

PR #1372, opened 2026-04-14
Add support for Adam(W) and weight decay 

jveitchmichaelis:
## Description

Adds support for different optimizers in the config. Shouldn't break anything backwards as this was never exposed in older versions of DeepForest.

Fixes a small bug/footgun in the scheduler config where if you specified something not in the list, it would silently default to  ReduceLROnPlateau.

Adam/w is widely used for training transformer-based models, and also supports a weight_decay parameter. 

## Related Issue(s)

In support of newer models being added to DeepForest.

## AI-Assisted Development

Claude

### DIFF
--- a/src/deepforest/conf/config.yaml
+++ b/src/deepforest/conf/config.yaml
@@ -46,6 +46,11 @@ train:
 
     # Optimizer initial learning rate
     lr: 0.001
+    optimizer:
+        type: SGD
+        weight_decay: 0.0
+        momentum: 0.9
+        betas: [0.9, 0.999]
 
     # Data augmentations for training
     # Augmentations must be a list of augmentation names, or a list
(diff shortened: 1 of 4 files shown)


bw4sz:
Yes, I've meant to do this. I sometimes wonder if there isn't a better kwargs type way of getting at pytorch-lightning, I don't love this pattern where we just wrap all things into the config. But I guess it does meet the style of trying to set reasonable defaults for users?

bw4sz (approved):
I'm approving this and we can discuss the larger pattern at Thursday meeting.

jveitchmichaelis:
Anything that's a trainer arg specifically can probably be a config group e.g. `+trainer.strategy= ... ` and then we get CLI support for limit_batches and other things.

(Ie none of these options would be in the config, you'd pass a dict that gets added onto whatever is passed to the trainer, sort of like a **kwargs)

PR #1285, opened 2026-01-25
optimize coordinate validation, handling both -ve and OOB error

musaqlain:
Achieved an **~8.2x speedup** (9.5 min -> 1.1 min) in dataset validation by reducing disk I/O and vectorizing coordinate checks.

**Changes**

* **Optimized I/O:** Caches image dimensions to avoid repeated file opens.
* **Vectorization:** Replaced `iterrows` with Pandas boolean masking.
* **Unified Error Reporting:** Merged checks for negative/OOB coordinates into a single report that lists all invalid boxes to aid debugging.
* **Polygon Support:** Automatically converts non-rectangular geometries to valid bounding boxes.

Fixes #1244

### AI-Assisted Development

- [x] I used AI tools (e.g., GitHub Copilot, ChatGPT, etc.) in developing this PR
- [x] I understand all the code I'm submitting, thanks

PR #1373, opened 2026-04-15
Speed up validation for datasets with same-size images

jveitchmichaelis:
## Description

Most of the time we train on datasets where images are the same size. We spend a lot of time during validation opening images to check what size they are, so we can confirm that annotations are in bounds. A simple optimization if we know up-front that the dataset doesn't have varying sized images, is to take the size of the first one and assume it's correct for the rest of the dataset.

This defaults to `False`, which is the brute force approach, but in cases where we know the dataset is good, or the images are the same size, this can save quite a bit of time and disk thrashing.

Also cleaned up the keypoint validation checker to have similar structure.

## AI-Assisted Development

Claude

### DIFF
--- a/src/deepforest/conf/config.yaml
+++ b/src/deepforest/conf/config.yaml
@@ -83,12 +83,15 @@ train:
     fast_dev_run: False
     # preload images to GPU memory for fast training. This depends on GPU size and number of images.
     preload_images: False
+    # Skip per-image dimension validation when all images share the same size.
+    same_size_images: False
 
 
 validation:
     csv_file:
     root_dir:
     preload_images: False
+    same_size_images: False
     size:
 
     # For retinanet you may prefer val_classification, but the default val_loss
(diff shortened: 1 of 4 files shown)


bw4sz:
I wonder if instead of this we should just make validate_coordinates() something a user can call from DeepForest.main, it seems really heavy every time that we run we open every image. I think the underlying idea here is wrong and while this fixes it, we should probably allow models to fail and then have docs to check validate coordinates to help users find that error?

jveitchmichaelis:
Or have a CLI script which runs the sanity checks? But yes exactly, the current behaviour is to open every image every time which adds minutes to large training runs.

ethanwhite:
We originally added this code because we were getting a lot of error reports that were from users with buggy annotations, which is tricky to debug.

Related to #1285, which is probably close to ready and where I think @bw4sz is saying that we should keep the checks in place?

bw4sz:
closed in favor of a solution to issue #1374

PR #1375, opened 2026-04-17
allow users to disable validate_coordinates

### LINKED ISSUE #1374: Allow option to turn off validate_coordinates at scale
Given the large datasets, we want to allow `validate_coordinates`: True in config, so we can turn it off at scale.

musaqlain:
Resolves #1374 

Question: Should I keep these tests or are these overhead?

#Ai usage
-I understand all the code I'm submitting
-I have reviewed and validated all AI-generated code

#Tool
Gemini

### DIFF
--- a/src/deepforest/conf/config.yaml
+++ b/src/deepforest/conf/config.yaml
@@ -89,12 +89,15 @@ train:
     fast_dev_run: False
     # preload images to GPU memory for fast training. This depends on GPU size and number of images.
     preload_images: False
+    # Validate that annotation coordinates fall within image bounds before training.
+    validate_coordinates: True
 
 
 validation:
     csv_file:
     root_dir:
     preload_images: False
+    validate_coordinates: True
     size:
 
     # For retinanet you may prefer val_classification, but the default val_loss
--- a/src/deepforest/conf/config.yaml
+++ b/src/deepforest/conf/config.yaml
@@ -88,12 +88,15 @@ train:
     fast_dev_run: False
     # preload images to GPU memory for fast training. This depends on GPU size and number of images.
     preload_images: False
+    # Validate that annotation coordinates fall within image bounds before training.
+    validate_coordinates: True
 
 
 validation:
     csv_file:
     root_dir:
     preload_images: False
+    validate_coordinates: True
     size:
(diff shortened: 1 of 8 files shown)


musaqlain:
there is no formatting issue in code, I think this 'pre-commit' error can be resolved by rerunning it----as it is saying `error during build`

vickysharma-prog:
Try commenting pre-commit.ci run. This will re-trigger the checks.

ethanwhite:
@musaqlain can you rebase this on `main` so that precommit will pass and then we'll try to get it reviewed

musaqlain:
hi, now the precommit are passed. ready for review.thanks @ethanwhite

ethanwhite (commented):
Thanks @musaqlain. This looks good overall. My one question is about the implementation across the three geometries. In `training.py` you've explicitly added it to the `PointDataset` class but not to `BoxDataset` or `PolygonDataset`. In `main.py` you've added it in the box and polygon cases, but not the point case. And then there are tests for boxes and points. So, I'm both confused by the mix of code changes that depend on geometry type and about why the new tests are passing for points when that change isn't implemented for that geometry in `main.py`. Can you explain these differences for me?

musaqlain:
Thanks for the thorough review @ethanwhite! You're right, the implementation was inconsistent across geometries, it oversight from me. Here's the explanation and fix:

1. Why the mix: The `validate_coordinates` flag and logic live in the base class `TrainingDataset.__init__`. `BoxDataset` inherits it directly, but `PointDataset` and `PolygonDataset` override `__init__`, so they need to explicitly accept and forward the parameter.

2. Why the point tests passed: The tests call `PointDataset(validate_coordinates=False)` directly, not through `main.py`. So the constructor path worked, but the `main.py → load_dataset → PointDataset` path wasn't forwarding the flag, which you correctly caught.

3. Fixes pushed:
  - Added `validate_coordinates` to the `PointDataset` call in `main.py` (that was actually missing :-)).
- Added a unit test `test_polygon_dataset_validate_coordinates_disabled`

ethanwhite (approved):
Thanks @musaqlain!

Issue #336, opened 2022-08-19
ImportError: cannot import name 'main' from 'deepforest'


ThomasHaarlem:
**Describe the bug**
```
Traceback (most recent call last):

  File "C:\Users\Thomas\anaconda3\envs\deepf\lib\site-packages\spyder_kernels\py3compat.py", line 356, in compat_exec
    exec(code, globals, locals)

  File "c:\gisdb\python\pydeepf.py", line 10, in <module>
    from deepforest import main

ImportError: cannot import name 'main' from 'deepforest' (C:\Users\Thomas\anaconda3\envs\deepf\lib\site-packages\deepforest\__init__.py)
```

**To Reproduce**
I've a clean environment 'deepf' in Anaconda for the DeepForest package. I installed Spyder 5.3.0 to run the code in the section ["getting started](https://deepforest.readthedocs.io/en/latest/getting_started.html)" from read the docs.

```
from deepforest import main
from deepforest import get_data
import os
import matplotlib.pyplot as plt
model = main.deepforest()
model.use_release()

img = model.predict_image(path="/Users/benweinstein/Documents/NeonTreeEvaluation/evaluation/RGB/TEAK_049_2019.tif",return_plot=True)

#predict_image returns plot in BlueGreenRed (opencv style), but matplotlib likes RedGreenBlue, switch the channel order.
plt.imshow(img[:,:,::-1])

```

**Environment (please complete the following information):**
 - OS: Windows 10 Home 64-bit
 - Python version and environment : 3.7, Anaconda1.10.0, Spyder 5.3.0 

**Screenshots**

**Additional context**
* In Anaconda Navigator I added the channel conda-forge in my fresh environment 'deepf' to install DeepForest.
* when I search for "ImportError: cannot import name 'main' from 'deepforest'" I only see results about errors after upgrading PIP. I did not use PIP to install DeepForest.

* I start the script with ```help(modules)``` and I see ```deepforest``` in the list.
* This is the first time I install a task-specific module like DeepForest. So maybe I forgot to change something user-specific in a config-file? Or another really easy step in the process which is obvious for more experienced users? If this is the case, maybe I should have asked for a feature for more installation-guiding. My apologies.

[edit]
* when I carefully follow the path in the error message it says:
```C:\Users\Thomas\anaconda3\envs\deepf\lib\site-packages\deepforest\__init__.py```
[...]

Issue #335, opened 2022-08-18
Loss function is not computed.


luforestal:
Hello,

I am trying to train with RGB images of different size. However, since epoch 0 the loss function is not being computed. I only have a NaN from the middle of the iterations at best. 

```
annotations_file = ('train.csv')
validation_file =  ('val.csv')
m = main.deepforest()
callback = ModelCheckpoint(dirpath='temp/dir',
                                 monitor='box_recall', 
                                 mode="max",
                                 save_top_k=1,
                                 filename="box_recall-{epoch:02d}-{box_recall:.2f}")

m.config['gpus'] = '0'
m.config["train"]["csv_file"] = annotations_file
m.config["train"]["root_dir"] = 'train'
m.config["score_thresh"] = 0.4
m.config["train"]['epochs'] =  1
m.config["validation"]["csv_file"] = validation_file
m.config["validation"]["root_dir"] = 'val'
m.create_trainer(logger=TensorBoardLogger(save_dir='logdir/'), 
                                  callbacks=[callback])

m.use_release()
start_time = time.time()
m.trainer.fit(m)

save_dir = os.path.join(os.getcwd(),'pred_result')
try:
    os.mkdir(save_dir)
except FileExistsError:
    pass
results = m.evaluate(annotations_file, annotations_file, iou_threshold = 0.4, savedir= save_dir)
print(f"--- Training on CPU: {(time.time() - start_time):.2f} seconds ---")
```

![image](https://user-images.githubusercontent.com/76424439/185345709-9b7a055c-b6df-4811-855c-f2a7c664101b.png)

 - Python version and environment : Python 3.8.10 

Also, I have not been able to save any good checkpoints:

m.trainer.save_checkpoint('last_checkpoint.pl')

bw4sz:
Just as a sanity check, can you share your annotations file, I might have a malformed box or class. I'd be interested in exploring this more, retinanet should be formatting the sizes to a standard on entry, and I have an open issue https://github.com/weecology/DeepForest/issues/347 to allow users more control over this, but in principle, torchvision should allow you to batch images of different sizes. Can you confirm whether it works with images of the same size? What use case is this where there are images of many sizes?

ethanwhite:
Sorry we lost track of this @luforestal. If you're still looking for help can you confirm that this issue is still happening on the current release and then we can take another look.

PR #1378, opened 2026-04-19
fixed pre-commit docformatter error

### LINKED ISSUE #1266: Add Python 3.14
docformatter currently fails on Python 3.14 due to its dependency on untokenize. This issue is already tracked upstream.
A future docformatter release is expected to resolve the problem. Once that release is available, we can proceed with updating our dependency.

Reference: https://github.com/PyCQA/docformatter/issues/327

musaqlain:
### What

Fixes the `docformatter` pre-commit hook and pins `language_version: python3.12`
for hooks affected by Python 3.14 AST removals on pre-commit.ci.

### Why

**docformatter (`InvalidManifestError`)**
- The previous `repo: local` definition was replaced with the official repo URL
- `v1.7.5` contains a `docformatter-venv` hook using the deprecated `language: python_venv`,
  which is rejected by pre-commit v4+, even if only the `docformatter` hook is used,
  pre-commit validates the **entire** manifest
- `v1.7.7` removes the problematic hook from its manifest
- However, `v1.7.7` still depends on `untokenize`, which uses `ast.Constant.s`,
  removed in Python 3.14, causing a build failure on pre-commit.ci
- Ref: PyCQA/docformatter#327, PyCQA/docformatter#340

**nbQA (`ast.Str` removal)**
- In Python 3.14, `ast.Str` was removed
- nbQA's notebook processing depends on `ast.Str`, so it fails on Python 3.14
- On `main`, this was hidden by pre-commit.ci's **cached environments**, the nbQA
  environment was built on an older Python and reused across runs
As you can in this image (this is from the last PR merged by Josh), the environment is actually using cached `nbQA` that is using 3.12 python version. 
Screenshot 2026-04-20 011557
- Updating the docformatter config invalidated the cache, forcing a fresh rebuild
  on Python 3.14, which exposed the latent failure (as u cann see in the image, it is using python3.14, and eventually it get failed)
Screenshot 2026-04-20 011758
Screenshot 2026-04-20 011901

- Pinning `language_version: python3.12` ensures both hooks build on a compatible Python

### Changes
- **docformatter**: `repo: local` → `repo: https://github.com/PyCQA/docformatter` at `v1.7.7`
- **docformatter**: added `language_version: python3.12` + `additional_dependencies: [tomli]`
- **nbQA hooks**: added `language_version: python3.12` to both `nbqa-ruff-format` and `nbqa-ruff`
- Removed redundant local hook fields (`name`, `entry`, `language`, `types`)

Fixes #1266

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,10 +17,12 @@ repos:
     hooks:
       - id: nbqa-ruff-format
         additional_dependencies: [ruff]
+        language_version: python3.12
         files: ^docs/user_guide/examples/.*\.ipynb$
       - id: nbqa-ruff
         args: ["--fix"]
         additional_dependencies: [ruff]
+        language_version: python3.12
         files: ^docs/user_guide/examples/.*\.ipynb$
 
   # Docstrings: auto-format
@@ -30,6 +32,7 @@ repos:
         name: docformatter
         entry: docformatter
         language: python
+        language_version: python3.12
         additional_dependencies: [docformatter]
         types: [python]
         args: ["--in-place"]


musaqlain:
this pr precedes all the recent PRs' who are in queue and are getting pre-commit.ci error during building.

cc @henrykironde

.pre-commit-config.yaml
```diff
@@ -12,26 +12,29 @@ repos:
...
     hooks:
       - id: docformatter
-        name: docformatter
-        entry: docformatter
-        language: python
-        additional_dependencies: [docformatter]
-        types: [python]
+        additional_dependencies: [tomli]
+        # TODO: remove language_version pin once docformatter drops the `untokenize` dependency
+        # which uses `ast.Constant.s` — removed in Python 3.14, causing pre-commit.ci to fail
```
henrykironde: Could we clean this up and remove the comments?

henrykironde: Could we minimize the changes here? Let’s try using the previous config and just add language_version: python3.12

.pre-commit-config.yaml
```diff
@@ -12,26 +12,29 @@ repos:
         files: citation_count.json
 
   # Notebooks: run Ruff via nbQA inside .ipynb
+  # language_version pin avoids ast.Str removal issue on Python 3.14
```
henrykironde: We do not need this comment

henrykironde (changes requested):

.pre-commit-config.yaml
musaqlain: done, reverted to the previous `repo: local` config and only added `language_version: python3.12` 

Also worth noting: docformatter released [v1.7.8](https://github.com/PyCQA/docformatter/pull/342) yesterday with Python 3.14 support. what do you think, should we use it? if we use this new version then we could switch from `repo: local` to `repo: https://github.com/PyCQA/docformatter` at `rev: v1.7.8` and remove the `language_version` pin for docformatter (nbQA pin would still be needed though).

musaqlain: update: I tested v1.7.8 on my fork and it reformats some existing files due to bug fixes in the new version ([#335](https://github.com/PyCQA/docformatter/pull/335), [#336](https://github.com/PyCQA/docformatter/pull/336)). so upgrading to v1.7.8 would be better as a separate follow-up PR. for now, the minimal `language_version: python3.12` pin should be good.


what do you think? 👀

henrykironde (changes requested):
Can we try to undo line 20 and 25 and see if it passes

musaqlain:
> Can we try to undo lines 20 and 25 and see if it passes?

Hello Henry, I already tested it without pinning the Python 3.12 version on lines 20 and 25, and it failed. This is because it then defaults to Python 3.14; since `ast.Str` was removed in 3.14, and nbQA depends on it, it triggers the following error:

```
nbQA failed to process docs/user_guide/examples/Bird_FineTuning.ipynb 
with exception "AttributeError("module 'ast' has no attribute 'Str'")"
```

The `language_version: python3.12` pin is required for nbQA until they release a fix.

--
Previously, nbQA was running correctly because it was using a cached 3.12 version. However, once I updated `.pre-commit-config.yaml`, it triggered nbQA to run using Python 3.14, which is why it failed.

henrykironde (approved):

henrykironde:
Thanks @musaqlain for these tests.

PR #1381, opened 2026-04-22
add point model

jveitchmichaelis:
# Point prediction model

This PR adds a model that can be used for keypoint/point detection. The (segmentation) model is trained to output a 1/4 size density map with sharp peaks corresponding to training data (i.e. trees). A local maximum detector is run over the map and these peaks are returned to the user, with scores integrated from a small radius. There is no technical limit on the number of predictions per forward pass, and the model seems to perform well in tiled scenarios.

The architecture is based on Treeformer, using the supervised branch, which in turn is a modification of DM-Count and some other past work. I've trained checkpoints on the authors' KCL dataset which replicates their results, and a pretrained checkpoint on NEON Lidar which seems to transfer quite well to other unseen datasets, like some of our test imagery in the everglades.

Default checkpoint is here: https://huggingface.co/weecology/deepforest-keypoint/tree/main and can be loaded with `deepforest(config='keypoint')`.

This branch also includes tested inference code (I've checked tiling) and refactors some of the prediction post-processing to handle the different geometries a little more cleanly. The upshot is `main` is largely unchanged.

I'm not a huge fan of adding scikit-image as a dependency, but their peak detection algorithm (`peak_local_max`) is very good and I've not found a suitable alternative. There is a discussion on GH about this, but no conclusion.

This branch does not have training support (forward doesn't return a loss), which is in a separate [branch](https://github.com/weecology/DeepForest/pull/1371) and will be split out separately for ease of review.

This example is from Hidden, resampled to 5 cm/px, but probably needs a tweak to the NMS (peak separation) threshold.

image

### Replication

Sample predictions trained and tested on the paper's KCL dataset (orange open circles are gt, annotated Google Earth imagery at 0.2 m/px):

IMG_158

density plot-6500

Note: this dataset is probably overfit. The paper reports 500 epochs, but the dataset is only 400 images. Nevertheless, the implementation here beats the paper benchmark in supervised mode and also the unsupervised benchmark (likely due to backbone pretraining) based on MAE (paper: 16.7/18.5, ours: 18.24 using density sum / <15 with peak extraction). We get a peak F1 of around 0.68. Convergence is seen after around 200 epochs, but we train for 500 to match the original hyperparameters.

image

Predictions after a single epoch on NEON LIDAR (purple are preds). The model used for the prediction above was trained for 10 epochs and will be updated when we have a final version as it still had training headroom.

prediction sample-825

prediction sample-825_2

### Model differences

- TreeFormer is, as far as I can tell, not invariant to input image shape. The vanilla implementation trains on 256 crops and then predicts over 256 tiles. If you change the input shape at test time, the model will under or over-predict tree density. This implementation changes the output to regress _density_ instead of absolute count, and learns a scale factor that is applied to the input image. This does only affect scale factor, but it's nice to be consistent here.
-  In the original paper the model learns spatial structure and count separately. In this version, we couple the two together by forcing the sum of the density map to match the object count for consistency. Typically during training, the model will quickly learn how to place tree "points" and then will slowly learn to correct the predict count output.

### Licensing

TreeFormer does not list a license, and the ancestral code from DM-Count (which is mostly copied verbatim in TreeFormer) is MIT-licensed. Here, the backbone model is taken directly from `transformers` and the various heads are re-implemented in PyTorch with some optimization from the original.

## Related Issue(s)

## AI-Assisted Development

Claude (Opus/Sonnet 4.6) and Codex 5.3/5.4

### DIFF
--- a/docs/user_guide/13_annotation.md
+++ b/docs/user_guide/13_annotation.md
@@ -73,7 +73,7 @@ Yes! Object detection models use non-annotated areas of an image as negative dat
 
 1. **Select Important Images**: Duplicate backgrounds or objects contribute little to model generalization. Focus on gathering a wide variety of object appearances.
 2. **Avoid Over-splitting Labels**: Often, using a superclass for detection followed by a separate model for classification is more effective. See the [`CropModel`](03_cropmodels.md) for an example.
-3. **Balance Accuracy and Practicality**: Depending on the goal (e.g., object counting or detection), keypoints can sometimes be used instead of precise boxes to simplify the process.
+3. **Balance Accuracy and Practicality**: Depending on the goal (e.g., object counting or detection), point/keypoint annotations can sometimes be used instead of precise boxes to simplify the process.
 
 ## Quick Video on Annotating Images
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,6 +51,7 @@ dependencies = [
     "pyyaml>=5.1.0",
     "rasterio",
     "safetensors<0.6.0",
+    "scikit-image>=0.25.2",
     "setuptools",
     "shapely>2.0.0",
     "slidingwindow",
--- a/src/deepforest/conf/config.yaml
+++ b/src/deepforest/conf/config.yaml
@@ -142,3 +142,8 @@ cropmodel:
     normalize:
     # Number of pixels to expand bbox crop windows for better prediction context.
     expand: 0
+
+point:
+  score_integration_radius: 5
+  nms_distance_thresh: 5.0
+  distance_threshold: 10
--- a/src/deepforest/conf/point.yaml
+++ b/src/deepforest/conf/point.yaml
@@ -1,0 +1,22 @@
+architecture: treeformer
+
+model:
+  name: 'weecology/deepforest-tree-point'
+  revision: 'main'
+
+# This model was trained on 10 cm/px data and has been tested
+# at resolutions around 5-20 cm/px. If your data is significantly
+# different, for example < 5 cm/px drone data, we suggest first
+# downsampling your orthomosaics.
+patch_size: 512
+patch_overlap: 0.1
+
+# Setting a threshold below 0.3 is likely to include a lot of
+# noisy/low-confidence predictions.
+score_thresh: 0.3
+
+point:
+  backbone: 'pvt_v2_b3'
+  score_integration_radius: 5
+  nms_distance_thresh: 5.0
+  distance_threshold: 10.0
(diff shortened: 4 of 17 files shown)


jveitchmichaelis:
Adding a doc page now, but this is good to review.

The coverage gaps are training related - I'll add a sanity test case that'll catch those for now.

Hmm. Maybe we want to call the checkpoint `deepforest-keypoint-tree`? Then we open the door to other classes more easily.

jveitchmichaelis:
@ethanwhite I haven't tapped you for a review, but probably good to have everyone's eyes on this.

bw4sz:
waiting on the metrics to review.

jveitchmichaelis:
Metrics added, I really thought we'd merged that already but I think just the dataset...

@bw4sz ready for you, and then I'll squash.

To avoid confusion, I've renamed all everything to `point` instead of `keypoint` as the names are interchangeable in the literature, but keypoint is mostly used for pose/skeleton estimation.

Prediction results tested with:

```
uv run deepforest --config-name point predict /tmp/odm_orthophoto_005m.tif --mode tile
```

And Treeformer's KCL paper results I've tested as (I didn't push the checkpoint, but might as well):

```
uv run deepforest evaluate data/treeformer/test.csv \
  --root-dir data/treeformer/images \
  architecture=treeformer \
  "model.name=scratch/london_025_hf" \
  score_thresh=0.15 \
  "point.score_integration_radius=2"
```

```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃      Validate metric      ┃       DataLoader 0        ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│      point_precision      │     0.703460156917572     │
│       point_recall        │    0.6459216475486755     │
│         val_loss          │    17.438983917236328     │
│          val_mae          │    16.968944549560547     │
└───────────────────────────┴───────────────────────────┘
```

jveitchmichaelis:
Once https://github.com/weecology/DeepForest/pull/1384 is in, I'll refactor the test here.

ethanwhite:
> Once #1384 is in, I'll refactor the test here.

Just merged #1384

bw4sz (approved):
Looks great. We should probably call this 'point' and not 'keypoint' anywhere, since keypoint makes me think of deeplabcut style things and I think our target user will be happier with point. Next step is to see what the score is on the MillionTrees eval. Great work!

PR #1384, opened 2026-04-27
manually specify repo IDs for huggingface test cases

jveitchmichaelis:
## Description

Existing unit test for HF model download will pull all models even if DeepForest can't handle them. This PR specifies checkpoints explicitly and runs two parametrized tests over the crop and box models. This will extend to point/polygon naturally.

## Related Issue(s)

CI failing


ethanwhite (approved):
LGTM!

PR #1388, opened 2026-05-01
security mitigation for lightning compromise

jveitchmichaelis:
## Description

Pytorch Lightning 2.6.2 and 2.6.3 (at least) have been compromised.

This PR bounds versions to <= 2.6.1 which is currently marked as "safe" and also adds a [cooldown](https://docs.astral.sh/uv/concepts/resolution/#dependency-cooldowns) for `uv` so that packages are only updated if the release is more than a week old. Users may need to update uv (uv self update).

## Related Issue(s)

https://github.com/Lightning-AI/pytorch-lightning/security/advisories/GHSA-w37p-236h-pfx3

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ dependencies = [
     "pandas",
     "pillow>6.2.0",
     "psutil",
-    "pytorch-lightning>=2.6.0,<3.0.0",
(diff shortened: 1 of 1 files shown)


ethanwhite (approved):
Thanks @jveitchmichaelis

PR #1391, opened 2026-05-13
fix: move predict_image input tensor to the model's device

### LINKED ISSUE #1390: predict_image fails on CUDA: input tensor not moved to model device
## Summary

`deepforest.main.deepforest.predict_image`raises`RuntimeError: Input type (torch.FloatTensor) and weight type (torch.cuda.FloatTensor)` whenever the model has been moved to a CUDA device (e.g. via `model.model.to('cuda')`). The helper `deepforest.predict._predict_image_` constructs the input tensor on CPU and never moves it to the model's device before the forward pass.

## Where

[`src/deepforest/predict.py`](https://github.com/weecology/DeepForest/blob/main/src/deepforest/predict.py), function `_predict_image_`:

```python
def _predict_image_(model, image, path, nms_thresh):
    image = torch.tensor(image).permute(2, 0, 1)
    image = image / 255

    with torch.no_grad():
        prediction = model(image.unsqueeze(0))   # crash if model on cuda
```

The image tensor is always allocated on CPU. If the underlying detector (`model.model`) was moved to CUDA, the forward pass fails with the device mismatch error.

## Reproduction

```python
from deepforest import main
m = main.deepforest()
m.load_model(model_name='weecology/deepforest-tree', revision='main')
m.model.to('cuda')                 # equivalent to what Lightning does in trainer.predict
m.predict_image(path='some_rgb.tif')
# RuntimeError: Input type (torch.FloatTensor) and weight type (torch.cuda.FloatTensor)
# should be the same or input should be a MKLDNN tensor and weight is a dense tensor
```

Encountered downstream in the geoai QGIS plugin: https://github.com/opengeos/geoai/issues/742.

## Suggested fix

Move the input tensor to the model's device before the forward pass:

```python
image = torch.tensor(image).permute(2, 0, 1)
image = image / 255
image = image.to(next(model.parameters()).device)
with torch.no_grad():
    prediction = model(image.unsqueeze(0))
```

Happy to send a PR if this fix looks reasonable.

## Environment

- DeepForest `main` (verified on the installed copy at `deepforest/predict.py`)
- Reported with DeepForest installed via pip on Windows + QGIS 3.42.2, but the bug is platform-independent.

giswqs:
## Summary

Fixes #1390.

`deepforest.predict._predict_image_` builds the input tensor on CPU and never moves it to the model's device, so any caller that has placed the model on CUDA (`model.model.to('cuda')`, or a Lightning trainer that has trained on GPU) gets:

```
RuntimeError: Input type (torch.FloatTensor) and weight type (torch.cuda.FloatTensor)
should be the same or input should be a MKLDNN tensor and weight is a dense tensor
```

This PR moves the input tensor to `next(model.parameters()).device` before the forward pass. Guarded with `try/except StopIteration` to handle the (unrealistic) case of a parameter-less model gracefully.

## Repro

```python
from deepforest import main
m = main.deepforest()
m.load_model(model_name='weecology/deepforest-tree', revision='main')
m.model.to('cuda')
m.predict_image(path='some_rgb.tif')
# before this PR: RuntimeError as above
# after this PR: returns a DataFrame of detections
```

## Changes

- `src/deepforest/predict.py`: move the input tensor to the model's device inside `_predict_image_`.
- `tests/test_main.py`: add `test_predict_image_on_cuda` regression test, skipped when CUDA is unavailable.

## Test plan

- [x] `ruff format --check` and `ruff check` on the changed source files (clean, using the version pinned in `.pre-commit-config.yaml`).
- [ ] CI: existing CPU `predict_image` tests should still pass; new CUDA test is auto-skipped on CPU-only runners.
- [ ] Verified locally that the workaround applied downstream in [opengeos/geoai#746](https://github.com/opengeos/geoai/pull/746) becomes unnecessary once this lands.

### DIFF
--- a/src/deepforest/predict.py
+++ b/src/deepforest/predict.py
@@ -35,6 +35,10 @@ def _predict_image_(
     """
     image = torch.tensor(image).permute(2, 0, 1)
     image = image / 255
+    try:
+        image = image.to(next(model.parameters()).device)
+    except StopIteration:
+        pass
 
     with torch.no_grad():
         prediction = model(image.unsqueeze(0))
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -411,6 +411,19 @@ def test_predict_image_fromfile(m):
     assert not prediction.empty
 
 
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
+def test_predict_image_on_cuda(m):
+    """Regression: model on CUDA must not raise a device mismatch in predict_image."""
+    m.model.to("cuda")
+    try:
+        prediction = m.predict_image(
+            path=get_data(path="2019_YELL_2_528000_4978000_image_crop2.png")
+        )
+    finally:
+        m.model.to("cpu")
+    assert isinstance(prediction, pd.DataFrame)
+
+
 def test_predict_image_fromarray(m):
     image_path = get_data(path="2019_YELL_2_528000_4978000_image_crop2.png")
 


bw4sz:
Thanks @giswqs! @jveitchmichaelis will have a look at this.

jveitchmichaelis:
Thanks, this looks pretty sensible to me. I'll double check I can reproduce too.

jveitchmichaelis:
@giswqs thanks for raising this, I can replicate although I was surprised that it failed. The additional cause is that the model isn't moved to the device until a trainer uses it and `predict_image` is a low level call that doesn't invoke `trainer.predict` which would normally handle device placement.

Alternatively, you could use `tiled` prediction and I think you would avoid this (if you want a single tile, should work if you set the patch size to the size of the image).

So I'd propose:

1. I think this solution is fine in the short term, if this is how you're using the model downstream.
2. Longer term I think we would replace the guts of this function with a call to `trainer.predict` using a `SingleImage` dataset (or `MultiImage`). However this would only work with paths, so maybe (1) is still required if the user passes an array.

I'll leave some comments on the test case in the review.

@bw4sz  any thoughts there? On the one hand I feel we shouldn't encourage manually moving the model, because `accelerator` is in the config. On the other, if you specify `cuda`, the model doesn't actually get moved until the trainer is invoked. The complication is that Lightning also handles device names for us, so if we tried to add that in `predict_image` we would need to add an `auto` -> expected device helper.

bw4sz:
We can allow this for now, but definitely let's get predict_image into the trainer context that feels like the right pattern.

PR #1395, opened 2026-05-26
Edge cases for empty frames + test with non-empty frames.

bw4sz:
In the BOEM workflow I found a funny edge case where if the test dataset has both empty frames and non-empty frames, it produces a non-intuitive results. It was mathematically correct, but just confusing and too literal. We were dividing the recalls by the total number of images, so the average box_recall score was (number of matches/total number of images), but given that some of those images are empty, it produces a strange result. So if you had two images, 1 with one ground truth, correctly matched, and one that was empty, the average box_recall was 0.5, when it really should be 1 match /1 image with GT + empty image, not 1 match / (2 images). 

## AI-Assisted Development

I have coalesced around a cursor strategy where I identify errors, discuss with the agent. I then have cursor develop failing tests, which then I implement a correction for and then we all agree its passing. This feels reasonable since there is an adversarial element to it.

### DIFF
--- a/src/deepforest/metrics.py
+++ b/src/deepforest/metrics.py
@@ -63,6 +63,11 @@ class RecallPrecision(Metric):
         )
         self.add_state("num_empty_frames", default=torch.tensor(0), dist_reduce_fx="sum")
         self.add_state(
+            "num_images_with_ground_truth",
+            default=torch.tensor(0),
+            dist_reduce_fx="sum",
+        )
+        self.add_state(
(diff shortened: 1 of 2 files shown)


jveitchmichaelis:
Think you're missing the logic fix here - I only see tests

bw4sz:
That was really sad. Pushed.

jveitchmichaelis (approved):
LGTM!

PR #1396, opened 2026-06-01
point model training support

### LINKED ISSUE #809: Point detection workflow

jveitchmichaelis:
## Description

This PR adds training functionality for the default `point` model. Most of the code has been distilled from my experimental branch. It should replicate the existing checkpoint, and I'm running a quick test to make sure.

Changes are relatively small in scope - added the optimal transport loss code, fleshed out `compute_losses` for the model and fixed a small bug in point visualization where the VertexAnnotator wouldn't accept a palette (we instead convert points to "detections" with a radius and plot them as circles).

I've included a `point_pretrain` config which was used to train the current default point model on NEON data, replication should be possible with:

`uv deepforest --config-name point_pretrain train`

## Related Issue(s)

Closes https://github.com/weecology/DeepForest/issues/809 as we now support training, visualization of predictions, simple unit tests + should also include multi-class support in theory but we've not tested it.

## AI-Assisted Development

Claude code for assistance with merging the changes from the other branch.

### DIFF
--- a/src/deepforest/conf/point_pretrain.yaml
+++ b/src/deepforest/conf/point_pretrain.yaml
@@ -1,0 +1,63 @@
+# NEON pretraining config for point model.
+
+defaults:
+  - config
+  - _self_
+
+architecture: treeformer
+
+# Train from ImageNet PvT weights, not the released point checkpoint.
+model:
+  name: null
+  revision: main
+
+num_classes: 1
+label_dict:
+  Tree: 0
+
+batch_size: 32
+workers: 8
+accelerator: auto
+precision: bf16-mixed
+matmul_precision: high
+score_thresh: 0.15
+patch_size: 512
+
+# Replace the CSV paths here with the training dataset e.g. full NEON pretrain (all_annotations_train_fold_0.csv)
+train:
+  csv_file: MISSING
+  root_dir: MISSING
+  lr: 2e-4
+  optimizer:
+    type: AdamW
+    weight_decay: 1e-5
+  epochs: 40
+  scheduler:
+    type: cosine
+    params:
+      T_max: 40
+      eta_min: 1e-6
+  augmentations:
+    - HorizontalFlip: {p: 0.5}
+    - VerticalFlip: {p: 0.5}
+    - Rotate: {degrees: 180, p: 0.5}
+
+validation:
+  csv_file: MISSING
+  root_dir: MISSING
+  val_accuracy_interval: 1
+
+point:
+  backbone: OpenGVLab/pvt_v2_b3
+  score_integration_radius: 2
+  distance_threshold: 10.0
+  density_sigma: 3.0
+  mae_weight: 0.025
+  ot_weight: 0.1
+  density_l1_weight: 0.05
+  count_cls_weight: 0.1
+  sinkhorn_reg: 1.0
+  num_of_iter_in_ot: 100
+  losses: [count, ot, density_l1]
+  norm_cood: true
+  enforce_count: false
--- a/src/deepforest/losses/ot_loss.py
+++ b/src/deepforest/losses/ot_loss.py
@@ -1,0 +1,240 @@
+"""PyTorch implementation of Sinkhorn-Knopp for optimal transport.
+
+Based on `ot.bregman.sinkhorn` from the Python Optimal Transport library,
+https://pythonot.github.io, rewritten in PyTorch and adapted from the
+DM-Count/TreeFormer repositories.
+
+Some small changes have been made to the original code for stability and
+logging.
+
+Original implementation:
+https://github.com/cvlab-stonybrook/DM-Count/blob/master/losses/ot_loss.py
+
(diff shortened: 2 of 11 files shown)


jveitchmichaelis:
This looks good enough for review. I’ll run some more extensive tests to see if there’s any other features that need to be pulled in, but the core losses are all here.

src/deepforest/conf/point_pretrain.yaml
```diff
@@ -0,0 +1,62 @@
...
+batch_size: 32
+workers: 8
+accelerator: auto
+precision: bf16-mixed
+matmul_precision: high
+score_thresh: 0.15
+patch_size: 512
+
+train:
+  csv_file: /blue/ewhite/veitchmichaelisj/deeplidar/output/folds/all_annotations_train_fold_0.csv
```
bw4sz: Should we have these kinds of default paths in the /src dir? Would it effect anything if changed it to blank?

src/deepforest/conf/point_pretrain.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - HorizontalFlip: {p: 0.5}
+    - VerticalFlip: {p: 0.5}
+    - Rotate: {degrees: 180, p: 0.5}
+
+validation:
+  csv_file: /blue/ewhite/veitchmichaelisj/deeplidar/output/folds/all_annotations_val_fold_0.csv
+  root_dir: /blue/ewhite/veitchmichaelisj/deeplidar/output
+  val_accuracy_interval: 1
+
+point:
```
bw4sz, line 50: wow that is alot of hyperparamters. Do you have an intuition which is most powerful, to give a hint in the docs?

src/deepforest/losses/ot_loss.py
```diff
@@ -0,0 +1,240 @@
...
+                # compute l2 square distance, it should be source target distance. [#gt, #cood * #cood]
+                if self.norm_coord:
+                    x = im_points[:, 0].unsqueeze(1) / output_w * 2 - 1
+                    y = im_points[:, 1].unsqueeze(1) / output_h * 2 - 1
+                else:
+                    x = im_points[:, 0].unsqueeze(1)
+                    y = im_points[:, 1].unsqueeze(1)
+
+                x_dis = (
+                    -2 * torch.matmul(x, x_cood) + x * x + x_cood * x_cood
```
bw4sz, line 170: I'll need to read the paper, can you give me like a 3 sentence intuition of how to interpret the loss?

bw4sz:
I had one tiny issue that isn't probably related to this PR, but just in case. If you let the trainer default to the csv logger it hits

```
ValueError: dict contains fields not in fieldnames
```
which basically means that the schema is set on trainer.start and it hits metrics it wasn't 
Pre-log the metric with a value of 0 or float('nan') in your setup() hook or on the first batch, so the logger registers the column header from the very beginning.

```python

def on_train_start(self):
    # Register the metric key so CSVLogger knows about it from step 1
    self.log("epoch level metric", 0.0)
```

I think the right thing to do is merge this PR and address this at the module level, its not specific to this PR.

jveitchmichaelis:
Which metric is this triggered for? I’ll have a look if I missed something when I pulled these changes from the other branch.

bw4sz:
All of the epoch level metrics, this is just the csv logger, the tensorboard and comet logger handle this gracefully. 

```
train_sinkhorn_beta_abs_max_epoch, train_ot_loss_epoch, train_count_mae_epoch
```

I think this is a different PR that we have an object that captures the metric names and logs them to 0 or 'nan' on train start, since this would be true for any metric in any of the workflows.

tests/conftest.py
```diff
@@ -93,3 +93,21 @@ def m(download_release, tmp_path_factory):
     m.load_model("weecology/deepforest-tree")
 
     return m
+
+@pytest.fixture(scope="session")
+def m_point(download_release, tmp_path_factory):
+    m = main.deepforest(config="point")
```
bw4sz, line 99: This line confuses me, isn't the .yaml called point_pretrain.yaml? not point?

bw4sz (approved):

PR #1398, opened 2026-06-04
Use trainer.predict for single images

jveitchmichaelis:
## Description

Use the trainer for single image prediction, similar to other predict routes (predict_file).

I've used `SingleImage` as the dataset which seems sensible, but this accepts a `patch_size` option which is not the behavior we had before. Prediction would use the entire image. Here I've forced patch_size to be the same size as the image **only** in `deepforest.predict_image` which is a hack, but the behaviour is the same as before.

I've refactored `_dataloader_wrapper_` for convenience and added crop_model support for single images. It now uses `_crop_models_wrapper_` which has the same logic for multiple models.

There are some inconsistencies that we could look into - predict_file returns an empty dataframe for no predictions, while predict_tile and predict_image (previously and here) return None. Some prediction paths convert labels from int->str - plan to make everything consistent there?

## Related Issue(s)

Would supersede and fix https://github.com/weecology/DeepForest/pull/1391

## AI-Assisted Development

N/A

### DIFF
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -525,12 +525,15 @@ class deepforest(pl.LightningModule):
         )
         return loader
 
-    def predict_image(self, image: np.ndarray | None = None, path: str | None = None):
+    def predict_image(
+        self, image: np.ndarray | None = None, path: str | None = None, crop_model=None
+    ):
         """Predict a single image with a deepforest model.
 
         Args:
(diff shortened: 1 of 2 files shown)


jveitchmichaelis:
Ready for review and we can address the inconsistency separately. For now this should fix the device placement bug.

bw4sz (approved):

PR #1399, opened 2026-06-10
Add end-to-end polygon (Mask R-CNN) workflow

bw4sz:
Implement instance segmentation paralleling the box and point workflows:
- Add Mask R-CNN model (task="polygon"), encapsulating the torchvision background-class convention by shifting labels in forward()
- Refactor PolygonDataset to mirror BoxDataset (csv_file, 3-tuple, joint image/box/mask augmentation)
- Support polygon prediction formatting and mask-based geometry detection
- Wire segmentation mAP and polygon recall/precision into validation
- Enable polygon mosaicking in predict_tile via reduce_polygons
- Add tests for the dataset, model, formatting, training and prediction

## AI-Assisted Development

- [ ] I used AI tools (e.g., GitHub Copilot, ChatGPT, etc.) in developing this PR
- [ ] I understand all the code I'm submitting
- [ ] I have reviewed and validated all AI-generated code

### DIFF
--- a/src/deepforest/datasets/prediction.py
+++ b/src/deepforest/datasets/prediction.py
@@ -110,8 +110,11 @@ class PredictionDataset(Dataset):
 
     def determine_geometry_type(self, batched_result):
         """Determine the geometry type of the batched result."""
-        # Assumes that all geometries are the same in a batch
-        if "boxes" in batched_result.keys():
+        # Assumes that all geometries are the same in a batch. Mask R-CNN
+        # results contain both "masks" and "boxes"; masks take precedence.
+        if "masks" in batched_result.keys():
+            geom_type = "polygon"
+        elif "boxes" in batched_result.keys():
             geom_type = "box"
         elif "points" in batched_result.keys():
             geom_type = "point"
--- a/src/deepforest/datasets/training.py
+++ b/src/deepforest/datasets/training.py
@@ -584,10 +584,11 @@ class PointDataset(TrainingDataset):
 class PolygonDataset(TrainingDataset):
     """Dataset for instance segmentation with polygon masks.
 
-    Parallels :class:`BoxDataset` but additionally rasterizes each polygon
-    geometry into a binary instance mask. Targets contain ``boxes``
-    (the polygon bounds), ``labels`` and ``masks`` for use with Mask R-CNN.
-    Labels are zero-indexed to match the box and point workflows.
+    Parallels :class:`BoxDataset` but additionally rasterizes each
+    polygon geometry into a binary instance mask. Targets contain
+    ``boxes`` (the polygon bounds), ``labels`` and ``masks`` for use
+    with Mask R-CNN. Labels are zero-indexed to match the box and point
+    workflows.
     """
 
     _data_keys = [DataKey.IMAGE, DataKey.BBOX_XYXY, DataKey.MASK]
--- a/src/deepforest/datasets/training.py
+++ b/src/deepforest/datasets/training.py
@@ -582,37 +582,25 @@ class PointDataset(TrainingDataset):
 
 
 class PolygonDataset(TrainingDataset):
+    """Dataset for instance segmentation with polygon masks.
+
+    Parallels :class:`BoxDataset` but additionally rasterizes each polygon
+    geometry into a binary instance mask. Targets contain ``boxes``
+    (the polygon bounds), ``labels`` and ``masks`` for use with Mask R-CNN.
+    Labels are zero-indexed to match the box and point workflows.
+    """
+
     _data_keys = [DataKey.IMAGE, DataKey.BBOX_XYXY, DataKey.MASK]
 
-    def __init__(
-        self,
-        annotation_file,
-        root_dir,
-        transforms=None,
-        augmentations=None,
-        label_dict=None,
-        preload_images=None,
-        validate_coordinates=True,
-    ):
-        super().__init__(
-            csv_file=annotation_file,
-            root_dir=root_dir,
-            transforms=transforms,
-            augmentations=augmentations,
-            label_dict=label_dict,
-            preload_images=preload_images,
-            validate_coordinates=validate_coordinates,
-        )
-
     def _validate_coordinates(self) -> None:
-        """Validate that all points occur within the image.
+        """Validate that all polygons are valid and within the image.
 
         Raises:
-            ValueError: If any point occurs outside the image
+            ValueError: If any polygon is invalid or exceeds image bounds.
         """
         errors = []
-        for _, row in self.annotations.iterrows():
-            img_path = os.path.join(self.root_dir, row["image_path"])
+        for image_path, group in self.annotations.groupby("image_path"):
+            img_path = os.path.join(self.root_dir, image_path)
             try:
                 with Image.open(img_path) as img:
                     width, height = img.size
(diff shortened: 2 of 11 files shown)


src/deepforest/main.py
```diff
@@ -844,29 +798,17 @@ def validation_step(self, batch, batch_idx):
...
+        self.log("val_loss", losses.detach(), on_epoch=True, batch_size=len(images))
 
         # In eval model, return predictions to calculate prediction metrics
         self.model.eval()
         with torch.no_grad():
             preds = self.model.forward(images, targets)
 
         # Compute precision, recall and empty frame metrics.
-        if self.model.task == "box" or self.model.task == "point":
+        if self.model.task in ("box", "point", "polygon"):
```
jveitchmichaelis, line 811: Redundant? At least for now we can compute PR for any task so we can always add this metric.

src/deepforest/datasets/prediction.py
```diff
@@ -118,17 +108,13 @@ def get_image_basename(self, idx):
...
-        return {
-            "image_path": self.get_image_basename(idx),
-            "window_bounds": self.get_crop_bounds(idx),
-        }
-
     def determine_geometry_type(self, batched_result):
         """Determine the geometry type of the batched result."""
-        # Assumes that all geometries are the same in a batch
-        if "boxes" in batched_result.keys():
+        # Assumes that all geometries are the same in a batch. Mask R-CNN
```
jveitchmichaelis, line 113: I would remove the reference to Mask R-CNN here and say "Polygon model results that contain both ... "

src/deepforest/models/maskrcnn.py
```diff
@@ -0,0 +1,201 @@
...
+    def __init__(
+        self,
+        backbone_weights: str | None = None,
+        num_classes: int = 1,
+        nms_thresh: float = 0.05,
+        score_thresh: float = 0.5,
+        label_dict: dict = None,
+        **kwargs,
+    ):
+        backbone = torchvision.models.detection.maskrcnn_resnet50_fpn(
```
jveitchmichaelis, line 37: There is also V2. https://docs.pytorch.org/vision/main/models/mask_rcnn.html

> Improved Mask R-CNN model with a ResNet-50-FPN backbone from the [Benchmarking Detection Transfer Learning with Vision Transformers](https://arxiv.org/abs/2111.11429) paper.

tests/test_maskrcnn.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def test_predict_outputs_masks():
+    model = build_model(score_thresh=0.0)
+    model.eval()
+    x = [torch.rand(3, 300, 400), torch.rand(3, 400, 300)]
+    predictions = model(x)
+
+    assert len(predictions) == 2
+    assert sorted(predictions[0].keys()) == ["boxes", "labels", "masks", "scores"]
```
jveitchmichaelis, line 45: Should also check the shape of the masks.

tests/test_maskrcnn.py
```diff
@@ -0,0 +1,151 @@
...
+    targets[0]["masks"][0, 10:50, 10:50] = 1
+
+    loss_dict = model(images, targets)
+
+    assert "loss_mask" in loss_dict
+    # The caller's target labels must remain zero-indexed
+    assert targets[0]["labels"].tolist() == [0]
+
+
+def test_check_model():
```
jveitchmichaelis, line 71: Redundant with test_predict_outputs_masks?

tests/test_maskrcnn.py
```diff
@@ -0,0 +1,151 @@
...
+
+    m.trainer.fit(m)
+    m.trainer.validate(m)
+
+    # Segmentation mAP and polygon recall/precision are logged
+    logged = m.trainer.logged_metrics
+    assert any("polygon" in key for key in logged) or "map" in logged
+
+
+def test_polygon_predict_image(polygon_root_dir):
```
jveitchmichaelis, line 111: _synthetic and we should add a real example when trained.

tests/test_maskrcnn.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # Random weights may or may not produce detections; if they do, they must
+    # be polygons in the standard results schema.
+    if result is not None:
+        assert "geometry" in result.columns
+        assert "label" in result.columns
+        assert result.geometry.iloc[0].geom_type in ("Polygon", "MultiPolygon")
+
+
+def test_polygon_predict_tile():
```
jveitchmichaelis, line 132: This is fine for now but we could move prediction tests to somewhere else and parametrize it over the different models.

src/deepforest/main.py
```diff
@@ -119,10 +118,15 @@ def setup_metrics(self):
             )
 
         # Segmentation metrics
```
jveitchmichaelis, line 120: Instance segmentation

src/deepforest/datasets/training.py
```diff
@@ -582,239 +582,221 @@ def __getitem__(self, idx) -> tuple:
...
-        # Convert to tensor
-        image = augmented_image.squeeze()
-        boxes = augmented_boxes.squeeze()
-        masks = augmented_masks.squeeze()
-        labels = torch.as_tensor(targets["labels"], dtype=torch.int64)
-        iscrowd = torch.as_tensor(targets["iscrowd"], dtype=torch.bool)
-        area = torch.as_tensor(targets["area"])
-        image_id = targets["image_id"]
-
-        # Filter out-of-bounds boxes and sync with transformed
```
jveitchmichaelis, line 788: Keep this comment?

src/deepforest/main.py
```diff
@@ -951,40 +904,32 @@ def _compute_epoch_metrics(self) -> dict:
...
-        synced_metrics = {}
-
-        for key, value in metrics.items():
-            if isinstance(value, torch.Tensor):
-                synced_metrics[key] = value.to(self.device)
-            elif isinstance(value, Number):
-                synced_metrics[key] = torch.tensor(value, device=self.device)
-            else:
-                synced_metrics[key] = value
+        elif self.model.task == "polygon":
```
jveitchmichaelis, line 907: Not obvious why this should be separate logic to boxes?

src/deepforest/main.py
```diff
@@ -1227,13 +1165,9 @@ def evaluate(
...
         results = {}
-        if isinstance(validation_results, list):
-            if validation_results and isinstance(validation_results[0], dict):
-                results.update(validation_results[0])
-        elif isinstance(validation_results, dict):
-            results.update(validation_results)
+        results.update(self.trainer.logged_metrics)
         results["predictions"] = self.predictions
-        if self.model.task == "box" or self.model.task == "point":
+        if self.model.task in ("box", "point", "polygon"):
```
jveitchmichaelis, line 1170: Drop the condition here as this is valid for all model tasks

src/deepforest/predict.py
```diff
@@ -152,6 +155,39 @@ def reduce_points(predictions: pd.DataFrame, nms_thresh: float) -> pd.DataFrame:
     return predictions.iloc[np.flatnonzero(kept)].reset_index(drop=True)
 
 
+def reduce_polygons(predictions: pd.DataFrame, iou_threshold: float) -> pd.DataFrame:
```
jveitchmichaelis, line 158: Does this need to be a different function? Polygons aren't used at all here, only box bounds. If the result is that the dataframe/preds are filtered by index then it shouldn't matter if we ran reduce_boxes.

Might need to refactor with an `output_columns` argument?

src/deepforest/datasets/training.py
```diff
@@ -582,239 +582,221 @@ def __getitem__(self, idx) -> tuple:
...
-    def annotations_for_path(self, image_name, image_path, return_tensor=False):
-        try:
-            print("opening image")
-            with Image.open(image_path) as img:
-                width, height = img.size
-        except Exception as e:
-            print(f"Failed to open image {image_path}: {e}")
-            return
-        image_annotations = self.annotations[self.annotations.image_path == image_name]
+    def annotations_for_path(self, image_path, return_tensor=False) -> dict:
```
jveitchmichaelis, line 671: I don't think this function works? We're loading an MS-COCO JSON file during testing (and presumably for training), but this treats it as if it's a CSV with geometry entries etc. See the previous diff where we handle the segmentation section.

jveitchmichaelis (changes requested):
@bw4sz as far as your concerns about this, my opinion is the general feel of this PR is good. It cribs the right bits from other sections of the code. Can you feed the review notes back into Cursor?

bw4sz:
Sorry, I actually didn't mean to trigger this into main yet. Thanks for looking at it. I thought I pushed to my fork. Let's hold this open and I will clean it up on monday.

bw4sz:
Testing this branch on Million Trees, the losses drop, but the evaluation inference looks really poor compared to Detectron mask RCNN. We'll need to pull out a cleaner standalone example and compare the two and see if we get similar results. Something doesn't seem right here. Let's take it away from all the complexity of million trees and get a tiny, well-constructed data set and explore.

Issue #1402, opened 2026-07-06
Skip all black tiles in predict tile


bw4sz:
We often have orthomosaics that have these edge areas. if predict tile hits this, we should just have a default arg set to True that drops an all black or all data NA tile.

jveitchmichaelis:
Closing as implemented in #1404

PR #1404, opened 2026-07-07
add option to skip predictions on empty/uniform frames

jveitchmichaelis:
## Description

Adds support to optionally skip frames that empty. `non-empty` is defined as `(image.min() != image.max())` which conveniently covers all black or all white data (as well as anything in-between) and works whether or not the image is normalized. I don't think there is a case where we'd want predictions on a uniform image. The real check is made with `close()` to avoid float comparison issues.

Filtering is applied in `predict_step` which should affect almost all prediction paths. I deliberately did not apply this to `predict_image` or `_predict_image` but we could add it there if necessary. And this is also future-compatible to moving predict_image to use the Lightning path.

## Related Issue(s)

Implements https://github.com/weecology/DeepForest/issues/1402
Would be used by https://github.com/weecology/DeepForest/pull/1398

## AI-Assisted Development

Claude Opus with a lot of feedback.

### DIFF
--- a/src/deepforest/conf/config.yaml
+++ b/src/deepforest/conf/config.yaml
@@ -41,6 +41,9 @@ annotations_xml:
 rgb_dir:
 path_to_rgb:
 
+# Skip prediction on tiles that are either all black/white, or nodata
+skip_empty: true
+
 log_root: ./lightning_logs
 
 train:
--- a/src/deepforest/conf/schema.py
+++ b/src/deepforest/conf/schema.py
@@ -212,6 +212,9 @@ class Config:
     rgb_dir: str | None = None
     path_to_rgb: str | None = None
 
+    # Skip prediction on tiles that are all black/white or nodata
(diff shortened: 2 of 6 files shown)


jveitchmichaelis:
This is good to go pending @henrykironde to have a look at the readthedocs fail (I think a re-run will clear it).

henrykironde:
The readthedocs platform experienced some issues.

henrykironde:
I want to think we should be skipping such cases by default. Let me have a look and try. Do you have any sample image to through my way on /blue to test this.

jveitchmichaelis:
Pick any of the images from everglades? Probably easiest, there's lots of blank tiles from the orthos. I'll tag you for review if you want to check it out.

Maybe we want to think about having some orthomosaics on blue for testing in a fixed place.

And agree, it's currently set to `true` by default.

bw4sz (approved):

PR #1405, opened 2026-07-07
Feature/balanced detection batch sampling

bw4sz:
This PR allows users to balance empty and positive frames during detection training. On the BOEM dataset we found that 0.5 was fruitful. Default is null, normal random sampling. 

https://www.comet.com/bw4sz/boem/compare?compareXAxis=step&experiment-tab=panels&experiments=9d437e52a91d464cb8a94a5ef478e576,e708632048074606aa3764131ec5003f,3c03784ccd32416f8ef4b75b67c54b35&showOutliers=true&smoothing=0&viewId=new&xAxis=step

## Agent statement
I conceived of this idea and drafted a prototype alongside an agent. I then measured its performance through a series of comet experiments on the BOM dataset. When I was satisfied that it improved performance, I took over the pull request, edited the docs, reduced the test, and cleaned up the Python code.


bw4sz:
@jveitchmichaelis this is ready, I'm not 100% happy with the code styling of `_build_annotation_index_pools`, but it is functional and all internal code.

src/deepforest/datasets/training.py
```diff
@@ -70,6 +76,9 @@ def __init__(
         self._validate_labels()
         if validate_coordinates:
             self._validate_coordinates()
+        self.positive_indices, self.negative_indices = (
```
jveitchmichaelis, line 79: I don't know if there's a significant penalty here, but this is always computed regardless of whether the indices are used.

src/deepforest/datasets/training.py
```diff
@@ -135,6 +171,90 @@ def __getitem__(self, index) -> tuple:
...
+    match ``positive_indices`` positions, so ``self.sampler`` is **not** used when
+    building batches; only ``positive_indices`` and ``negative_indices`` are.
+    The ``sampler`` and ``drop_last`` arguments remain on ``__init__`` so
+    Lightning can inject values without type errors.
+
+    .. note::
+
+        Distributed sharding of the positive pool is not applied. Multi-GPU runs
+        typically need ``Trainer(use_distributed_sampler=False)`` or equivalent
+        unless you accept duplicate positive coverage per rank.
```
jveitchmichaelis, line 193: Is this an issue in practice?

src/deepforest/main.py
```diff
@@ -390,6 +358,9 @@ def load_dataset(
             preload_images: if True, preload the images into memory
             validate_coordinates: if True, check annotation coordinates fall within image bounds
             augmentations: augmentation configuration (str, list, or dict)
+            positive_batch_fraction: If set with shuffle=True, each batch contains
+                this fraction of annotated images and the rest hard negatives.
```
jveitchmichaelis, line 362: Are they actually hard negatives or empty? I understand hard negative mining to also include classes that are often mis-predicted, but not necessarily a blank image.

src/deepforest/main.py
```diff
@@ -433,13 +404,44 @@ def load_dataset(
...
+                )
+
+        if use_balanced_sampler:
+            batch_sampler = training.BalancedDetectionBatchSampler(
+                positive_indices=ds.positive_indices,
+                negative_indices=ds.negative_indices,
+                batch_size=batch_size,
+                positive_batch_fraction=positive_batch_fraction,
+            )
+            data_loader = torch.utils.data.DataLoader(
```
jveitchmichaelis, line 431: Style nit - move `DataLoader` out of the condition and set batch_sampler to `None` in the else?

tests/test_balanced_detection_sampler.py
jveitchmichaelis, line 1: Maybe a test through deepforest.main to confirm that setting the balanced sampler in the config is fine + we can load samples (no need to actually predict).

jveitchmichaelis (changes requested):

PR #1408, opened 2026-07-13
remove setuptools as a runtime dependency

jveitchmichaelis:
`setuptools` is blocking later versions of torch, torchvision etc and is not needed for runtime. Confirm by running `uv sync --upgrade` and see that torch is pinned to ~2.1.0. This PR removes it. Should not affect builds which use a separate dependency group.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,6 @@ dependencies = [
     "safetensors<0.6.0",
     "scikit-image>=0.25.2",
     "scipy>=1.15.3",
-    "setuptools",
     "shapely>2.0.0",
     "slidingwindow",
     "supervision",


bw4sz:
Makes sense.

ethanwhite (commented):
Since we still have build dependencies for this should we move it into:

```toml
[project.optional-dependencies]
```

Or does having it here take care of that?

```toml
[build-system]
requires = ["setuptools>=61.0", "wheel"]
```

Alternatively we can replace the build system.

jveitchmichaelis:
I think the latter, but will confirm.

henrykironde:
Yes this may be the better option known to me 
```python
[build-system]
requires = ["setuptools>=61.0", "wheel"]
```

ethanwhite:
@henrykironde - you had expressed some potential concerns about this one. Are you happy with it now or is there something we need to look at (and if so what)?

henrykironde (approved):

PR #1409, opened 2026-07-22
feat: allow predict_file to accept pandas DataFrame instead of only C…

### LINKED ISSUE #797: main.deepforest.predict_file should be able to take in a dataframe, not just a csv file.

proofbyhuman:
## Description
This PR resolves issue #797 by allowing the `predict_file` method to accept a pandas DataFrame directly, in addition to CSV file paths. 

**Changes Made:**
1. **`src/deepforest/main.py`**: Renamed the primary argument from `csv_file` to `input_file`. Added `csv_file` as an optional deprecated argument with a `DeprecationWarning` to maintain backward compatibility.
2. **`src/deepforest/datasets/prediction.py`**: Updated the `FromCSVFile` class `__init__` type hint to accept `str | pd.DataFrame`, removing the previous restriction that forced the input to be a string.

**Why this works:**
The downstream function `utilities.read_file` already natively supports both `str` and `pd.DataFrame` inputs. This change simply removes the unnecessary type restriction at the entry point, fulfilling the feature request without altering the core processing logic.

**Testing:**
Local environment constraints prevented running the full test suite locally, but the changes are minimal and strictly type-hint/deprecation additions. GitHub Actions will verify that existing tests continue to pass.

## Related Issue(s)
Closes #797

## AI-Assisted Development

- Qwen (AI assistant) was used to guide the step-by-step learning process, explain open-source contribution practices, and draft the code modifications. All code was manually reviewed, understood, and validated by the contributor before committing.

### DIFF
--- a/docs/user_guide/12_evaluation.md
+++ b/docs/user_guide/12_evaluation.md
@@ -198,7 +198,7 @@ m = main.deepforest()
 m.load_model(model_name="weecology/deepforest-tree", revision="main")
 
 csv_file = get_data("OSBS_029.csv")
-predictions = m.predict_file(csv_file=csv_file, root_dir=os.path.dirname(csv_file))
+predictions = m.predict_file(input_file=csv_file, root_dir=os.path.dirname(csv_file))
 predictions.head()
     xmin   ymin   xmax   ymax label     score    image_path
 0  330.0  342.0  373.0  391.0  Tree  0.802979  OSBS_029.tif
--- a/docs/user_guide/16_prediction.md
+++ b/docs/user_guide/16_prediction.md
@@ -130,7 +130,7 @@ For a list of images with annotations in a csv file, the `predict_file` function
 
 csv_file = get_data("OSBS_029.csv")
 original_file = pd.read_csv(csv_file)
-df = m.predict_file(csv_file, root_dir=os.path.dirname(csv_file))
+df = m.predict_file(input_file=csv_file, root_dir=os.path.dirname(csv_file))
 ```
 
 ```
--- a/src/deepforest/datasets/prediction.py
+++ b/src/deepforest/datasets/prediction.py
@@ -1,6 +1,7 @@
 import os
 
 import numpy as np
+import pandas as pd
 import rasterio as rio
 import slidingwindow
 import torch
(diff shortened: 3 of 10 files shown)


proofbyhuman:
Both points addressed.

**Call sites.** All 13 are updated — the 11 keyword ones (`src/deepforest/scripts/predict.py`, `tests/test_evaluate.py`, `tests/test_IoU.py`, `tests/test_main.py`, `tests/profile_evaluate.py`, `tests/profile_predict_file.py`, `docs/user_guide/12_evaluation.md`) plus the two positional ones in `tests/test_main.py` and `docs/user_guide/16_prediction.md`, which I switched to the explicit `input_file=` keyword.

**`root_dir` docstring.** It no longer promises a config fallback that isn't implemented. It now describes what the code actually does: if `root_dir` is None it comes from the directory of `input_file`, which therefore has to be a path rather than a DataFrame. I didn't implement the config fallback, since that would be new behaviour rather than a fix — happy to add it if you'd prefer that instead.

While updating the call sites I hit the bug that was actually breaking the DataFrame path: `FromCSVFile.prepare_items` wasn't forwarding `root_dir` to `read_file`, so a DataFrame input failed. Fixed in the same commit.

Local runs on this branch:

```
$ uv run pytest tests/test_IoU.py tests/test_evaluate.py tests/test_datasets_prediction.py
27 passed in 449.25s
$ uv run pytest tests/test_main.py -k predict_file
3 passed
$ uv run pytest tests/test_cli_predict.py
7 passed in 648.05s
```

vickysharma-prog:
Thanks @proofbyhuman this matches what #797 asked for few things to see 

`test_predict_file_deprecated_csv_file` is a good test comparing the two paths with `assert_frame_equal` is exactly right. Two things on the other one:

- `test_predict_file_dataframe_input` only asserts the result is non-empty and has the expected columns. The contract worth           pinning down is that a DataFrame read from the CSV gives the same predictions as the CSV itself. You already use
  `assert_frame_equal` in the test below it, so the same line here would turn a smoke test into a real one.
- `predict_file()` with neither argument raises `ValueError`, and nothing covers it. `pytest.raises(ValueError)` is already used            elsewhere in this file.

Nit: most `pytest.warns` calls in the repo pass `match=` (see `test_main.py` and `test_utilities.py`), so an unrelated `DeprecationWarning` can't satisfy the assertion.

Nit: `FromCSVFile.__init__` still takes `csv_file`, so `main.py` calls it as `FromCSVFile(csv_file=input_file, ...)`. #797 notes users rarely see that class, so it's fine either way.

henrykironde:
@proofbyhuman could you please [book an appointment with me to go through your PRs](https://bookings.cloud.microsoft/bookwithme/user/472ccdc8404e4a83abb9c84af8fc7c2b@ufl.edu?anonymous&ismsaljsauthenabled&ep=plink) ?

henrykironde:
@proofbyhuman, thank you for putting in these PRs. We’re closing them for now since we weren’t able to connect with you for a one on one live walkthrough.

PR #1417, opened 2026-07-29
Fix #1274: PadIfNeeded only pads when image is smaller than target size

### LINKED ISSUE #1274: PadIfNeeded augmentation is poorly named. It crops as well as pads.

proofbyhuman:
## Description

This PR fixes #1274 by implementing a custom `PadIfNeeded` augmentation in `src/deepforest/augmentations.py`.

Key changes:
- Implemented `PadIfNeeded` as a `GeometricAugmentationBase2D` subclass.
- Behavior: Pads the image to the specified target size if it is smaller than that size. If the image is larger than or equal to the target size along a dimension, it is left untouched (preventing unwanted cropping).
- Updated `_SUPPORTED_TRANSFORMS` mapping to use the custom `PadIfNeeded` class.
- Added comprehensive unit tests in `tests/test_augmentations.py` verifying behavior for smaller, larger, and equal image dimensions, both standalone and through `get_transform()`.
- Formatted and linted with `ruff`.

## Related Issue(s)

Closes #1274

## AI-Assisted Development

Gemini (Google DeepMind Agentic Assistant)

### DIFF
--- a/src/deepforest/augmentations.py
+++ b/src/deepforest/augmentations.py
@@ -88,6 +88,75 @@ class RandomPadTo(GeometricAugmentationBase2D):
         return input[..., : size[0], : size[1]]
 
 
+class PadIfNeeded(GeometricAugmentationBase2D):
+    """Pad the input tensor to a target size if it is smaller than that size.
+
+    If the input tensor is larger than or equal to the target size in a dimension,
+    no padding or cropping is applied along that dimension.
+
+    Args:
+        size: Target size as (height, width) or a single int for (size, size).
+        pad_mode: Padding mode ('constant', 'reflect', 'replicate', 'circular').
+        pad_value: Fill value for constant padding.
+        p: Probability of applying the transform.
+        same_on_batch: Apply same transformation to all batch elements.
+        keepdim: Maintain shape.
+    """
+
+    def __init__(
+        self,
+        size: tuple[int, int] | int = (800, 800),
+        pad_mode: str = "constant",
+        pad_value: float = 0,
+        p: float = 1.0,
+        same_on_batch: bool = False,
+        keepdim: bool = False,
+    ) -> None:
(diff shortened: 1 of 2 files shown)


proofbyhuman:
Both points addressed.

**Shared `inverse_transform`.** `RandomPadTo` and `PadIfNeeded` now both delegate to a module-level `_crop_to_size` helper. I kept it as a plain function rather than reworking the class hierarchy, so `RandomPadTo` is otherwise untouched.

**The `p=0.5` case.** You were right to flag it. The test fails, and the failure is real: with `p<1.0` Kornia applies the transform to a subset of the batch and writes the result back into the batch, which raises `RuntimeError` because `PadIfNeeded` deliberately changes the image size, so the padded and untouched elements no longer match. That can't be fixed inside `apply_transform` without changing what the augmentation is for, so I added the test as `xfail(strict=True)` with the reason inline rather than dropping the case silently.

If you'd rather `PadIfNeeded` reject `p<1.0` explicitly, or not expose `p` at all, that's a design call and I'm happy to follow whichever you prefer.

Local run on this branch, since the gated suite hasn't executed:

```
$ uv run pytest tests/test_augmentations.py
29 passed, 1 xfailed in 46.02s
```

henrykironde:
@proofbyhuman, thank you for putting in these PRs. We’re closing them for now since we weren’t able to connect with you for a one on one live walkthrough.

Issue #1380, opened 2026-04-20
Better documentations of checkpoints


bw4sz:
The alive/dead model doesn't have a card. 

https://huggingface.co/weecology/cropmodel-deadtrees

I know this is hugging face, but I think the issue will get more visibility here. I need to go back and look at this checkpoint.

proofbyhuman:
Hi @weecology team,

I would like to contribute improvements to the checkpoint documentation!

**Proposed Improvements:**
1. Fix code snippet in `docs/user_guide/11_training.md` (correcting variable names `m` vs `model`, proper string path usage with `tempfile.TemporaryDirectory()`, standard `.ckpt` extension, and missing `get_data` import).
2. Add a clear comparative reference table contrasting PyTorch Lightning checkpoints (`.ckpt`) with Hugging Face Hub format checkpoints (`.safetensors` + `config.json`).

I've tested the code snippet against the Python engine and validated formatting with pre-commit. Ready to open the PR!

PR #1414, opened 2026-07-24
fix: provide user-friendly error message for Hugging Face HTTP 401 au…

proofbyhuman:
# fix: provide user-friendly error message for Hugging Face HTTP 401 authentication errors

## Description

This PR improves error reporting when downloading models or crop models from Hugging Face Hub.

Previously, if a user attempted to load a private/gated model without valid credentials or an expired token, Hugging Face raised an unhandled `HTTPError: 401 Client Error`.

**Changes:**
1. Wrapped `model_class.Model().create_model()` in `deepforest.load_model` (`src/deepforest/main.py`) and `hf_hub_download` in `CropModel.from_pretrained` (`src/deepforest/model.py`) to catch HTTP 401 / Unauthorized exceptions.
2. Raises a clear `ValueError` explaining how to configure the `HF_TOKEN` environment variable or run `huggingface-cli login`.
3. Added unit test `test_load_model_hf_auth_error` in `tests/test_main.py`.

## Related Issue(s)

Closes #1401

## AI-Assisted Development

- [x] I used AI tools (e.g., GitHub Copilot, ChatGPT, etc.) in developing this PR
- [x] I understand all the code I'm submitting
- [x] I have reviewed and validated all AI-generated code

**AI tools used (if applicable):**
Google Antigravity AI assistant for design implementation, unit testing, and pre-commit validation.

PR #1412, opened 2026-07-23
fix: automatically override num_workers=0 for TiledRaster prediction …

proofbyhuman:
# fix: automatically set num_workers=0 for TiledRaster prediction dataset

## Description

This PR improves usability when predicting on out-of-memory rasters (`TiledRaster` / `dataloader_strategy="window"`). 

Previously, if `config.workers > 0`, `predict_tile` raised a strict `ValueError`, halting execution and requiring users to manually mutate `config.workers` and recreate the trainer.

**Changes:**
1. In `predict_dataloader`, if `isinstance(ds, prediction.TiledRaster)` and `num_workers > 0`, `num_workers` is automatically overridden to `0` with an informative warning (`UserWarning`), eliminating script crashes.
2. Removed the halting `ValueError` check in `predict_tile`.
3. Added unit test `test_predict_tile_window_auto_workers` in `tests/test_main.py`.

## Related Issue(s)

Relates to #1392 (this improves the `workers` usability discussed in that thread; it does not fully resolve the original not-tiled raster problem, so it does not close the issue).

## AI-Assisted Development

- [x] I used AI tools (e.g., GitHub Copilot, ChatGPT, etc.) in developing this PR
- [x] I understand all the code I'm submitting
- [x] I have reviewed and validated all AI-generated code

**AI tools used (if applicable):**
Google Antigravity AI assistant and Claude Code for design implementation, unit testing, and pre-commit validation.

PR #1411, opened 2026-07-23
feat: print informative message when validating coordinates during da…

proofbyhuman:
# feat: add informative message when validate_coordinates is enabled

## Description

This PR adds a clear stdout log message when `validate_coordinates=True` (default) during `TrainingDataset` initialization, informing the user about the validation process:

```python
Validating coordinates for N images  
  proofbyhuman/antigravity

PR #1418, opened 2026-07-29
Fix #999: Replace hardcoded paths in Australia notebook with get_data and tempdir

### LINKED ISSUE #999: Fix 'Working with deepforest data' docs page
This page is rendered via a notebook (I think) and currently has a number of errors due (at least) to a hard coded path:

Image

proofbyhuman:
## Description

This PR fixes #999 by removing hardcoded local user paths in the 'Working with deepforest data' documentation notebook (`docs/user_guide/examples/Australia.ipynb`).

Key changes:
- Replaced hardcoded paths (`/Users/benweinstein/Downloads/Plot13Ortho.tif`) with `get_data("australia.tif")`.
- Replaced hardcoded output directories with cross-platform `tempfile.gettempdir()`.
- Added required imports (`os` and `tempfile`) to code blocks.

## Related Issue(s)

Closes #999

## AI-Assisted Development

Gemini (Google DeepMind Agentic Assistant)


vickysharma-prog:
Hi @proofbyhuman  welcome to Deepforest, you have 10 PRs open (#1409–#1418). I read all ten diffs and their linked issues. Consolidating into one comment instead of ten threads.

## The set isn't mergeable as a set

- **#1416 contains #1415's commit.** Its commit list starts with `docs: improve checkpoint documentation and fix code snippet in user g…` and its diff duplicates the entire `docs/user_guide/11_training.md` change from #1415.
- **#1409, #1410, #1412, #1416 all edit overlapping regions of `src/deepforest/main.py`.** #1409 rewrites `predict_file`'s signature; #1416 edits inside that same function. #1412 and #1416 both change `predict_dataloader`.
- **#1413 adds a new `predict_file(csv_file=...)` call site** — the keyword #1409 deprecates.

Whichever merges first conflicts with the rest.

## CI hasn't run

The green checks are only `pre-commit.ci` and `readthedocs`. The test suite is gated on maintainer approval, which you asked for on #1412. No test in any of these ten has executed. #1409 also states the suite wasn't run locally.

## Per-PR

**#1410 — duplicates existing code.** `ImagesCallback.on_train_start` at `src/deepforest/callbacks.py:53` already logs sample training images at train start, same default of 5 (`dataset_samples=5`). The existing one guards on `distributed.is_global_zero(trainer)` (yours has no rank guard — every DDP rank would log), skips on `fast_dev_run`, and samples the actual dataset via `dataset[idx]`, i.e. post-augmentation. Yours re-reads the raw CSV, so it shows pre-augmentation annotations and won't catch the transform/coordinate bugs the description cites. It's also always-on rather than opt-in. Suggest closing.

**#1409 — call sites not updated.** 11 sites still pass `csv_file=` as a keyword and would emit `DeprecationWarning`: `src/deepforest/scripts/predict.py:55` (production code), `tests/test_evaluate.py:24,50,315,379`, `tests/test_IoU.py:14`, `tests/test_main.py:433,1169`, `tests/profile_evaluate.py:15`, `tests/profile_predict_file.py:17`, `docs/user_guide/12_evaluation.md:201`. Two more pass positionally (`tests/test_main.py:441`, `docs/user_guide/16_prediction.md:133`) — those won't warn, they silently rebind to `input_file`.

Also: `root_dir` now defaults to `None`, but the docstring's "if none, uses `image_dir` in config" fallback isn't implemented, so `None` fails.

**#1416 — orphans `config.batch_size`.** After this change nothing reads it, but it stays in `src/deepforest/conf/config.yaml:8` and `src/deepforest/conf/schema.py:148`. Anyone setting it is silently ignored, no warning. Needs removal or explicit deprecation. #889's item 3 ("update the config doc") also isn't covered. Minor: `bird.yaml` keeps `batch_size: 32` alongside the new keys, and there's an unrelated lambda reformat in `evaluate()`.

**#1415 — wrong issue.** `Closes #1380` is incorrect. #1380 asks for a HuggingFace model card for `weecology/cropmodel-deadtrees`; this edits `docs/user_guide/11_training.md`. Separately, the rewritten example calls `m.create_trainer(fast_dev_run=True)` then `m.trainer.fit(m)` without configuring `config.train.csv_file`, so it fails as written.

**#1414 — fragile detection.** Catches bare `Exception` and substring-matches `"401"` / `"Unauthorized"`. `huggingface_hub` raises typed `GatedRepoError` / `HfHubHTTPError` with `.response.status_code` — catch those and check `== 401`. Substring matching false-positives on any error text containing "401" (commit hash, path, port). The test patches with a plain `Exception("401 Client Error...")`, so it validates the substring match against its own mock, not real HF behavior. Also #1401 is a CI failure on a gated repo; a better message doesn't fix it, so this likely doesn't close the issue.

**#1413 — network-dependent pin.** Pins `0.80`/`0.72` against a live download of `weecology/deepforest-tree`. Fails for everyone the moment the model is republished, including on unrelated PRs. Deriving from integer counts (44/55, 44/61) is right, but consider pinning a `revision` or adding a marker so it can be deselected. Also uses `csv_file=`, deprecated by your #1409.

**#1412 — removes a deliberate error.** Deletes the `workers > 0` `ValueError` in `predict_tile` and replaces it with a silent auto-override. That error looks intentional; turning a loud failure into a quiet correction is a maintainer decision. Test also calls `m.load_model("weecology/deepforest-tree")` — network-dependent and slow.

**#1417 — matches #1274's spec.** `inverse_transform` is identical to `RandomPadTo`'s, worth sharing rather than duplicating. `apply_transform` returns a different shape than its input — fine at the default `p=1.0`, but you expose `p`, and if Kornia applies to a subset of a batch at `p<1.0` the shapes may not concat. Worth testing with `p=0.5`.

**#1418 — valid.** `australia.tif` does ship in `src/deepforest/data/`. But the notebook's stored cell outputs weren't regenerated, so the rendered docs page still shows results from the old `Plot13Ortho.tif`. #999 is about the page rendering incorrectly, so re-run the notebook before committing.

**#1411 — fine.** Bare `print()` matches the surrounding code (`datasets/training.py:71`). `callbacks.py` uses `pl_module.print()` (rank-aware) but that isn't available in a dataset constructor.

## On the AI-assisted workflow
Please review the Ai given code before submitting 
You disclosed AI use on all ten via the template and named the tool each time — that's correct and not the issue. What's showing is that the review step is being skipped: #1411's description contains a stray `proofbyhuman/antigravity` string, #1410's ends with `verification.ification`, #1415 targets the wrong issue, and #1410 rebuilds something already in `callbacks.py`. Those are all catchable in a minute of reading. The "I have reviewed and validated all AI-generated code" checkbox is the load-bearing part, not the disclosure.

The specific gap is that these tools don't survey the existing codebase before adding to it — grepping for an existing implementation before starting would have caught #1410, and probably the #1409/#1413/#1416 overlap too.

## Suggestion

Pick 2–3, draft the rest until those land. #1417, #1418 and #1409 look fine. That gives maintainers something reviewable and lets later PRs build on a merged base instead of conflicting.

proofbyhuman:
Hola Viki Sharma, estoy muy agradecido con la devolución, tendré en cuenta
todas las sugerencias y actuaré en consecuencia muchas gracias .
atte.
Ramiro guevara

proofbyhuman:
Thanks for taking the time to go through all ten — one thread was far more useful than ten.

The process points are done: #1410 and #1414 are closed, and #1411, #1412, #1413, #1415 and #1416 are drafts. That leaves #1409, #1417 and #1418.

On re-running this notebook: I can't execute it as-is, and since the blocker is pre-existing code I'd rather ask than change something you didn't ask me to touch.

Cell 7 ends with:

```python
plot = visualize.plot_results(results=boxes, image=r)
plt.imshow(plot)
plt.show()
```

`plot_results` returns `matplotlib.figure.Figure` (or `Axes` with `axes=True`), so `plt.imshow` raises `TypeError: Image data of dtype object cannot be converted to float`. Those three lines are unchanged from `main` — this PR doesn't touch them.

You're right that the outputs are the crux. To be precise about the current state: this branch *cleared* the stored outputs, because they embedded `/Users/benweinstein/Downloads/...` paths from the original run. With `nbsphinx_execute = "never"` in `docs/conf.py`, the rendered page now has no outputs at all rather than stale ones — which doesn't fix #999 either. So re-running is a hard requirement for this PR, not a nice-to-have; I just can't get there while cell 7 raises.

The minimal fix would be dropping the `plt.imshow(plot)` / `plt.show()` lines, since `plot_results` already draws with `show=True`. Would you prefer I do that here so the notebook can be executed and its outputs regenerated, or keep this PR to the path change and open a separate one for the execution fix?

One more thing I noticed while checking: `australia.tif` is 751x708 with 3 bands (EPSG:32753), not 4. The markdown in cells 3 and 5 walks through the four-band error and dropping the alpha channel, which was written for `Plot13Ortho.tif` and no longer matches what the code does. Let me know if you want that in scope too.

vickysharma-prog:
The path fix is right and `australia.tif` does ship in `src/deepforest/data/`.

One thing worth deciding. `docs/conf.py:82` sets `nbsphinx_execute = "never"`, so the docs build renders whatever outputs are committed rather than re-running the notebook. The stored outputs are now `"outputs": []`, so the errors from #999 are gone, but the two plots the page used to render are gone with them and it'll show code cells only.

`nest_detection.ipynb` ships with no outputs either, so that isn't out of place here.

@ethanwhite, you filed #999 is the page fine rendering code with no outputs, or do you want the plots back? If the latter, running the notebook against `australia.tif` and saving it puts the outputs back into the same `.ipynb`, no new files.

Minor: the cells still write `Plot13Ortho_crop.tif` and `Plot13Ortho_crop.shp` even though the input is now `australia.tif`.

PR #1425, opened 2026-08-16
build: require numpy >= 2.0.0

### LINKED ISSUE #1422: [Compatibility Issue] Missing numpy>=2 constraint — np.typing crashes on numpy<2
## Affected File

- `src/deepforest/visualize.py` (lines 17, 20, 136, 306, 373)
- `src/deepforest/datasets/training.py` (line 119)

## Current Code

```python
# visualize.py
def plot_prediction_data(...) -> np.typing.NDArray[np.bool_]:          # line 17
def plot_prediction_points(...) -> np.typing.NDArray[np.bool_]:       # line 20
def dynamic_fire_rating(...) -> np.typing.NDArray[np.bool_]:           # line 136
def fire_rating_image_preprocess(...) -> np.typing.NDArray[np.bool_]:  # line 306
def calculate_fire_rating(...) -> np.typing.NDArray[np.float32]:       # line 373

# datasets/training.py
def load_image(self, idx) -> np.typing.NDArray[np.float32]:            # line 119
```

No `from __future__ import annotations`, no safe `import numpy.typing` — type annotations are evaluated at definition time.

## Root Cause

`numpy.typing` is only exposed as a module attribute in **numpy >= 2.0.0**. On numpy < 2.0, `np.typing.NDArray` triggers `AttributeError: module 'numpy' has no attribute 'typing'`.

## Dependency

`pyproject.toml` declares `"numpy"` with no version constraint. Transitive dependencies also do not force numpy>=2: `pip install --dry-run` with `numpy<2` added resolves successfully to **numpy 1.26.4**.

```
$ pip install --dry-run -r deps.txt numpy<2
  → numpy 1.26.4 ✓  (no constraint violated)
  → import deepforest → np.typing.NDArray → AttributeError
```

## Impact

- **Severity: High** — hard crash on import if numpy<2 is installed or resolved

## Solution

Add `numpy>=2` to `dependencies` in `pyproject.toml`, or add one of the following to `visualize.py`:

- `import numpy.typing`
- `from __future__ import annotations`
- `from numpy.typing import NDArray` and use bare `NDArray`

## References

- NumPy 2.0 release notes — `numpy.typing` exposed as module attribute
- Similar fix: [numexpr#540](https://github.com/pydata/numexpr/issues/540)

xyf5432:
Fixes #1422

## Description

Require numpy >= 2.0.0 in `pyproject.toml`. The `np.typing.NDArray` annotations in `src/deepforest/visualize.py` and `src/deepforest/datasets/training.py` are evaluated at module import time, and on numpy < 2.0 `np.typing` is only accessible if some earlier import has executed `import numpy.typing` — an incidental side effect that depends on which packages are installed and their versions (e.g. pandas 3.x triggers it at module level, pandas 2.x does not).

As discussed in #1422, pinning the floor is preferred over deferring annotation evaluation: numpy 2.0.0 is over two years old, the rest of the dependency stack already requires modern releases, and pinning eliminates this class of version-dependent behavior.

```diff
-    "numpy",
+    "numpy>=2.0.0",
```

## Verification

- Dependency declaration is constraint-only — no code paths affected.
- The remaining dependencies (geopandas, rasterio, kornia, pytorch-lightning, etc.) all support numpy 2.x.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ dependencies = [
     "hydra-core",
     "kornia>=0.8.2",
     "matplotlib",
-    "numpy",
+    "numpy>=2.0.0",
     "omegaconf",
     "opencv-python-headless>=4.5.4",
     "pandas",


jveitchmichaelis:
Thanks for raising this. I would prefer to constrain numpy to >2 unless there is a reason we need to support v1 as the rest of our stack is already constrained to new releases, and it's reasonable to expect numpy 2 as well (even if technically code won't break with 1).

ethanwhite:
Pinning to >=2.0.0 is good for me since 2.0.0 is now over 2 years old (which is forever with the computer vision stack). We've had some other weird edge cases with <2.0 installs so this would eliminate those. I also tried to replicate this on 1.26.4 and couldn't, probably due to the upstream import mentioned in the PR.

ethanwhite (approved):

PR #1426, opened 2026-09-06
fix: allow empty validation ranks in DDP metrics

### LINKED ISSUE #1407: multi-gpu ddp fails if any one rank has all empty validation.
DDP validation hangs (NCCL ALLGATHER timeout) when a rank's val shard has no non-empty images

Where: deepforest/main.py, _compute_epoch_metrics() (box task branch)

```
if len(self.iou_metric.groundtruth_labels) > 0:
    metrics.update(self.iou_metric.compute())
    output = self.mAP_metric.compute()
...
metrics.update(self.precision_recall_metric.compute())
```

Bug: self.iou_metric.groundtruth_labels reflects only the local rank's accumulated state (only non-empty-target images get update()-ed). Under DDP, if one rank's validation shard happens to contain zero non-empty images (easy to hit with shuffle=False + a small/empty-frame-heavy val set), that rank skips iou_metric.compute() / mAP_metric.compute() entirely — but those are collective ops, and every other rank still calls them. The skipping rank never issues the matching all_gather, so the rest hang until the NCCL watchdog kills the job.

Confirmed via crash logs: the watchdog dump showed one rank exactly one collective op behind the other two, with all ranks' stack traces pointing at _compute_epoch_metrics → torchmetrics.Metric.compute() → gather_all_tensors → NCCL all_gather.

Fix: don't gate a distributed .compute() call on rank-local state — always call it (torchmetrics returns a sane empty-state result) so every rank issues the same collectives:

```
metrics.update(self.iou_metric.compute())
output = self.mAP_metric.compute()
output = {k: v for k, v in output.items() if k != "classes"}
metrics.update(output)
metrics.update(self.precision_recall_metric.compute())
```

Repro: multi-GPU DDP, small/empty-frame-heavy validation CSV, val_accuracy_interval=1. Reliably reproduces with 3 ranks on a ~3.5k-row, ~99%-empty-frame validation set.

hardikkaurani:
## Summary

Fixes #1407.

When running validation with DDP, a rank can receive an empty validation shard while other ranks have validation samples. Previously, `_compute_epoch_metrics()` skipped `IntersectionOverUnion.compute()` when the local rank had no ground-truth labels.

This caused ranks with data to enter TorchMetrics distributed synchronization while the empty rank skipped it, which could result in a distributed validation hang.

### Changes

- Always call `IntersectionOverUnion.compute()` so every DDP rank participates in distributed synchronization.
- Handle the expected zero-sample `ValueError` when the entire validation set is empty.
- Re-raise unexpected `ValueError` exceptions instead of silently swallowing them.
- Add regression tests covering:
  - empty rank still calling `compute()`
  - all-ranks-empty state
  - unexpected `ValueError` propagation

### Validation

- `pytest tests/test_main.py -k "test_compute_epoch_metrics"` -> 3 passed
- `ruff check src/deepforest/main.py` -> passed
- `git diff --check` -> passed
- Python 3.11.9
- torchmetrics 1.9.0

### DIFF
--- a/src/deepforest/main.py
+++ b/src/deepforest/main.py
@@ -985,17 +985,20 @@ class deepforest(pl.LightningModule):
         metrics = {}
 
         if self.model.task == "box":
-            # IoU and mAP
-            if len(self.iou_metric.groundtruth_labels) > 0:
+            # IoU and mAP - call compute unconditionally so all DDP ranks participate in collective sync
+            try:
                 metrics.update(self.iou_metric.compute())
-                # Lightning bug: claims this is a warning but it's not. See issue #16218 in Lightning-AI/pytorch-lightning
-                output = self.mAP_metric.compute()
+            except ValueError as e:
+                if "No samples to concatenate" in str(e):
+                    pass
+                else:
+                    raise
 
-                # Remove classes from output dict
-                output = {
-                    key: value for key, value in output.items() if not key == "classes"
-                }
-                metrics.update(output)
+            output = self.mAP_metric.compute()
+
+            # Remove classes from output dict
+            output = {key: value for key, value in output.items() if not key == "classes"}
+            metrics.update(output)
             metrics.update(self.precision_recall_metric.compute())
 
         elif self.model.task == "point":
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1493,3 +1493,52 @@ def test_detections_per_img_and_topk_candidates_config():
     m.create_model()
     assert m.model.detections_per_img == 500
     assert m.model.topk_candidates == 2000
+
+
+def test_compute_epoch_metrics_empty_rank_calls_compute(m, monkeypatch):
+    """Test that a rank with no local groundtruth labels still calls compute()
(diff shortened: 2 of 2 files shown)


ethanwhite:
Hi @hardikkaurani - Thanks for contributing to DeepForest! To keep our reviewing load manageable in an age of AI assistants we ask that folks contributing changes please engage in discussions in issues before submitting PRs:

> Issue discussion: For non-trivial changes, discuss your approach in an issue first. This can be done by either responding to an open issue indicating an interest in addressing the issue and an outline of how you plan to do so, or by creating a new issue describing the changes you think should be made (e.g., what the bug or feature is), that you would be interested in helping implement them, and what your general plan is for doing so.

https://deepforest.readthedocs.io/en/stable/development/contributing.html#contributing-code

Once we've had a chance to discuss your planned approach and we're all on the same page we'll ask you to go proceed with a PR.

Given that approach I'm going to close your current PRs, ask that you pick one of them to pursue first, and engage in a discussion in the appropriate issue.

